PR #416, opened 2014-12-12
Remove the X11 dependency on OSX (& some other improvements)

ManuelBlanc:
X11 is a pretty hefty dependency on OSX. However it's only used to access the system clipboard.
This functionality has been implemented natively, and X11 can be done away with.

Minor improvements:
* Made CMake detect the include path for `libintl` headers.
* Modified the input module to use the command key for shortcuts.


piotrdz:
Sorry that it's taken so long to address this pull request, but I've only recently got my hands on Mac OS X virtual machine. I've merged your changes manually and verified that the code compiles on OSX. However, I can't tell much more since I can't get OpenGL to work under VirtualBox. But I trust that the code works in runtime as well :).

Issue #419, opened 2014-12-26
CBOT - strings can be assigned to arrays of ints and cause crash


VictorDenisov:
int a[2];
a[0] = "a";
a[1] = "b";

This program compiles, but when is started crashes the game.

krzys-h:
This must've been fixed at some point, now it correctly results in a compilation error

PR #420, opened 2015-01-04
Don't crash when number is assigned to string

VictorDenisov:
Fix for issue #419

### DIFF
--- a/src/CBot/CBot.cpp
+++ b/src/CBot/CBot.cpp
@@ -1582,7 +1582,7 @@ CBotInstr* CBotExpression::Compile(CBotToken* &p, CBotCStack* pStack)
             return NULL;
         }
 
-        CBotTypResult    type1 = pStack->GetTypResult();
+        CBotTypResult    type2 = pStack->GetTypResult();
 
         // get the variable assigned to mark
         CBotVar*    var = NULL;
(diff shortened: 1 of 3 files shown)


krzys-h:
I don't think this is a proper way to fix that issue. By design, CBot doesn't do automatic type casting, so something like this should just NOT work and show compile error instead (and it does so in current version):
```C++
string a;
a = 0;
```
The same thing should happen when working with arrays, as described in issue #419 (and that doesn't work in current version, it just crashes the game on execution):
```C++
int[] a;
a[0] = "test";
```
This might change when we rewrite whole CBot engine, but currently I'd prefer typecasting in CBot to stay as it is right now.

VictorDenisov:
Int to float, boolean to string, even string to float and string to int casts already exist, but these two shouldn't? How so? Is it better to crash the app instead? I just used an existing mechanism for type casts and changed order of assignments.

krzys-h:
These casts exist, but don't happen automatically.
```C++
int a = 5.5; // ok, works
string b = true; // compiles but crashes
float c = "1.5"; // doesn't compile, invalid assigment
int d = "5"; // doesn't compile, invalid assigment
```

To convert anything to string, you have to force a cast like so:
```C++
string a = ""+true;
string b = ""+5;
string c = ""+1.5;
```
To cast a string to number, you do this:
```C++
int a = strval("5");
float b = strval("1.5");
```

krzys-h:
These casts exist, but don't happen automatically.
```C++
int a = 5.5; // ok, works
string b = true; // compiles but crashes
float c = "1.5"; // doesn't compile, invalid assigment
int d = "5"; // doesn't compile, invalid assigment
```

To convert anything to string, you have to force a cast like so:
```C++
string a = ""+true;
string b = ""+5;
string c = ""+1.5;
```
To cast a string to number, you do this:
```C++
int a = strval("5");
float b = strval("1.5");
```

PR #490, opened 2015-07-05
Set CBotVar::m_binit type to specific enum instead of plain int.

dreamer-dead:
Hi!

The main purpose of this request is to fix compiler warnings about CBotVar::m_binit member.
It was just int variable and there were many lines when m_binit was initialized with a boolean value.
Also sometimes literal '2' was used as value for this member (if I understand it correctly, that means an object value or a pointer value).
I used class based enumeration for all possible values to fix unwanted type casts.
As a minor changes I can mention comparison with NAN (replaced with std::isnan() function), and changing unordered_map::operator[] to unordered_map::find (to avoid cases when the map doesn't contain requested key).

Hope that it can be useful.

### DIFF
--- a/src/CBot/CBot.cpp
+++ b/src/CBot/CBot.cpp
@@ -459,7 +459,7 @@ bool CBotListInstr::Execute(CBotStack* &pj)
         if (!p->Execute(pile)) return false;
         p = p->GetNext();
         if (p == NULL) break;
-        if (!pile->IncState()) ;//return false;                    // ready for next
+        (void)pile->IncState();                                  // ready for next
     }
 
     return pj->Return(pile);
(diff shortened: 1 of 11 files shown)


dreamer-dead:
It seems that Travis builds failed due to version of CMake on Travis worker machine.
Is that OK?

dreamer-dead:
Thanks for the review!
I fixed names and used auto as you pointed.
Also my local helper function VarIsNAN() now uses enum::operator>() as the old code does.

piotrdz:
All right, now I can merge your changes. Don't worry about Travis, we should actually get rid of it.
Thanks for having patience to deal with CBot code :-). If you want to help us out some more, you can join our forums or IRC channel.

PR #544, opened 2015-07-29
fix type highlighting in the CBOT editor

### LINKED ISSUE #539: Wrong syntax highlighting in the CBOT editor
It is said that

    In a program, the name of a type is always written on green background.

But it is not:

image

Also, I don't understand why `this` is considered a type (it's on green background).

vieux:
Fixes #539

ping @MrSimbax

### DIFF
--- a/src/CBot/CBotString.cpp
+++ b/src/CBot/CBotString.cpp
@@ -29,6 +29,7 @@
 //Map is filled with id-string pars that are needed for CBot language parsing
 const std::map<EID,const char *> CBotString::s_keywordString =
 {
+    {ID_THIS,       "this"},
     {ID_IF,         "if"},
     {ID_ELSE,       "else"},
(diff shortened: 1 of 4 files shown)


krzys-h:
I know "this" is not a type, but it's a keyword so it should be highlighted somehow.
@MrSimbax what do you think?

vieux:
@krzys-h @MrSimbax I updated the PR to fix the whole highlighting:

screen shot 2015-07-29 at 00 35 42


I highlighted this as a regular keyword, it might be a good idea to create a new color for it.

MrSimbax:
@vieux @krzys-h I think that's a fairly good solution. Unless we will try to make a new color, of course, but it should be first discussed and I don't think that we really need it, especially for only one keyword. Maybe we should highlight all variable names in the future? Anyway, I have no objections against merging this.

PR #575, opened 2015-08-18
New CBot syntax highlighting

krzys-h:
This is my proposal for new CBot syntax highlighting. Tell me if you like it or not.
image

### DIFF
--- a/src/graphics/engine/text.cpp
+++ b/src/graphics/engine/text.cpp
@@ -659,11 +659,11 @@ void CText::DrawString(const std::string &text, std::vector<FontMetaChar>::itera
         {
             if (hl == FONT_HIGHLIGHT_TOKEN)
             {
-                c = Color(0.612f, 0.475f, 0.208f, 1.0f); // #9C7935
+                c = Color(0.490f, 0.380f, 0.165f, 1.0f); // #7D612A
             }
             else if (hl == FONT_HIGHLIGHT_TYPE)
             {
-                c = Color(0.365f, 0.498f, 0.251f, 1.0f); // #5D7F40
+                c = Color(0.31f, 0.443f, 0.196f, 1.0f); // #4F7132
             }
             else if (hl == FONT_HIGHLIGHT_CONST)
             {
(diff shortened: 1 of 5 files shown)


RaptorParkowsky:
I like it, but the previous hightligting used the texture parts from, well, some button#.png or other textures. I'd like to remove that texture parts if they are not used somewhere else. This would be another step to make UI textures as clean as possible and good step into CEGUI.

krzys-h:
I believe it might have used to in the original game, but in the Gold Edition it seems to just use color values in the code. You can even see it disables texturing before rendering the text:
https://github.com/colobot/colobot/pull/575/files#diff-8a4a7a28f640852bfde25e15f022644bR836
You would probably have to ask @piotrdz, I believe he is the one who rewrote the CText for the new engine

MrSimbax:
The font doesn't look like it's monospaced. I'd also use colors which are more distinct, contrast.

piotrdz:
@RaptorParkowsky: CText does not use any texture files for rendering actual font characters, but it draws icons in text using texture files:
https://github.com/colobot/colobot/blob/dev/src/graphics/engine/text.cpp#L855-L877

krzys-h:
@piotrdz Those "icons" are just buttons that are embedded in SatCom help files
@MrSimbax Yeah, the font seems to be not monospaced (again, we fixed this already at some point...). You could play around with the colors a bit and see if you can make it look beter.

piotrdz:
[jenkins] Build finished successfully

krzys-h:
This is how it looks with a monospaced font:
image

krzys-h:
@MrSimbax is that better?
image

piotrdz:
[jenkins] Build finished successfully

krzys-h:
@piotrdz Don't you think that showing the check as finished is enough? It doesn't need to spam the comments all the time

piotrdz:
@krzys-h: it should be disabled now.

MrSimbax:
I meant lighter text or darker background. Ugh, it's hard to discuss really, as it's a matter of taste. Wish we could have custom themes by now :/ I think it's fairly good, you can't match everyone's personal taste anyway.

Mentions of colors in SatCom needs to be changed. They should be rephrased like "types are shown like that:" instead of "types are always written with green color".

tomangelo2:
We could add later a tab in Options to customize the colors.
Also we should think in future for special mode for colorblind people.

krzys-h:
image
I think this can be merged now. Any other throughts before that?

PR #657, opened 2015-09-28
fix MacOSX install instruction and compilation after switch to SDL2

vieux:
@piotrdz @krzys-h this might break windows/linux, I only compile on Mac, please take a look.


piotrdz:

piotrdz:
@piotrdz [Jenkins]: test this please.

krzys-h:
USE_SDL_MAIN should be also removed from CMake files if not needed anymore. I'll try compilation on Windows later.

krzys-h:
49ee00d90d01d63fc1baaf8596d589162e839034

PR #676, opened 2015-11-01
Add unit tests for class CBotString.

Grunaka:
Add comments in class CBotString.
Change existing comments in class CBotString to be compliant with doxygen.
Add function "const char* CBotString::CStr() const".
Add null pointer check in function "const CBotString& CBotString::operator+=(const CBotString& str)"
Remove functions "CBotString CBotString::Mid(int nFirst, int nCount) const" and "CBotString CBotString::Mid(int nFirst) const" because they were doublons of function "CBotString CBotString::Mid(int nFirst, int lg) const"

### DIFF
--- a/src/CBot/CBotDll.h
+++ b/src/CBot/CBotDll.h
@@ -250,37 +250,146 @@ private:
 
 
 ////////////////////////////////////////////////////////////////////////
-//
-// as part of MFC CString not used here.
-//
-// ( all functions are not implemented yet )
-
-/** \brief CBotString Class used to work on strings */
+/**
+ * \brief CBotString Class used to work on strings
+ */
 class CBotString
 {
 public:
+
+    /**
+     * \brief CBotString Default constructor.
+     */
     CBotString();
+
+    /**
+     * \brief CBotString
+     * \param p
+     */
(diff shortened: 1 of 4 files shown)


piotrdz:

krzys-h:
@piotrdz You are the one responsible for tests and code style so I'm leaving the merge to you. It looks good, except for being a bit pointless since we're planning to rewrite the whole CBot engine anyway.

Grunaka:
@krzys-h What change do you have planned for CBot Engine ?
I would like to start cleaning the CBot interface as suggested in the colobot forum thread "Tasks for volunteers". Let me know if this is still relevant.

piotrdz:
Tests are not exactly pointless, as they will help verify that rewritten code works the same as old implementation. That said, we would like to rewrite CBOT at more fundamental level, so we would need acceptance tests at the interpreter level, running sample CBOT code and expecting some meaningful results.
@Grunaka: the tasks I wrote about on forum are still relevant. I'll give you more information on IRC.
I'll accept this pull request, as any tests are better than no tests for CBOT, but I'd expect further tests to be written at higher level.

PR #678, opened 2015-11-08
Fix #670

### LINKED ISSUE #670: Clicking the area where start battle button would be crashes game in CBOT editor.
This is after starting a code battle mission.
The start battle button appears and all buttons in the interface lose their functionality.
The keyboard still works after that.  You can still move and resize the editor window
but the buttons and the text field do not move with it.

version 0.1.6-alpha compiled locally unmodified.
OS: Xubuntu x86_64

piotrwalkusz1:


piotrdz:

piotrdz:
We really need a bot account for Jenkins :P
Anyway, code looks good, so I merged it with only minor whitespace change.

PR #690, opened 2015-12-05
Clean CBot Interface

Grunaka:
I have finish to clean CBot interface.
Basically I moved each class in their own source and header files.


piotrdz:

krzys-h:
The fact that you should include only the CBot.h file (the public API part) and not the individual internal files should be documented somewhere

krzys-h:
I'll test and merge this pull request today

krzys-h:
Sorry that it took so long, I didn't have time to test it until now

Issue #653, opened 2015-09-27
Add alternative way to rotate camera with joystick/keyboard


tomangelo2:
Currently rotating camera is available only by sticking cursor to game window edge. There could be separate buttons to rotate a camera left/right. It could help playing on joysticks.

krzys-h:
I started implementing this already earlier, in some modes you can use numpad keys to rotate the camera. There are no keybindings yet.

krzys-h:
After bd72086704a6b61222a9708b848875585d84ae4e camera navigation with keyboard is mostly finished and joystick should be quite easy to implement

Issue #649, opened 2015-09-25
It's impossible to play with joystick without using a mouse


tomangelo2:
Navigating in UI and aiming with Shooters is possible only with mouse. It makes game impossible to complete only with a joystick.

krzys-h:
Any ideas on how to fix that?

tomangelo2:
Maybe one analog could control cursor, when neutral, a cursor will be centered. But in this case moving or flying controls should be moved somewhere else.

krzys-h:
If we do that, we should still keep an option for people who have some crazy joysticks with >=5 axes ;)

PR #709, opened 2016-01-29
Right click camera

krzys-h:
Just found some old code in the CCamera class that does what I wanted to to for a long time ;) I wanted to test how right-click to rotate camera would work in Colobot. After some experimenting I think it's a lot easier to use and more precise than the old move cursor to the edge one. Feel free to test the dev-right-click-camera build and discuss here.
Currently works only in object back camera.


krzys-h:
The next commit changes camera smoothing to be a lot faster. In my opinion this is a lot better. This applies globally, not only to camera movement (try changing objects for example).

RaptorParkowsky:
I think it would be better to rotate camera by middle mouse button in both X and Y axis, just like in Blender or something. Then we would add right mouse button to other feature, like dolly-zooming or changing size of lens, or maybe other more common thing in gameplay itself. Optionally some of the camera features would be available only with "freecam" or other camera related cheats.

I don't know, camera smoothing was already pretty fast and elegant to me. Even effective. But I must download this build somewhere somehow when it'll be available and check this all by myself for final opinion.

pastuh:
Adlast i can rotate camera with mouse.

But.. im still missing ability to move camera while i watch code window..
I mean.. i want to see code and model actions too ..
Camera should follow model while i watch code window :)

krzys-h:
I'm still waiting for opinions on this

RaptorParkowsky:
I highly prefer to using a middle mouse button and bring back an option to switch into old "cursor-to-the-edge" option as most of RTS games have both solutions. Old "cursor-to-the-edge" rotating should be turned off as default and switching it on shouldn't deactivating our new middle click camera rotating.

I'm not able to test your recent changes, especially for camera moving for known reason though :/ .

krzys-h:
I still think right click is better than middle click. Right mouse button is easier to press and just feels more correct to me. Waiting on opinion from others. I don't see any reason why we couldn't make it accept both, Colobot doesn't use any other mouse buttons anyway.
Unfortunately current keybinding system is unable to handle mouse buttons :(

krzys-h:
Both right and middle mouse button work now.

tomaszkax86:
We should definitely add control option for binding to mouse. Remember we might want gamepads to support this feature as well. Primary action/selection (left mouse button) could also use binding option in the future. Also, middle mouse button could be used to move camera when nothing is selected and in CBot editor. This feature could become part of potential future in-game level editor, you could freely move and rotate camera around with mouse.

tomangelo2:
Well, mouse is currently irreplaceable, there are issues #653 and #649 about that. Without mouse you can't aim, navigate in UI or control a camera. With these commits there will be fixed issue with camera, but aiming and UI navigating will still be available only with mouse.

RaptorParkowsky:
@tomangelo2 With that moving around UI without a mouse (only by keyboard or gamepad) we could do something like it was in BuzzingCars EPSITEC's game. Only if we had it's source code... Otherwise we just probably wait for CEGUI implementation.

krzys-h:
> Only if we had it's source code...

At this point of the development having the source code would be no better than reimplementing it because you would need to adjust that code to all our refactoring anyway

RaptorParkowsky:
@krzys-h But it's the same engine like Colobot so it would be probably easy to implement and/or refactor in GOLD though.

krzys-h:
> it would be probably easy to implement

So like I said, it would be our own reimplementation anyway, so having the original source code doesn't matter

krzys-h:
And obviously CEGUI will be a reimplementation of the whole UI code anyway

tomangelo2:
@RaptorParkowsky But our code is currently quite modified comparing to original Colobot, having BC source code with their modifications and trying to copy that to our code with our modifications wouldn't be so much better than just implement this once while rewriting interface to CEGUI.

RaptorParkowsky:
Okay, finally tested this one and it works perfectly for me. I think it's worth merging with 0.1.7 release.

PR #714, opened 2016-02-10
3 issues

Erihel:
I didn't split in in different pull requests. Commits are not tht big.


piotrdz:

krzys-h:
In SatCom, the tabs on the left are not displayed properly. I guess they are not really needed since we are displaying only the generic CBot tab, but then at least the active tab arrow should be hidden.
Other than this small detail, everything seems to work fine. Thanks!

Issue #519, opened 2015-07-19
Crash after opening SatCom - Satellite view


tomangelo2:
Only happen on dev-release build, on dev-debug there is everything allright.
```
Program received signal SIGSEGV, Segmentation fault.
0x007ec86d in PHYSFS_tell (handle=0xaa88568)
    at /opt/mxe/tmp-physfs/physfs-2.0.3/physfs.c:2038
2038    /opt/mxe/tmp-physfs/physfs-2.0.3/physfs.c: No such file or directory.
(gdb) backtrace
#0  0x007ec86d in PHYSFS_tell (handle=0xaa88568)
    at /opt/mxe/tmp-physfs/physfs-2.0.3/physfs.c:2038
#1  PHYSFS_seek (handle=0xaa88568, pos=22900)
    at /opt/mxe/tmp-physfs/physfs-2.0.3/physfs.c:2054
#2  0x0041121f in CSDLFileWrapper::SDLSeek(SDL_RWops*, int, int) ()
#3  0x00636c5a in RWread (stream=0x19e67b38, offset=22900, buffer=0x0,
    count=0) at SDL_ttf.c:365
#4  0x00729d1d in FT_Stream_Seek (stream=0x19e67b38, pos=22900)
    at /opt/mxe/tmp-freetype/freetype-2.5.0.1/src/base/ftstream.c:65
#5  0x0073be0b in tt_loader_init (loader=loader@entry=0x22f750,
    size=size@entry=0x19e97718, glyph=glyph@entry=0x19e975e8,
    load_flags=load_flags@entry=0, glyf_table_only=<optimized out>,
    glyf_table_only@entry=0 '\000')
    at /opt/mxe/tmp-freetype/freetype-2.5.0.1/src/truetype/ttgload.c:2133
#6  0x0073d652 in TT_Load_Glyph (load_flags=0, glyph_index=36,
    glyph=0x19e975e8, size=0x19e97718)
    at /opt/mxe/tmp-freetype/freetype-2.5.0.1/src/truetype/ttgload.c:2249
#7  tt_glyph_load (ttslot=0x19e975e8, ttsize=0x19e97718, glyph_index=36,
    load_flags=0)
    at /opt/mxe/tmp-freetype/freetype-2.5.0.1/src/truetype/ttdriver.c:403
#8  0x007293af in FT_Load_Glyph (face=face@entry=0x19e96d20,
    glyph_index=glyph_index@entry=36, load_flags=load_flags@entry=0)
[...]

Issue #708, opened 2016-01-28
Crash related to TTF and PHYSFS (in SatCom?)


krzys-h:
I think I saw this already in some other issue, can't find it now. I think @piotrdz was working on fixing this at some point. Happened after clicking a link in SatCom.
```
(gdb) bt
#0  0x00007fceb45e389b in __GI___waitpid (pid=14702, stat_loc=stat_loc@entry=0x7ffc72ec0b80, options=options@entry=0) at ../sysdeps/unix/sysv/linux/waitpid.c:40
#1  0x00007fceb455cffb in do_system (line=<optimized out>) at ../sysdeps/posix/system.c:148
#2  0x00000000007bd294 in CSystemUtilsLinux::SystemDialog (this=0x119a800, type=SDT_YES_NO, title="Try to save?", 
    message="You can try saving the game at the moment of a crash. Keep in mind, the game engine is in\nan unstable state so the saved game may be corrupted or even cause another crash.\n\nDo you want to try saving n"...) at /home/krzys_h/Pulpit/colobot/src/app/system_linux.cpp:68
#3  0x00000000007ba52f in CSignalHandlers::ReportError (errorMessage="SIGSEGV, segmentation fault") at /home/krzys_h/Pulpit/colobot/src/app/signal_handlers.cpp:175
#4  0x00000000007b9990 in CSignalHandlers::SignalHandler (sig=11) at /home/krzys_h/Pulpit/colobot/src/app/signal_handlers.cpp:57
#5  <signal handler called>
#6  0x00007fceb518c8cb in PHYSFS_tell () from /usr/lib/x86_64-linux-gnu/libphysfs.so.1
#7  0x00007fceb518cc08 in PHYSFS_seek () from /usr/lib/x86_64-linux-gnu/libphysfs.so.1
#8  0x00000000007deaad in CSDLFileWrapper::SDLSeek (context=0x5db2d70, offset=22900, whence=0) at /home/krzys_h/Pulpit/colobot/src/common/resources/sdl_file_wrapper.cpp:132
#9  0x00007fceb66efd59 in ?? () from /usr/lib/x86_64-linux-gnu/libSDL2_ttf-2.0.so.0
#10 0x00007fceb1635a2b in FT_Stream_Seek () from /usr/lib/x86_64-linux-gnu/libfreetype.so.6
#11 0x00007fceb164c908 in ?? () from /usr/lib/x86_64-linux-gnu/libfreetype.so.6
[...]

Issue #619, opened 2015-08-30
Clicking the 'categories' link in Windows build may cause crash


attila123:
I got this with recent CI builds e.g. build 97:
http://compiled.colobot.info/jenkins/view/colobot-gold-download/job/colobot-gold/branch/dev/97/

Steps to reproduce:
- Stop (if running) and start colobot (to start from a well-defined state)
- Start e.g. the first Earth mission
- Press F2 to display the CBOT programming help
- Click on the 'categories' link (under 'Constants')
- Colobot crashes:
```
[DEBUG]: Starting music...
[DEBUG]: Music loaded from cache
[INFO]: Game paused - SatCom opening animation
[INFO]: Game resumed
[INFO]: Game paused - SatCom

Unhandled exception occured!
==============================
SIGSEGV, segmentation fault
==============================

This is usually caused by a bug. Please report this on http://github.com/colobot
/colobot/issues
including information on what you were doing before this happened and all the in
formation below.
==============================
You are running version git-HEAD~r6711154 from CI build #97

The game was in phase PHASE_SIMUL (ID=7)
Last started level was: category=missions chap=1 rank=1
==============================

Sorry for inconvenience!
```
In this help page if you select first e.g. the 'extern' link, then click the 'back' button, and only then click 'categories', then it does not crash.
In Linux I could not reproduce this.

piotrdz:
Seems to be fixed now.

attila123:
Unfortunately it still crashes with the latest Jenkins build 99:

```
[DEBUG]: Music loaded from cache
[INFO]: Game paused - SatCom opening animation
[INFO]: Game resumed
[INFO]: Game paused - SatCom

Unhandled exception occured!
==============================
SIGSEGV, segmentation fault
==============================

This is usually caused by a bug. Please report this on http://github.com/colobot
/colobot/issues
including information on what you were doing before this happened and all the in
formation below.
==============================
You are running version git-HEAD~ra47da3f from CI build #99

The game was in phase PHASE_SIMUL (ID=7)
Last started level was: category=missions chap=1 rank=1
==============================

Sorry for inconvenience!
```

krzys-h:
Unfortunately, I still can't reproduce this :( Tried the exact same build on Windows and no crash. Could you try to at least provide a gdb backtrace?

attila123:
Hi,
Now I noticed that it does not crash in full screen (or if the game was started in full screen).
So you need to make sure to start up the game in windowed mode.
[...]

PR #717, opened 2016-02-11
Fix crash related to TTF and PHYSFS

Erihel:
TTF fonts will be loaded to memory instead of file pointers

Added new class CSDLMemoryWrapper that loads data from PHYSFS into memory block
and closes file after. This closes issues #519, #708 and #619


piotrdz:

Erihel:
Please test this before merging. I do not experience crashes anymore but it may be just my luck.

krzys-h:
> Can one of the admins verify this patch?

Can @piotrdz disable this automatic message as it is not really necessary? ;)

---

Both I and @tomangelo2 have failed to reproduce this issue in this build, so I think this confirms the fix. Did you figure out what exactly was wrong with the previous implementation?

PR #737, opened 2016-03-20
Fix for various issues with initializing variables in CBOT

melex750:


src/CBot/CBotInstr/CBotDefArray.cpp
```diff
@@ -96,17 +103,23 @@ CBotInstr* CBotDefArray::Compile(CBotToken* &p, CBotCStack* pStack, CBotTypResul
...
-            if ((inst->m_listass = CBotTwoOpExpr::Compile(p, pStk)) != nullptr)
+            pStk->SetStartError(p->GetStart());
+            if ( IsOfType(p, ID_SEP) )
             {
-                if (!pStk->GetTypResult().Compare(type))  // compatible type ?
-                {
-                    pStk->SetError(CBotErrBadType1, p->GetStart());
-                    goto error;
-                }
+                pStk->SetError(CBotErrBadLeft, p->GetPrev());
```
krzys-h: This should be a new error type, "expression expected after =" (in all the CBotDef* classes)

krzys-h:
This code doesn't work, I'm pretty sure it did previously:
```C++
extern void object::Nowy()
{
	
	string a[2];
	
}
```

I'm not sure why it didn't fail the unit tests.

melex750:
This works:
```string[] a;```
That's the way it's set up in CBotDefString, I didn't change it.
It would be easy to make that work though.

I thought if strings will be arrays of chars at some point, something like that might look strange.

File: src/CBot/CBotInstr/CBotListArray.cpp [hunk 2]
```diff
@@ -45,57 +45,111 @@ CBotListArray::~CBotListArray()
...
+                if ( p->GetType() == ID_CLBLK ) break;
+
+                pStk->SetError(CBotErrClosePar, p);
+                goto error;
             }
         }
         else
         {
             pStk->SetStartError(p->GetStart());
+            if (classItem && IsOfType(p, TokenTypVar, ID_NEW)) // don't allow func() or var between "{ }"
```
krzys-h: What is the reason these are not allowed?

src/CBot/CBotInstr/CBotListArray.cpp
```diff
@@ -45,57 +45,111 @@ CBotListArray::~CBotListArray()
...
             if (nullptr == ( inst->m_expr = CBotTwoOpExpr::Compile( p, pStk )))
             {
                 goto error;
             }
             CBotVar* pv = pStk->GetVar();                                       // result of the expression
+            CBotTypResult retType;
+            if (pv != nullptr) retType = pv->GetTypResult(CBotVar::GetTypeMode::CLASS_AS_INTRINSIC);
 
-            if (pv == nullptr || !TypesCompatibles( type, pv->GetTypResult()))     // compatible type?
+            if (pv == nullptr || (type.Eq(CBotTypString) && !retType.Eq(CBotTypString)) ||
```
krzys-h: I'm pretty sure TypesCompatibles would be enough here

krzys-h:
You are right, this code did not work previously. It should though, so fix it if you can.

src/CBot/CBotInstr/CBotListArray.cpp
melex750: In a class definition:
```
int b = 1;
int x[] = {b}; // doesn't work
int a[] = {move(1)}; // crashes the game

public class ClassName
{
    SomeClass sc[] = {new SomeClass}; // doesn't work and/or crashes
    ClassName cn2[] = {new ClassName}; // crashes game
}
```
The same goes for non-arrays.

src/CBot/CBotInstr/CBotListArray.cpp
melex750:
```
string[] s = {1, 2}; // crashes the game
string s2 = 1;      // also crashes
```
?

krzys-h: The first one properly throws CBotErrBadType1.

The second one doesn't work, but I think it is valid code and should implicitly cast to string on execution like it does in most other places. See the VarImplicitCast test:
```C++
// TODO: I don't actually know what the exact rules should be, but it looks a bit wrong
// TODO: Current version of this code causes a failed assertion
TEST_F(CBotUT, DISABLED_VarImplicitCast)
{
    ExecuteTest(
        "extern void ImplicitCast()\n"
        "{\n"
        "    int a = 5;\n"
        //"    ASSERT(a == \"5\");\n"
        "    string b = a;\n"
        "    ASSERT(b == \"5\");\n"
        //"    ASSERT(b == a);\n"
        "    \n"
        "    string c = \"2.5\";\n"
        //"    ASSERT(c == 2.5);\n"
        //"    float d = c;\n"
        //"    ASSERT(d == c);\n"
        //"    ASSERT(d == 2.5);\n"
        "}\n"
        "\n"
        "extern void AssignImplicitCast()\n"
        "{\n"
        "    string a = 2;\n"
        "    ASSERT(a == \"2\");\n"
        //"    a = 3;\n"
        //"    ASSERT(a == \"3\");\n"
        "    string b = 2.5;\n"
        "    ASSERT(b == \"2.5\");\n"
        //"    b = 3.5;\n"
        //"    ASSERT(b == \"3.5\");\n"
        "}\n"
    );
}
```

src/CBot/CBotInstr/CBotListArray.cpp
krzys-h: The solution is not to block them, but to make them work correctly. I know that current implementation of CBot probably makes this really hard.

```C++
public class ClassName {
    int b = 1;
    int x[] = {b}; // this should work assuming b is declared before x
    int a[] = {move(1)}; // this should be invoked on 'new' before constructor

    SomeClass sc[] = {new SomeClass}; // this should be valid code
    ClassName cn2[] = {new ClassName}; // this is valid code on compilation but should throw a stack overflow on execution as it recurses infinitely
}
```

Even if we decide to block that temporarily, this is a bad way to do so. Instead of looking for specific tokens here, it would be better tell the expression compilation code to disallow those.

src/CBot/CBotInstr/CBotListArray.cpp
krzys-h: Actually, the first one should be valid code too and cast numbers to string

melex750: I'll remove the checks for string.  It may be, that it only crashed in a class definition but I thought otherwise.

src/CBot/CBotClass.cpp
```diff
@@ -561,29 +562,27 @@ bool CBotClass::CompileDefItem(CBotToken* &p, CBotCStack* pStack, bool bSecond)
             while ( IsOfType( p, ID_OPBRK ) )   // a table?
             {
                 CBotInstr* i = nullptr;
-
+                pStack->SetStartError( p->GetStart() );
                 if ( p->GetType() != ID_CLBRK )
+                {
                     i = CBotExpression::Compile( p, pStack );           // expression for the value
+                    if (i == nullptr || pStack->GetType() >= CBotTypBoolean) // must be a number
```
krzys-h: This is not really a correct fix. Not everything that is a number is a valid index, only integers are.
```C++
float a[];
a[2.3] = 7.5;
a[true] = 6; // this throws a runtime error
```

melex750: Agreed, that should be ```!= CBotTypInt```.

Your example code is runtime, which I didn't change.

src/CBot/CBotInstr/CBotListArray.cpp
melex750: I wasn't sure exactly how that should work.
I think the last example maybe did overflow a stack, just not the CBotStack.
I can remove those checks, but I'm not comfortable 
messing with the expression compilation code.
I agree that's how it should be done.

piotrdz:

melex750:
The items at the right of AClass didn't work before,
 each var would keep the size of the one before the comma even without index brackets.
```c++
public class BClass{ int[] i = {1,2,3}; }
public class AClass
{
	   int i[] = {1,2,3},          j[][] = {{4, 5} , {/*move(1)*/}, null, {}};
	  bool b[][] = {{true,false}}, c[][][] = {{{true},{false}}};
	 float f[] = {1.2, 3.14},      g[][]   = {{6.3},{4.5}};
	string s[] = {"A","B","C"},    t[][]   = {{"d"},{"e"}}, u[][] = {{"f","g"}};
	
	// These compile, but crash when compiler finds a "new AClass()".
	// The same is true for other functions.
	//object o[] = {radar(Any)},     p[][] = {/*radarall(Any), {radar(Any)}*/};
	
	point pt[] = {new point(0,3,0)}, q[][] = { {new point(1,2,3)} };
	BClass bc[] = {new BClass()};
}
extern void object::ArrayInit()
{
	//string qw = 1; // compiles, crash on run
	//string wx[] = {1}; // doesn't compile
	// All of this works.
	   int i[3] = {1,2,3},       j[4][2] = { {4,5} , {move(1)}, null, { } };
	  bool b[] = {true,false},  c[][] = {{false}, null};
	 float f[] = {1.2, 3.14},   g[][] = {{6.3},{4.5}};
	string s[] = {"A","B","C"}, t[2][] = {{"d"},{"e"}};
	object o[] = {radar(Any)},  p[][] = {radarall(Any), {radar(Any)}, {null}};
	point pt[] = {new point(0,0,0)}, q[][] = { {}, {new point(1,2,3)} };
	message( (s[1]) );
	AClass ac[] = {new AClass()};
	message(""+(ac[0].pt[0]) );
	message(""+(ac[0].bc[0].i[2]) );
}
```
Strings are as they have been, only: ```string s = "" + 1;```, but they have more array syntax now.

src/CBot/CBotInstr/CBotListArray.h
```diff
@@ -41,7 +41,7 @@ class CBotListArray : public CBotInstr
      * \param classItem
```
melex750: I could swear i deleted that :/

src/common/restext.cpp
```diff
@@ -715,6 +715,7 @@ void InitializeRestext()
     stringsCbot[CBot::CBotErrBadIndex]      = TR("Incorrect index type");
     stringsCbot[CBot::CBotErrPrivate]       = TR("Private element");
     stringsCbot[CBot::CBotErrNoPublic]      = TR("Public required");
+    stringsCbot[CBot::CBotErrNoExpression]  = TR("expression expected after =");
```
krzys-h: All the other error messages start with an uppercase letter

krzys-h:
You seem to be doing a lot of testing of various possible syntaxes, how about adding some of them to unit tests while you're at it? This will help to ensure we won't break it accidentaly in the future.
https://github.com/colobot/colobot/blob/dev/test/unit/CBot/CBot_test.cpp
PS. There is more non-working syntax mentioned in that file that I discovered while writing some first tests.

melex750:
I was mostly testing both sides of comma because it's a point of recursion
when initializing >= 2D arrays or defining multiple variables on the same line.
(defining class member variables in the same line, loops at comma)
The game would often crash if a comma was missing somewhere between```{ {} {} }```.
It was [always here that it seemed to crash](https://github.com/colobot/colobot/blob/dev/src/CBot/CBotInstr/CBotDefInt.cpp#L90).

melex750:
Now that CompileDefItem() [finds the right method](https://github.com/colobot/colobot/pull/737/files#diff-8821e5c2bf3dc26e42c3536cfabcebb2R615) on the second pass,
CBotFunction::Compile() will accept the pre-compiled method [as input here](https://github.com/colobot/colobot/blob/dev/src/CBot/CBotClass.cpp#L655-L656), instead of passing nullptr.
Then there's no longer a need to swap pointers [just below that](https://github.com/colobot/colobot/blob/dev/src/CBot/CBotClass.cpp#L661-L666).
I tested this, and it works as expected.

krzys-h:
Merged. I hope this doesn't break anything in unpredictable ways :see_no_evil:

krzys-h:
@melex750 Any other issues you fixed that need to be closed? ;)

PR #745, opened 2016-03-25
Update the colobot manpage to reflect the evolutions up to 0.1.7

OdyX:
Also refresh the .pot and the french translation


piotrdz:

Issue #722, opened 2016-03-02
Rename SatCom button in main menu and make it bigger


RaptorParkowsky:
I think the name "SatCom" for that button isn't really straight to the point of this feature.

Propositions:

- "Programming help" - more like in the official SatCom UI in the missions. I think for now should be good.
- "CBOT Documentation"

or
- "SatCom Viewer" -  more universal and might be more practical for future SatCom development.

And I think this button, if possible with current UI code, could be a little bigger.

Erihel:
Changing label will require changing all translations and those are frozen. SatCom is the same in all translations. I would stay with SatCom. What do others think?

krzys-h:
I think SatCom can stay

RaptorParkowsky:
Well, I think "SatCom" name can stay but only for now as this isn't really straight to the point name for this place of the game. SatCom is this Astronaut's satellite communicator device which shows to him instructions about the mission/exercise, etc, but this button opens CBOT documentation outside of the game's world in the SatCom's UI. It doesn't open any other informations about the game's world and plot.

PR #749, opened 2016-03-29
Change SatCom button size in main menu

Erihel:
Change button size as proposed in issue #722


piotrdz:

Issue #736, opened 2016-03-19
"Origin of last message" button may cause permanent pause in level simulation


RaptorParkowsky:
When the message pop up and you hit just once on "KEYPAD .", then everything is ok and you can just admire the source of the last message frozen in time.

The problem happens when you hit the same button twice or more while message pop up, or try to switch between more message pop ups. When you'll back to gameplay, the time in the game is still frozen and even simple restart doesn't work. You must go back to menu and then run the level again after few seconds. Otherwise strange stuff will happen...

PR #750, opened 2016-03-29
Deactive pause before jumping to next visit

Erihel:
When pause was active setting new pause resulted in strange behavior.
This should fix issue #736.


piotrdz:

PR #751, opened 2016-03-30
Fix spelling error, and add Keywords to the desktop file

OdyX:


piotrdz:

krzys-h:
@piotrdz Could you disable this message? :P

PR #756, opened 2016-04-02
Smok trying hard to be uesful

Smok94:
Hope now everything is ok.

### DIFF
--- a/src/level/robotmain.h
+++ b/src/level/robotmain.h
@@ -150,6 +150,9 @@ const int SATCOM_PROG       = 4;
 const int SATCOM_SOLUCE     = 5;
 const int SATCOM_MAX        = 6;
 
+const int CMD_HISTORY_SLOTS = 10;
+const int CMD_MAX_LENGHT = 50;
+
 class CRobotMain : public CSingleton<CRobotMain>
 {
 public:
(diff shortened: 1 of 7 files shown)


piotrdz:

Smok94:
Holly... It was hard but already learned how to use git shell.

src/level/robotmain.h
```diff
@@ -150,6 +150,9 @@ const int SATCOM_PROG       = 4;
 const int SATCOM_SOLUCE     = 5;
 const int SATCOM_MAX        = 6;
 
+const int CMD_HISTORY_SLOTS = 10;
+const int CMD_MAX_LENGHT = 50;
```
krzys-h: 1. It's LENGTH not LENGHT
2. There should be no limit to the length. We are actively working on removing all such limits and refactoring to std::string.

krzys-h:
I have merged your first two commits - 6216daedb32c607ac1405475b1d0f4c22559198e
If you want the command history to be merged, you will have to rewrite it using C++ features, probably std::deque and std::string. Look at @piotrdz 's code related to CRobotMain::m_selectionHistory as it does nearly the same thing but with objects instead of strings.

Issue #32, opened 2012-07-07
CBOT - using static initializers in classes crashes the game


Mrocza:
This code:

public class exchange 
 { 
   static private bool global = null; 
 }

Causes immediate game crash at hitting "OK" button in program editor.

krzys-h:
Status after #737 :

The code from the first post does work as expected

The code from second posts hangs the game

melex750:
It throws this error when it calls the wrong method in the baseclass.
```CBotInstr::Execute(CBot::CBotStack*&): Assertion `0' failed.```

The instruction to initialize a static variable is executed on compile,
and arrays should be compiled with CBotDefArray for the proper
instruction to be executed. This may be an easy fix.

Otherwise maybe re-working CBotClass::CompileDefItem()
to use some of the compiling system that is used in functions
and other instruction blocks.
[[Something like this]](https://github.com/colobot/colobot/blob/dev/src/CBot/CBotInstr/CBotInstr.cpp#L211-L222) might fix other issues with classes too.

I think the first option.  Things other than static arrays work ok.

PR #762, opened 2016-04-05
Fix initializing arrays in class definition

melex750:
issue #32
and [TEST_F(CBotUT, DISABLED_ArraysInClasses)](https://github.com/colobot/colobot/blob/dev/test/unit/CBot/CBot_test.cpp#L528)


piotrdz:

PR #768, opened 2016-04-19
Fix function and method overloading

melex750:
Specifically, with array types as parameters.

Currently, this compiles and shouldn't.
```c++
void test(int[] i){}
void test(int[] i){}
extern void object::New()
{
}
```
This crashes the game on compile.
```c++
public class AClass
{
    void testMethod(int[] i){}
}
extern void object::New()
{
}
```
This is the bug that broke UserLevelHelper userlevel,
when method overloading was fixed.


piotrdz:

PR #773, opened 2016-05-01
Group source files for IDE projects in CMake

MrSimbax:
Before:

After:

I've made a pull request, because I'm not sure if what I've done will mess up with other IDEs than Visual Studio, so I'd be glad if other people could test it.


MrSimbax:
```
CMake Error at src/CMakeLists.txt:9 (get_filename_component):
  get_filename_component unknown component DIRECTORY
Call Stack (most recent call first):
  src/CBot/CMakeLists.txt:169 (group_sources)
```

Too old cmake?

krzys-h:
@MrSimbax Use PATH instead

> PATH      = Legacy alias for DIRECTORY (use for CMake <= 2.8.11)

krzys-h:
I don't see a reason to split headers and source files into separate lists. They are both used in the same places anyway, and it would be easier to add new entries if you didn't have to scroll through the whole file to add it in two separate places.
Also, I don't see a reason to split source and header files into separate directories in IDE, you usually want to have them next to each other, it's easier to switch between them when doing stuff that way.

src/CMakeLists.txt
```diff
@@ -3,6 +3,19 @@ set(CMAKE_CXX_FLAGS "${COLOBOT_CXX_FLAGS} ${MXE_CFLAGS}")
...
 
+# Groups source files for IDE project
+macro(group_sources SOURCES GROUP_ROOT)
+    foreach(SOURCE_FILE ${SOURCES})
+        get_filename_component(GROUP "${SOURCE_FILE}" PATH)
+        if(PLATFORM_WINDOWS)
+            string(REPLACE "/" "\\" GROUP "${GROUP}")
+            set(GROUP "${GROUP_ROOT}\\${GROUP}")
+        else()
+            set(GROUP "${GROUP_ROOT}/${GROUP}")
```
krzys-h: Are you sure this is correct? The documentation says you should always use backslashes:

> The \<name\> of the group may contain backslashes to specify subgroups
>
> https://cmake.org/cmake/help/v3.0/command/source_group.html

MrSimbax: I thought that this groping could be considered for other IDEs as well. If not, then should I place this whole macro in if(MSVC)?

MrSimbax:
Ok, I'll place them in one list.

About the two folders, that's how it's done by default in Visual Studio, although I'd prefer to have them in one too. Meh, nobody would probably complain about that, so ok.

src/CMakeLists.txt
krzys-h: I don't know if any other IDE cares about this setting, but if it does, it's going to expect path to be split by \\. Just use \\ always, independent from the system, like you did before 0b2ff9098e61aca87471dfbf9070d6d1b8c5047d.

Issue #436, opened 2015-01-31
Passing arguments as this.something causes weird behaviour


krzys-h:
```C++
public class TestClass {
    public point test_point;

    public void test(point bot_position)
    {
        message("Class: Test point is "+test_point.x+"; "+test_point.y);
        message("Class: Bot is at "+bot_position.x+"; "+bot_position.y);
    }
}

extern void object::TestBugThis()
{
    TestClass test = new TestClass();
    test.test_point.x = 100;
    test.test_point.y = 200;
    message("Main program: Test point is "+test.test_point.x+"; "+test.test_point.y);
    message("Main program: Bot is at "+this.position.x+"; "+this.position.y);
    test.test(this.position);
}
```
This causes an error that position is not a member of *this* while calling test.test()

```C++
public class TestClass {
    public int something;
    public point test_point;

    public void test(point bot_position)
    {
        message("Class: Test point is "+test_point.x+"; "+test_point.y);
        message("Class: Bot is at "+bot_position.x+"; "+bot_position.y);
    }
}

extern void object::TestBugThis()
{
    TestClass test = new TestClass();
    test.test_point.x = 100;
    test.test_point.y = 200;
    message("Main program: Test point is "+test.test_point.x+"; "+test.test_point.y);
    message("Main program: Bot is at "+this.position.x+"; "+this.position.y);
    test.test(this.position);
}
```
This causes test.test() to receive test.test_point as an argument.
![2](https://cloud.githubusercontent.com/assets/1517255/5987871/24a0d3fc-a944-11e4-9cf9-21e27603f156.png)

Looks to be related to #207 and #256 in some way.

krzys-h:
This works correctly:
```C++
public class TestClass {
    public point test_point;

    public void test(point bot_position)
    {
        message("Class: Test point is "+test_point.x+"; "+test_point.y);
        message("Class: Bot is at "+bot_position.x+"; "+bot_position.y);
    }
}

extern void object::TestBugThis()
{
    TestClass test = new TestClass();
    test.test_point.x = 100;
    test.test_point.y = 200;
    message("Main program: Test point is "+test.test_point.x+"; "+test.test_point.y);
    message("Main program: Bot is at "+this.position.x+"; "+this.position.y);
    point testing = this.position;
    test.test(testing);
}
```

Issue #256, opened 2013-11-29
CBOT: passing object reference more than once does not work.


arturcz:
Following code does not work in colobot:
```cbot
public class Count {
    void Count() {
    }
    float range(point p1, point p2) {
        return distance2d(p1, p2);
    }
}

public class Move {
    object o;
    Count c;
    void Move(object ob) {
        o=ob;
        c=new Count();
        message("From Move::Move: "+ob.category+" "+o.category);
    }
    void run(point p) {
        message("From Move::run: "+o.category+" P: "+o.position);
        turn(direction(p));
        move(c.range(o.position,p)); // here is the problem
    }
}

extern void object::Test()
{   
    message("From test: "+category);
    Move m=new Move(this);
    object a=radar(RedFlag);
    m.run(a.position);
}

```
It results with error message: This is not a member of this class. It looks like the CBot engine does not recognize the o as a valid object in context of calling c.range() method. Also replacing this code with:
```cbot
float d=c.range(o.position,p);
move(d);
```
also does not help.

Test using Debian package 0.1.2-2.

PR #781, opened 2016-05-24
Fix "this" bug "256 & 436"

melex750:
----------------------------------------------------
Found another bug while fixing "this" bug.
The following bug happens because RestoreState leaves
all instance.variables with m_ident == 0.
```c++
extern void object::New()
{
    point pnt(2,4,6);
    Test  tst(1,2,3);
    message("waiting");
    wait(3);             // save the game here, then reload.
    message("" + pnt.x); // error: this is not a member of this class
    message("" + tst.b); // error: this is not a member of this class
}
public class Test {
    void Test(float A, float B, float C) {
    // saving the game here also breaks a, b, c.
    this.a = A; b = B; c = C;
    }
    float a, b, c;
}
```
----------------------------------------------------
Fix for #256 and #436:

Games that were saved while executing a method call
in a user defined class are not directly compatible
with this fix and will cause a failed assert.

It can be made compatible, but it's probably not needed.
There are probably very few such saved games, if any.

### DIFF
--- a/src/CBot/CBotInstr/CBotInstrMethode.cpp
+++ b/src/CBot/CBotInstr/CBotInstrMethode.cpp
@@ -119,7 +119,7 @@ bool CBotInstrMethode::ExecuteVar(CBotVar* &pVar, CBotStack* &pj, CBotToken* pre
 
         pThis->SetName("this");
         pThis->SetUniqNum(-2);
-        pile1->AddVar(pThis);
+        pile1->SetVar(pThis);
         pile1->IncState();
     }
     int        i = 0;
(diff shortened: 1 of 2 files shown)


piotrdz:

krzys-h:
Thanks! I'll test this when I have a moment.
In the meanwhile, jenkins, test this please. (see, now I know what that message from @piotrdz actually means :stuck_out_tongue:)

piotrdz:
[jenkins] Build finished successfully

krzys-h:
Another one of biggest bugs in CBot fixed. Thanks!

melex750:
My pleasure, thank you.

PR #783, opened 2016-05-29
[INFO] Conversion to string

melex750:
@krzys-h, I saw you were working on this so I'm sharing what I did a few days ago.

But first, recent changes appear to have broke these conversions:
```c++
message("" + this);      // output: "Pointer"
message("" + array);     // output: "("
message("" + position);  // output: "point("
```
As for this PR:
The first two commits are good, imo.
I'm not very happy with the third, but it works.

Fixes these bugs:
[TEST_F(CBotUT, DISABLED_FunctionBadReturn)](https://github.com/colobot/colobot/blob/dev/test/unit/CBot/CBot_test.cpp#L664)
Currently, you can return a string from **any** non-void function
and crash the game if you assign that to something other than a string.

[TEST_F(CBotUT, DISABLED_BadStringAdd_Issue535)](https://github.com/colobot/colobot/blob/dev/test/unit/CBot/CBot_test.cpp#L936)
```c++
anArray[0] = "A"; // "issue-419"

voidfunc() + "";  // crash on run

message( "" + voidfunc() ); // crash on run

aVar += ;
aVar = ;
```

### DIFF
--- a/src/CBot/CBotInstr/CBotExpression.cpp
+++ b/src/CBot/CBotInstr/CBotExpression.cpp
@@ -71,6 +71,13 @@ CBotInstr* CBotExpression::Compile(CBotToken* &p, CBotCStack* pStack)
             return nullptr;
         }
 
+        if ( p->GetType() == ID_SEP )
+        {
+            pStack->SetError(CBotErrNoExpression, p);
+            delete inst;
+            return nullptr;
+        }
+
         inst->m_rightop = CBotExpression::Compile(p, pStack);
         if (inst->m_rightop == nullptr)
         {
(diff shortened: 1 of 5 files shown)


melex750:
Did it start building automatically? ~not good.

krzys-h:
Huh, it looks like the new Jenkins configuration builds PRs from outside contributors automatically and I don't even see a config option to disable it...

melex750:
Possible solution [HERE.](https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin)
This looks like it might be the Jenkins plugin you were using before.

melex750:
Sorry, I forgot to mention that this PR lets this work too:
```c++
string str;
str = 2.4; <-- expressions like this are commented-out in unit tests.
str += 4.2;

   NOTE: arithmetic and boolean operations require
   parentheses when concatenating strings.

message("" + (2+2) + "" );  without "( )"  output is "22"
```

PR #789, opened 2016-06-09
Dev boost filesystem removal

Erihel:
This pull request removes boost::filesystem from resource manager. Now boost::filesystem is only used before resource manager takes control. I did not touch CBot code and how files are handled there.

There's also small change to save naming.

### DIFF
--- a/src/level/robotmain.cpp
+++ b/src/level/robotmain.cpp
@@ -105,6 +105,7 @@
 
 #include "ui/screen/screen_loading.h"
 
+#include <algorithm>
 #include <iomanip>
 #include <stdexcept>
 #include <ctime>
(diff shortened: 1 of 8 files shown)


krzys-h:
:+1:, looks good, I'll test this when I have some time

src/level/robotmain.cpp
```diff
@@ -5482,93 +5483,47 @@ void CRobotMain::SetAutosaveSlots(int slots)
...
 }
 
 int CRobotMain::GetAutosaveSlots()
 {
     return m_autosaveSlots;
 }
 
-int CRobotMain::AutosaveRotate(bool freeOne)
+// Remove oldest saves with autosave prefix
+void CRobotMain::AutosaveRotate()
```
krzys-h: Removing the freeOne parameter broke behaviour of autosave rotation when changing amount of autosave slots in settings. Now when you adjust the number of slots, the AutosaveRotate function will remove one save too much (it assumes that a new save is going to be always created immediately after, which is not the case)

Erihel: I guess `AutosaveRotate()` call can be removed when setting slot number. This way it'll remain simple and it'll delete old saves on next autosave.

Issue #683, opened 2015-12-02
Polymorphism works in CBOT


melex750:
It was already implemented, but broken, like inheritance.
SOURCE_CODE - (2 TYPOs) + 1 IF + 1 ELSE = polymorphism ?

This compiles and runs. (when it's not abbreviated)
is this considered polymorphism ??
```java
public class Shape
{
    string color;
}

public class Rectangle extends Shape { }
public class Triangle extends Shape { }
public class Circle extends Shape { }

extern void object::Poly_What??!!!( )
{
    Shape[] shapes;

    shapes[0] = new Rectangle("red", 4, 5);
    shapes[1] = new Triangle("green", 3, 4);
    shapes[2] = new Circle("blue", 2);

    Shape a_shape = new Triangle("yellow", 1, 2); // <--methods, vars. just works. fixed

    shapes[3] = a_shape;

    a_shape.What_Am_I() ; // Pointer To Triangle(....color="yellow"....

    message(""+shapes[2]); // Pointer To Circle(radius=2 ) extends Shape (color=blue )

    radar(); will not output to subclass of object anymore:  fixed.

    // none of these should compile/run, and they don't.
    Triangle t ;// = new Shape(...); //compile error "wrong type..." compiled before: fixed
 
    // t = new Shape(...); // <----------runtime error "wrong type..."  :)

    // t = new Circle(...); // <--- no compile, never did   
  
    //Rectangle r = new Triangle(...);//  ^^ same ^^

}
```
I have a mostly complete fix for this. I'll check back later.

Otaka:
Can you make a build with all these changes, I want to test it

melex750:
There are still problems with this issue. I will post what i have found soon.
I have no idea how to make a build, sorry. I compile the game myself on linux.

For the record: (and for anyone who reads this.)
I am only attempting to research these issues, and
my contributions should not be considered actual fixes.

krzys-h:
> I am only attempting to research these issues, and
> my contributions should not be considered actual fixes.

But still, they may help to find an actual fix, so it would be nice if you published them.

melex750:
I have already tested what I have with the latest pull request.
I would be glad to wait and refer to those files.

melex750:
While it is possible to compile examples of polymorphism in CBOT
its functionality has not been fully implemented. But it could be done.

Relevant info about source code related to this issue:
https://www.dropbox.com/s/nwqtdnitvpe4phf/PolyNotes.cpp?dl=1

[...]

Issue #433, opened 2015-01-30
Circular class references (diffrent types) in CBot


krzys-h:
First, make a program with a class like this:
```C++
public class TestClass {
    int test;
}
```
Click OK in the editor and start the program once. Then, edit the program to look like this:
```C++
public class OtherClass {
    TestClass testclass;
}

public class TestClass {
    int test;
    OtherClass otherclass;
}
```
Everything compiles properly and works as expected.

Now, restart the game and try compiling the second program again. The game will refuse to compile it now, telling you that it doesn't know what TestClass is (in line 2).

melex750:
This also happens when you type:
 ```java
public class OtherClass extends TestClass
```
A temporary workaround, short of moving TestClass above OtherClass
is to //comment out the offending instructions, click compile, uncomment
then it compiles ok. I may be easier to move TestClass in many cases.

This is due to being compiled from top to bottom,  Compile1 attempts to
simulate forward declarations, but fails because it also parses variables.

Functions also have a Compile1 for this purpose.

I ran Doxygen with GraphVis' Dot utility on the CBot Library.

This is one of many nice flowcharts showing how this works.
https://www.dropbox.com/s/uksguc8ar9fbwzc/CBotFunction.png

cbot-coder:
I am still seeing something like this behavior, where if I have a class hierarchy, sometimes the subclasses fail to compile; workaround is to comment them out, then compile, then uncomment them and they compile. It seems the first time around it doesn't know about the parent class, but after commenting out child, it compiles parent class, and knows about it, so uncommenting child class now works. It's not just a simple top-down ordering issue though, because some classes work fine even if parent is declared below the child.

PR #807, opened 2016-07-31
Add Full Support for Polymorphism

### LINKED ISSUE #540: CBot class inheritance
I have tested colobot version 0.1.5.0.
I thought that if cbot has classes, that means that should be inheritance of the classes, but I am not sure if it still in work or it is a bug, or I am doing something wrong because it does not documented.
I am trying the following code:

extern void object::New() {
	B b=new B();
	b.test();
}

public class A {
       public int baseMethod(){
		message("Base method");
		return 2;
	}
}

public class B extends A{
	public void test(){
		message("Child method");
		int returnValue=baseMethod();
		message("After baseMethod executing");
		message("Return value "+returnValue);
	}
}

The code is compiled without any errors but when it runs it shows the messages:
"Child method"
"After baseMethod executing"
"Variable not initialized"

Expected result:

"Child method"
"Base method"
"After baseMethod executing"
"Return value 2"

CyborgMaster:
This should resolve #540 and #683.  

Inheritance and polymorphism were mostly there.  I cleaned up a couple of compile time checks that prevented things like `Base b = new Child()` from compiling (even though the engine would execute it fine).  

I fixed the issue with a grandchild not being able to call a grandparent method.  Turns out `CBotClass` was only doing one level of parent checking.  Now it works with recursion to handle any depth.

Fix the weird assignment `Child c = new Base()`, which used to compile and actually kind of run, but would behave very strangely.  Now it throws the proper `Wrong type for assignment` error.

Dynamic Dispatch (virtual functions) now work correctly.  `Base b = new Child(); b.overriden_method();` now properly calls the child class method instead of the base class one.  this required calling methods based off the instance type instead of the variable type in `CBotInstrMethode`.  Which then also required improving the way the `super` pointer relates to parent pointers in `CBotVarClass` (`m_pParent` is now a pointer to the parent instance instead of the parent instance itself so `super` can clone the pointer and so garbage collection works properly).

I believe I have followed all the coding conventions and didn't break anything, but I haven't figured out how to run the linter or test suite yet, so I'm not sure.  I don't have time tonight to figure them out.  Hopefully I'll be able to get them running tomorrow, but I wanted to put this up here in case anyone else wanted to look at / test / comment on my approach.  

In short, the follow code now does what you would think it does: 

``` cpp
public class Base {
	public void foo() {
		message("foo");
	}

	public void greet(string name) {
		message("Hi " + name);
	}
}

public class Child extends Base {
	public void greet(string name) {
		super.greet("AWESOME " + name);
	}
}

public class Grandchild extends Child {
	public void greet(string name) {
		super.greet("lame " + name);
	}
}

extern void test() {
	Grandchild c = new Grandchild();
	Base b = c;  // compiles and runs correctly 
	c.greet("jeremy");  // outputs "Hi AWESOME lame jeremy"
	c.foo();  // calls grandparent method
	b.greet("jeremy2");  // outputs "Hi AWESOME lame jeremy2"
	b.foo();

        // these polymorphic assignments compile and run correctly
	Base b2;
	Grandchild g;
	b2 = new Grandchild();
	g = b2;

	Base b3 = new Child();
	Grandchild c3 = b3;  // Fails with "Wrong type for assignment"

	Child c4 = new Base();  // Fails with "Wrong type for assignment"
}
```

Let me know what you think. 

Notes to self for tomorrow: 
+ Explain how down casting works in CBot and how that affects compile time vs run time checks for polymorphic assignments.  
+ Figure out constructor / destructor inheritance

Above notes explained in a further comment on this PR [here](https://github.com/colobot/colobot/pull/807#issuecomment-236444003)

### DIFF
--- a/src/CBot/CBotClass.cpp
+++ b/src/CBot/CBotClass.cpp
@@ -354,9 +354,7 @@ bool CBotClass::ExecuteMethode(long& nIdent,
 
     if (m_parent != nullptr)
     {
-        ret = m_parent->m_pCalls->DoCall(name, pThis, ppParams, pResult, pStack, pToken);
-        if (ret >= 0) return ret;
-        ret = m_parent->m_pMethod->DoCall(nIdent, name, pThis, ppParams, pStack, pToken, m_parent);
+        ret = m_parent->ExecuteMethode(nIdent, name, pThis, ppParams, pResult, pStack, pToken);
     }
     return ret;
 }
--- a/src/CBot/CBotInstr/CBotDefClass.cpp
+++ b/src/CBot/CBotInstr/CBotDefClass.cpp
@@ -174,7 +174,8 @@ CBotInstr* CBotDefClass::Compile(CBotToken* &p, CBotCStack* pStack, CBotClass* p
             CBotClass* result = pStk->GetClass();
             if ( !pStk->GetTypResult(CBotVar::GetTypeMode::CLASS_AS_POINTER).Eq(CBotTypNullPointer) &&
                ( !pStk->GetTypResult(CBotVar::GetTypeMode::CLASS_AS_POINTER).Eq(CBotTypPointer) ||
-                 ( result != nullptr && !result->IsChildOf(pClass) )))     // type compatible ?
+                 ( result != nullptr && !(pClass->IsChildOf(result) ||
+                                          result->IsChildOf(pClass)))))     // type compatible ?
             {
                 pStk->SetError(CBotErrBadType1, p->GetStart());
                 goto error;
@@ -268,9 +269,10 @@ bool CBotDefClass::Execute(CBotStack* &pj)
             // evaluates the expression for the assignment
             if (!m_expr->Execute(pile)) return false;
 
+            CBotVar* pv = pile->GetVar();
+
             if ( bIntrincic )
             {
-                CBotVar*    pv = pile->GetVar();
                 if ( pv == nullptr || pv->GetPointer() == nullptr )
                 {
                     pile->SetError(CBotErrNull, &m_token);
@@ -280,8 +282,15 @@ bool CBotDefClass::Execute(CBotStack* &pj)
             }
             else
             {
+                CBotClass* c = pv->GetTypResult().GetClass();
+                if ( !c->IsChildOf(pClass))
+                {
+                    pile->SetError(CBotErrBadType1, &m_token);
+                    return pj->Return(pile);
+                }
+
                 CBotVarClass* pInstance;
-                pInstance = (static_cast<CBotVarPointer*>(pile->GetVar()))->GetPointer();    // value for the assignment
+                pInstance = (static_cast<CBotVarPointer*>(pv))->GetPointer();    // value for the assignment
                 pThis->SetPointer(pInstance);
             }
             pThis->SetInit(CBotVar::InitType::DEF);
--- a/src/CBot/CBotInstr/CBotDefClass.cpp
+++ b/src/CBot/CBotInstr/CBotDefClass.cpp
@@ -174,7 +174,7 @@ CBotInstr* CBotDefClass::Compile(CBotToken* &p, CBotCStack* pStack, CBotClass* p
             CBotClass* result = pStk->GetClass();
             if ( !pStk->GetTypResult(CBotVar::GetTypeMode::CLASS_AS_POINTER).Eq(CBotTypNullPointer) &&
                ( !pStk->GetTypResult(CBotVar::GetTypeMode::CLASS_AS_POINTER).Eq(CBotTypPointer) ||
-                 ( result != nullptr && !pClass->IsChildOf(result) )))     // type compatible ?
+                 ( result != nullptr && !result->IsChildOf(pClass) )))     // type compatible ?
             {
                 pStk->SetError(CBotErrBadType1, p->GetStart());
                 goto error;
--- a/src/CBot/CBotInstr/CBotDefClass.cpp
+++ b/src/CBot/CBotInstr/CBotDefClass.cpp
@@ -282,11 +282,14 @@ bool CBotDefClass::Execute(CBotStack* &pj)
             }
             else
             {
-                CBotClass* c = pv->GetTypResult().GetClass();
-                if ( !c->IsChildOf(pClass))
+                if ( !(pv == nullptr || pv->GetPointer() == nullptr) )
                 {
-                    pile->SetError(CBotErrBadType1, &m_token);
-                    return pj->Return(pile);
+                    CBotClass* c = pv->GetTypResult().GetClass();
+                    if ( !c->IsChildOf(pClass))
+                    {
+                        pile->SetError(CBotErrBadType1, &m_token);
+                        return pj->Return(pile);
+                    }
                 }
 
                 CBotVarClass* pInstance;
--- a/src/CBot/CBotInstr/CBotExprVar.cpp
+++ b/src/CBot/CBotInstr/CBotExprVar.cpp
@@ -118,10 +118,18 @@ CBotInstr* CBotExprVar::Compile(CBotToken*& p, CBotCStack* pStack, CBotVar::Prot
                     if (IsOfType(p, ID_DOT))
                     {
                         CBotToken* pp = p;
+                        CBotToken* nameStart = p;
 
-                        if (p->GetType() == TokenTypVar)    // must be a name
+                        // After a dot, names can start with a ~ in order to
+                        // facilitate calling super desconstructors
+                        if (p->GetType() == ID_NOT)
                         {
-                            if (p->GetNext()->GetType() == ID_OPENPAR)  // a method call?
+                            nameStart = p->GetNext();
+                        }
+
+                        if (nameStart->GetType() == TokenTypVar)    // must be a name
+                        {
+                            if (nameStart->GetNext()->GetType() == ID_OPENPAR)  // a method call?
                             {
                                 CBotInstr* i = CBotInstrMethode::Compile(p, pStk, var);
                                 if (!pStk->IsOk()) goto err;
--- a/src/CBot/CBotInstr/CBotExprVar.cpp
+++ b/src/CBot/CBotInstr/CBotExprVar.cpp
@@ -121,7 +121,7 @@ CBotInstr* CBotExprVar::Compile(CBotToken*& p, CBotCStack* pStack, CBotVar::Prot
                         CBotToken* nameStart = p;
 
                         // After a dot, names can start with a ~ in order to
-                        // facilitate calling super desconstructors
+                        // facilitate calling super destructors.
                         if (p->GetType() == ID_NOT)
                         {
                             nameStart = p->GetNext();
--- a/src/CBot/CBotInstr/CBotFunction.cpp
+++ b/src/CBot/CBotInstr/CBotFunction.cpp
@@ -34,6 +34,7 @@
 #include "CBot/CBotUtils.h"
 
 #include "CBot/CBotVar/CBotVar.h"
+#include "CBot/CBotVar/CBotVarClass.h"
 
 #include <cassert>
 #include <sstream>
--- a/src/CBot/CBotInstr/CBotFunction.cpp
+++ b/src/CBot/CBotInstr/CBotFunction.cpp
@@ -34,9 +34,13 @@
 #include "CBot/CBotUtils.h"
 
 #include "CBot/CBotVar/CBotVar.h"
+#include "CBot/CBotVar/CBotVarClass.h"
 
 #include <cassert>
 #include <sstream>
+
+#include <iostream>
+using namespace std;
 
 namespace CBot
 {
@@ -752,8 +756,16 @@ int CBotFunction::DoCall(long& nIdent, const std::string& name, CBotVar* pThis, 
             {
                 // sets the variable "super" on the stack
                 CBotVar* psuper = CBotVar::Create("super", CBotTypNullPointer);
-                psuper->Copy(pThis, false); // in fact identical to "this"
+                // psuper->Copy(pThis, false); // in fact identical to "this"
+                // cout << "here2" << endl;
+                // cout << typeid(*pThis).name() << endl;
+                // cout << pThis->GetClass()->GetName() << " " <<
+                //     dynamic_cast<CBotVarClass*>(pThis->GetPointer())->
+                //         m_pParent->GetClass()->GetName() << endl;
+                // cout << pThis->GetClass()->GetParent()->GetName() << endl;
+                psuper->Copy(pThis->GetPointer()->GetParent(), false);
                 psuper->SetUniqNum(-3);     // special value
+                //psuper->SetClass(pClass);
                 pStk->AddVar(psuper);
             }
             // initializes the variables as parameters
--- a/src/CBot/CBotInstr/CBotFunction.cpp
+++ b/src/CBot/CBotInstr/CBotFunction.cpp
@@ -34,13 +34,9 @@
 #include "CBot/CBotUtils.h"
 
 #include "CBot/CBotVar/CBotVar.h"
-#include "CBot/CBotVar/CBotVarClass.h"
 
 #include <cassert>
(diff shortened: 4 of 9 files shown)


krzys-h:
The tests segfault in CBotUT.ToString

CyborgMaster:
I'm sure can fix that. Are there instructions anywhere for how to run the test suite?

krzys-h:
Enable building tests:
```
$ cmake -DTESTS=ON ..
```
And run them:
```
$ make test
```

CyborgMaster:
Awesome.  I'm on it.  Might I suggest you add that info to `CONTRIBUTING.md` so new people can find it easier.  You also might want to include instructions on how to run the linter (which I also still need to figure out).

krzys-h:
Linter is fully documented in it's own README: https://github.com/colobot/colobot-lint/blob/master/README.md

CyborgMaster:
Oh that's right.  It's in a separate repository. I forgot that.  My apologies.

CyborgMaster:
So I'm getting the following output, which is different than what I see on Jenkins:

```
Test project /Users/jeremy/projects/colobot/build
    Start 1: colobot_ut
1/1 Test #1: colobot_ut .......................***Exception: SegFault  1.74 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   1.74 sec

The following tests FAILED:
	  1 - colobot_ut (SEGFAULT)
Errors while running CTest
make: *** [test] Error 8
```

It doesn't give me any line numbers or anything.  Any ideas?

krzys-h:
Try to run the ./colobot_ut file directly (it should be in the build directory). I forgot ```make test``` does that.

CyborgMaster:
Yup, just discovered that when running `./colobot_ut` instead of `make test`, then I get the expected tests and a SEGFAULT in the ToString test.  Now I'm off to solve the failure.  Thanks!

CyborgMaster:
Alright.  All the tests pass!  I'm going to now go figure out the linter and make sure that passes too.

CyborgMaster:
So, the linter looks like it might be a beast to set up.  I don't think I have time for that today.  Does the Jenkins build run it?

CyborgMaster:
Here are the explanations for the notes I mentioned in my original post.

### Downcasting
So one of the things that CBot does not have is any sort of casting `(int)`, `(Child)`, etc.  And I think that is perfectly fine.  What it does do is make compile time checking for downcasts different.

When you have code like this:
``` cpp
Base b = new Child();
Child c = b;
```
There is no good way at compile time to ensure that `b` is an instance of `Child`.  So right now the only compile time check is that `Base` and `Child` are in the same class tree, and then at runtime there is a check to make sure that the types are compatible (not assigning a parent class to a child pointer).  This was already working for assignments, just not for initializers, so I copied the logic from `CBotExpression` and `CBotLeftExpr` for the compile time and run time checks.

### Contructor inheritance
There is no implicit constructor inheritance in CBot, which I think is fine.  If there is a Base constructor and you want to have it called in the Child constructor, you have to specifically call it with `super` like so:
``` cpp
public class Base {
	void Base(int p1) {
		// Cool stuff here
	}
}

public class Child extends Base {
	void Child(int p1, int p2) {
		super.Base(p1);
		// More cool stuff
	}
}
```

This does mean that if `Child` has no constructor than the `Base` constructor isn't called either.  C++ has the concept of default constructors and if you don't explicitly call a base constructor it attempts to call the no-arg constructor.  I'm totally fine if that kind of magic doesn't happen in CBot, I would rather have it be explicit.  

Deconstructors are another matter.  Because they can't take parameters, it should be possible for the engine to automatically call the parent deconstrctor.  Also because the method starts with a '~', the previous super syntax (`super.~Base()`) doesn't work.  I'm not sure if we should try and make calling the parent deconstructor automatic, or get the super syntax to work.  I think automatic calling would be easy to implement.  I might try later today to give it a shot.

CyborgMaster:
I've now also added the ability to call super destructors.  I waffled back and forth between allowing a explicit super call and having them called automatically.  I ended up settling on explicit to match the explicit super constructor calls.  That way its consistent.  So now the following works:

``` cpp
public class Base {
	void ~Base() {
		message("base deconstructor");
	}
}

public class Child extends Base {	
	void ~Child() {
		message("child desconstructor");
		super.~Base();
	}
}
```

melex750:
Man you've been busy, I can tell you worked hard on this. :tada::+1:
I believe you're on the right track and I'd like to help.
I've been working on this too.

-----------------------
The only real problems I see with this PR, are the **changes to "super"** which **broke "this"**.
* A**```super.method();```**call, should **retain** the **context** of **"this"**.
* **"this"** should be the only main context within a single object.
* Basically, **"this"** should always point to the same object, because there is only one object.

Detailed explanation below, if it is needed.
I know somebody will say if I got anything wrong, and please do.
**Code example:**
```c++
public class Base {
	int i = 123;
	////////////////////////////
	void testThis() {
		message(""+this); " <-- "this" is not the child when called with "super""
		testOverride();        " but it was, and I think it should be "
	}
	////////////////////////////
	void testOverride() {
		message("Base");
	}
	////////////////////////////
	Base GetThis(){return this;}
}
public class Child extends Base {
	int j = 246;
	////////////////////////////
	void test() {
		   " When Child calls a method in Base,"this" should"
		     " still be the current object which Is-A child."
		super.testThis(); // <------ these have different outputs,
		/*this.*/testThis(); // <--/ but they call the same method.
	}
	////////////////////////////
	void testOverride() {
		message("Child");
	}
	////////////////////////////
	Child GetThis(){return super.GetThis();}
}
extern void object::New() {
	Child c();
	c.test();
	if (c != c.GetThis()) throw 1;
}
```
**Obligatory wall of text:**

1. As I understand it, from java documentation & tutorials  and from OOP concepts in general,
 **super**  is used to **access** overridden/hidden **members**, but **in the context** of the **current object**.

2. In CBot, the hidden parent instance exists to contain instance variables,
  and for reference-of-type to access inherited/overridden methods.

3. In CBot runtime, **super** is a copy of **this** for convenience.
  **super** as a pointer to the **hidden parent instance**
  isn't necessary, and also **breaks "this"**.

4. **super is not really supposed to be a pointer-to-something.**
  In java it's a keyword to tell compiler/runtime to find
  a member starting with inherited members from the closest relative.
Many people believe **java's super** is a reference variable,
but this not the case, it's just a keyword.

5. **```return super;```** should not compile and certainly not return a different object.

6. I think the "parent instance", isn't even necessary and could be
 removed at some point, but it works for now.
  In fact the concept of a "parent instance" doesn't seem to exist
  in any real programming language and breaks a few OOP rules too.

7.    **"There is no implicit constructor inheritance in CBot,..."**

  They're not inherited in java either, but **java inserts** an
  **implicit```super();```call** if one is not found on the
  first line of a subclass constructor.  There are also **explicit**
**```super();```**and**```this();```**constructor calls with or w/out args.
  This is **actually** partially **implemented** in CBot already. **;)**

8. I think parent class destructor should be called automatically,
   but calling it manually is better than not at all, or
   just do all the clean-up in a subclass destructor.

9. There are **other problems** with inheritance **not addressed** in this PR,
  the fixes for which don't conflict with what's been done here.

  Other than the "super" thing, these fixes look ok-ish to me.
  They're a lot like some I've been testing for about a month now.

10. I'd like to add to this effort, but **the changes to "super" are not good** IMO.
  There is an easier way to do what that fixed anyway.
 
11. This definitely needs some heavy testing, and defining how it *should* work.
  I've been testing inheritance/polymorphism with a 5kb CBot program,
  and TBH, it could use some more tests.

12.  So how can I help??.

CyborgMaster:
@melex750, great feedback!

I didn't think about how these changes affected the this pointer, you make a good point.  My goal wasn't actually to modify how super works at all.  The main goal was dynamic dispatch (calling a method based on the run-time instance type not the compile-time variable type).  This was mainly accomplished by changing `pClass = CBotClass::Find(m_className);` to `pClass = pThis->GetClass();` in `CBotInstrMethode`.  This works great, but prompted the changes in the way super was handled.  Super used to be a pointer to this, with a compile time variable type of the parent class, which, due to static dispatch, caused `super.method()` to work.  In order to get it to work with dynamic dispatch, and keep most of the engine working the same way, it had to point to the "parent" instance.  

But you're right, the weirdness it causes with `this` in parent classes is not good, and needs to be figured out.  Do you have any ideas on how this could be accomplished?  I think I have one quick inkling that I might try this evening. 

I think the biggest help you could provide would be with the tests you were talking about?  Can I somehow get ahold of your test file?  Maybe even add it to the unit tests?

To respond directly to some of your comments:

3.  `super` wasn't a direct copy of `this`.  They pointed to the same instance at run time, but the compile time type of super was the parent class, which was how static dispatch was working in the first place.  If it was an exact copy of this, the following would be an infinite loop:
``` cpp
public class Child extends Base {
  public overriden() {
    super.overriden();
  }
}
```
because that would be the same as calling `this.overriden`, which would be infinite recursion.  

4. I agree that super really shouldn't be a pointer.  However, that is the way it currently works in the engine and it wasn't clear how to rip it out

5.  `return super;` Currently does compile and run CBot, right now it returns `this`, which really isn't any more weird than returning the parent instance (neither of which anyone should normally be doing).

7.  When I said "constructor inheritance," I meant exactly what you are saying, the implicit super call inserted by java and c++ if you don't call it manually.  It seems like this implicit call would be difficult to implement in CBot because you have to deal with default constructors, and recognize if the super constructor has been called while compiling a child constructor.  Not impossible, just difficult.

8.  I'm totally up to having the parent destructors called automatically.  That wouldn't be hard at all, and I can change it in a few minutes.  I was just having a hard time stomaching the difference between "Parent constructors are *not* called automatically, you must manually call them", and "Parent destructors *are* called automatically, you cannot call them manually".  In C++ or Java, they are both called implicitly (unless you manually call a super constructor of course), and having one called implicitly and one explicitly in CBot seemed very strange.  It wouldn't be hard to change, I'm just not sure what would be easiest to understand for a new person learning programming, and to me, in that environment, the less magic the better.

9.  We should make a list of the other problems in an issue somewhere (start by pasting them here?), maybe they would be easy to implement in this PR?

11.  I totally agree about the testing.  Let's do it!  Can you send me the 5kb CBot program and I'll add it to the unit tests? Or you could do it yourself and open another PR?

krzys-h:
> I think the biggest help you could provide would be with the tests you were talking about? Can I somehow get ahold of your test file? Maybe even add it to the unit tests?

A big :+1: from me for any new unit tests. I wrote a bunch of them myself, but they still don't cover all functionality of the language.

CyborgMaster:
I have fixed the implementation of `this` so it now always points to the original object, even in `super` calls.  I also added some tests to hit a few bits of polymorphism and dynamic dispatch (although I'm sure not as thorough as @melex750's test file; I would still love to get ahold of it).   As one of the tests, I specifically added what he proposed in his comment:

``` cpp
TEST_F(CBotUT, ThisPointerInsideSuperCall)
{
    ExecuteTest(
        "public class Base {\n"
        "    Base GetThis() { return this; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    Base GetThis() { return super.GetThis(); }\n"
        "}\n"
        "extern void test() {\n"
        "    Child c();\n"
        "    ASSERT(c == c.GetThis());\n"
        "}\n"
    );
}
```

If anyone else can find issues with this PR, let me know and I'll be happy to fix them. :-)

krzys-h:
Looks good to me! I tested some things myself and it seems everything works fine. I'll wait for confirmation from @melex750 before merging though.

One minor thing I noticed while testing, these statements still give you access to the "parent instance", but I don't think this is a big deal for now and can be fixed in a future:
```C++
public class TestChild extends TestParent {
	TestParent test() {
		TestParent p = super;
		message(""+super);
		return super;
	}
}
```

I really appreciate any contributions to CBot, most developers (including me) gave up long ago saying only a full rewrite could allow us to do anything with this mess.

After this gets merged, we'll need to update ingame docs. ping @MrSimbax

I also ran colobot-lint on your code, I'll make sure to configure Jenkins to check this automatically soon:
```
[style] [code block placement] /home/krzys_h/Pulpit/colobot/src/CBot/CBotVar/CBotVarClass.cpp:87 Body of declaration or statement begins in a style that is not allowed
[style] [code block placement] /home/krzys_h/Pulpit/colobot/src/CBot/CBotInstr/CBotFunction.cpp:751 Body of declaration or statement begins in a style that is not allowed
[style] [code block placement] /home/krzys_h/Pulpit/colobot/src/CBot/CBotInstr/CBotInstrUtils.cpp:157 Body of declaration or statement begins in a style that is not allowed
```

CyborgMaster:
@krzys-h, thanks!  

About returning the super pointer.  I agree it's a little weird, but I'm not sure how it could be fixed with the current engine implementation.  The super pointer is implemented as a real pointer as far as the function is concerned, so there really isn't a way to prevent returning it.  We might be able to block it in the compile stage, even though the engine would still support it.  

You're welcome on the CBot contributions.  I agree the code base is royal mess (lots of inconsistencies, "magic" calls, no comments, etc), and it really should be thrown out and done over (which is why embedding another language like Lua or Python sounds really interesting).  In the meantime, I'm starting to get my legs under me inside the CBot code and will try and fix things as they really annoying me in game (for example, #433 is really bugging me right now).

I'll try and fix the lint errors tonight after work, but it might be tomorrow before I get to them.

CyborgMaster:
That should take of the linter errors.  Anything else that needs to be done before we merge this?

melex750:
I don't know,  testing how these changes affect
saving/reloading a game with a running program,
is difficult and time consuming.

It would be nice if we could convert SaveState and
RestoreState to use streams instead of ```FILE*```, so we
can simulate saving/reloading in unit tests.
Or Maybe use : ```FILE* f = fmemopen(buf, size, mode);``` (linux only, I think)

Thats apart from the fact that these changes way
over-complicate something so [simple as this](https://github.com/colobot/colobot/commit/c9c02f5461d62193b90fce4a80e7f7e4e90c28f2#diff-5dc8ecaa603b53ac17b74cabf268b831R151).
An A+ for effort, but this broke more things than it fixed.
I hope I can be a bigger help next time.

I made some unit tests for inheritance.

This [#809](https://github.com/colobot/colobot/pull/809) is what I've been testing for a 
couple of weeks while I was working on this #808.

CyborgMaster:
@melex750, impressive work!  I'll try and take a look at this today after work.  Assuming it takes care of all my use cases, we should drop my PR in favor of yours (I'll report back when I've confirmed it does).  

It's obvious you've got a lot more experience with the CBot engine than I do.  Sorry if I over complicated things in a few places. That was the best I could figure out with only ~6 hours over one weekend, and it looks like you've been able to absorb a lot more than me about how the engine works.  

I'm all for converting `SaveState` and `RestoreState`, those two ideas were things that I hadn't even broached yet in my changes and having some tests for them would have pointed out where I broke things.  

Thanks for your help, I'll let you know what I find out this evening.

Erihel:
Excellent work guys. Colobot project needs more developers and you are touching stuff that was always low priority due to people being afraid to break it. This PR has a lot of constructive feedback which is always appreciated.

Saving and loading running programmes from saves might be too much. I mean there was some talk about adding others languages into Colobot. Providing support for running programs in saves for many languages could be difficult or time consuming as @melex750 mentioned.

Colobot uses physfs for accessing files. CBot is one of places it's not supported yet (if not the only place). There are some wrappers for physfs https://github.com/colobot/colobot/tree/master/src/common/resources that might be useful.

CyborgMaster:
@Erihel Thanks! 

I'm pretty sure saving and loading running programs already works for CBot.  The only issue with this PR was how the new functionality affected that save and restore process.

krzys-h:
SaveState and RestoreState should definitely be refactored to streams, I wanted to do that long ago but I gave up when I saw how much code that is. And we should definitely get rid of this terrible hack: https://github.com/colobot/colobot/blob/master/src/CBot/CBotFileUtils.cpp#L32-L64

@Erihel I think the only thing except for CBot that doesn't use physfs yet is saving screenshots.

krzys-h:
Replaced by #809

Issue #384, opened 2014-11-11
Fluent syntax sugar


tr00per:
Right now you can't write:

    goto(radar(SpaceShip).position);

which should be equivalent to

    object target = radar(SpaceShip);
    goto(target.position);

Right now it triggers a compile-time error:

> Brak nawiasu zamykającego (Missing closing bracket)

It might be convenient to add such syntax sugar.

PR #808, opened 2016-08-04
Syntax sugar:  Accessing members of returned objects

melex750:
**#384**
```c++
extern void object::New()
{
    goto( radar(Titanium).position );

    bool b = radar(BotFactory).busy();
}
```
**This also allows [method chaining](https://en.wikipedia.org/wiki/Method_chaining#Java):**
```c++
    AClass ac();
     " provided that each chained method returns "this"... "

    ac.SetInt(4).SetBool(true).SetFloat(3.14).SetString("str");
```
And accessing array elements if that is wanted, it can be enabled with one line of code.
object o = radarall(PowerPlant)[0]; <-- element access(currently disabled)

### DIFF
--- a/src/CBot/CBotInstr/CBotExprRetVar.cpp
+++ b/src/CBot/CBotInstr/CBotExprRetVar.cpp
@@ -135,9 +135,14 @@ bool CBotExprRetVar::Execute(CBotStack* &pj)
     {
         pVar = pj->GetVar();
         pVar->Update(pj->GetUserPtr());
+        if (pVar->GetType(CBotVar::GetTypeMode::CLASS_AS_POINTER) == CBotTypNullPointer)
+        {
+            pile1->SetError(CBotErrNull, &m_token);
+            return pj->Return(pile1);
+        }
 
         if ( !m_next3->ExecuteVar(pVar, pile, &m_token, true, false) )
-                return false;
+            return false;
 
         if (pVar)
             pile1->SetCopyVar(pVar);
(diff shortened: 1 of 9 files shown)


krzys-h:
Could you please add some unit tests for this?

melex750:
I'll add some more tests a bit later.

krzys-h:
That's a lot more tests than I expected, thanks. We'll probably have to start splitting them into separate files soon, I'm starting to get lost in the test file already :stuck_out_tongue:

Some problems reported by colobot-lint (I should really make Jenkins run this automatically):
```
[style] [license header] /home/krzys_h/Pulpit/colobot/src/CBot/CBotInstr/CBotExprRetVar.cpp:1 File doesn't have proper license header; expected line was '/*'
[style] [license header] /home/krzys_h/Pulpit/colobot/src/CBot/CBotInstr/CBotExprRetVar.h:1 File doesn't have proper license header; expected line was '/*'
```

melex750:
Yeah, adding more tests is probably not necessary, the current ones test the feature completely.
I wasn't sure if I should add the license header, I'll do that in a minute.
I thought I could add a few comments to the code today, but... PBKAC ...writer's block.

melex750:
Another option to consider, is adding this after definition and "new" constructor calls.
```c++
AClass a = new AClass().WithFeatureA(1).WithFeatureB(2);
```

krzys-h:
@melex750 I think the syntax with constructor you mentioned should be supported. Do you want to add it in this PR or should I just merge it as it is?

melex750:
@krzys-h As is.
I think the other should be done a little differently.
Maybe methods only?

CyborgMaster:
I would love to have the constructor chaining work.  I have several places in the code where I would like to do things like 

``` cpp
if (new Base().checkForResource())
```
or stuff like that.

PR #809, opened 2016-08-04
Fixes and unit tests for inheritance

melex750:
The **unit tests** are somewhat big because they need
to test at least three levels of inheritance.

**Fixes these bugs:**

**Destructors** that are expected to be called,
**don't get called** after saving/reloading the game, 
when the game is saved with a running program.

**RestoreState** assigns wrong IDs to subclass instance variables.

**SegFault** when RestoreMethode doesn't restore a call to inherited method.
 Reproduce: save/reload the game while executing code inside the method.

**Inheritance #540 and #683:**

This doesn't work:    ```child.grandParentMethod();```

This doesn't compile: ```Base b = new SubClass();``` <-- bug
This does compile:    ```SubClass s = new Base();``` <-- bug

This currently compiles:
```c++
Base b;
b = new Subclass();
b.overriddenMethod(); <-- calls method in base (bug)
```
**At runtime: won't re-assign the pointer.**
```c++
b = new SubClass(); <------- unless the same type is already assigned.
b = new OtherSubClass(); <-- runtime: "Wrong type for the assignment"
b = new Base();     <------- runtime: "Wrong type for the assignment"
```
Implicit downcast was already implemented and works.

All of the above also applies to **arrays of classes**.

**Suggested improvement:**

Explicit cast operator and remove implicit downcast.
```c++
  Base base = new SubClass();
  SubClass s; 

  s = cast<SubClass>(base);
  cast<SubClass>(base).x = 4.2;
  cast<SubClass>(base).subclassMethod();
```


CyborgMaster:
I'm reading through your changes and comparing them to what I did.  I've had a question about the code base and I was wondering if you might be able to help me understand it better.   What is the different between a TypResult and a Type in CBot?  They seem to be similar, but different.

CyborgMaster:
You're code looks great.  I do think I have found one test case that passes on my PR but doesn't on your, although I'm not exactly sure why.

``` cpp
TEST_F(CBotUT, Temp3)
{
    ExecuteTest(
        "public class Base {\n"
        "    int foo() { return 1; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    int foo() { return 2; }\n"
        "}\n"
        "public int func(Base b) {\n"
        "    return b.foo();\n"
        "}\n"
        "extern void test() {\n"
        "    ASSERT(func(new Child()) == 2);\n"
        "}"
        );
}
```

CyborgMaster:
This also doesn't compile if I assign it to a variable of the correct type first:

``` cpp
extern void test() {
  Child c = new Child();
  ASSERT(func(c) == 2);
}
```

But if I assign it to a variable of the super type it does work:
``` cpp
extern void test() {
  Base b = new Child();
  ASSERT(func(b) == 2);
}
```

This might be related to the changes I made to `src/CBot/CBotInstr/CBotInstrUtils.cpp`,  should you add the same modifications on your PR?

CyborgMaster:
I've also got 4 or 5 other tests that could be added to help in the future.  How should I submit them to you for inclusion?  I could paste them here, or maybe submit a PR into your source branch?

melex750:
Yeah, I didn't do that cast the same way because I was considering another that would allow passing null also. [CBotTypResult](https://github.com/colobot/colobot/blob/master/src/CBot/CBotTypResult.h#L171) hold information about a variable, [CBotType](https://github.com/colobot/colobot/blob/master/src/CBot/CBotEnums.h#L35) is the type of value.

CyborgMaster:
I think I understand about CbotTypResult and CBotType.  Thanks.

What are your plans for including null?  Can we get them into this PR?  Is it something I could help with?

melex750:
It's an easy fix but I wonder what others think.

CyborgMaster:
If you put it up, either in this PR or another one, I'll give you my opinion.

krzys-h:
@CyborgMaster
CBotType is just a simple enum that contains the base type like: string, int, array, pointer.
CBotTypResult contains complete information about the type, like: array of ints (```int x[];```), array of strings limited to 4 values (```string x[4];```), pointer to class Y (```Y x;```), array of pointers to class Z (```Z x[];```) etc.
In some cases CBotTypResult may also store a compilation error code (from CBotError enum).
See http://compiled.colobot.info/job/colobot/job/colobot/job/dev/Doxygen/class_c_bot_1_1_c_bot_typ_result.html#details for examples.

The best way to add additional unit tests to this PR would be probably to send another PR to @melex750's repo, although I don't think I ever saw chaining PRs like that happen in practice ;)

I think I'll leave figuring out the best solution to you two as I have only a very general understanding of the CBot engine and it's hard for me to even compare the changes.

melex750:
On second thought, I don't think we should do any automatic casting when passing arguments,
at least for now, because it creates an ambiguous call.
Currently, there is nothing to resolve this.
```c++
public class BaseClass{}
public class SubClass extends BaseClass{}

void func(BaseClass bc){}
void func(SubClass sc){}

extern void Test()
{
    func(new SubClass());   which one will execute??
}

```

CyborgMaster:
@melex750 C++ does automatic up-casting of parameters, what does it do in a situation like that with an overloaded method? I'll see if I can run some tests this evening.

melex750:
I think CBot compiler should first search for a perfect match, then search for a match in each parent class type, if there is more than match throw a compile error.
This would be a bit more complex algorithm than the current one.

CyborgMaster:
I think it could get even more complicated that that.  What about runtime checks?

``` cpp
public class BaseClass{}
public class SubClass extends BaseClass{}

void func(SubClass sc){}

extern void Test()
{
    BaseClass b = new SubClass();
    func(b);   Should this work?
}
```

I guess the answer is that it shouldn't compile?  Does that conflict with the way dynamic dispatch works for  method calls `Base b = new Sub(); b.sub_method();`?

CyborgMaster:
I guess maybe we should do dynamic dispatch for method calls and static dispatch for function overloading?

krzys-h:
I think I once saw code that was supposed to find a best match for function when no exact match is found depending on the amount of upcasting/downcasting required. It probably needs to be updated to handle inheritance.
That code is here: https://github.com/colobot/colobot/blob/master/src/CBot/CBotInstr/CBotFunction.cpp#L469-L580

PS. Ignore the failing checks, something broke on Jenkins

CyborgMaster:
Sweet!  We should update that.  It would make some of my CBot code using inheritance a lot simpler, namely using a subclass as a function pointer (like they do in java). 

``` cpp
point findShipSpot() {
	object o = radar(SpaceShip);
	SpotChecker c = new ShipSpotChecker();
	return findSpot(c, o.position, 0, 1);
}

public class ShipSpotChecker extends SpotChecker {
	public bool check(point p) {
		object o = search(Any, p);
		return (distance2d(p, o.position) > 3);
	}
}
```

CyborgMaster:
It looks like implicit downcasting is broken.  The following test doesn't pass:

``` cpp
TEST_F(CBotUT, Downcasting)
{
    ExecuteTest(
        "public class Base {\n"
        "    int foo() { return 1; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    int subMethod() { return 2; }\n"
        "}\n"
        "extern void test() {\n"
        "    Base b = new Child();\n"
        "    Child c = b;\n"
        "    ASSERT(b.subMethod() == 2);\n"
        "}"
        );
}
```

CyborgMaster:
This does pass though, so it must be just when initializing a class var.

``` cpp
TEST_F(CBotUT, Downcasting)
{
    ExecuteTest(
        "public class Base {\n"
        "    int foo() { return 1; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    int subMethod() { return 2; }\n"
        "}\n"
        "extern void test() {\n"
        "    Base b = new Child();\n"
        "    Child c;\n"
        "    c = b;\n"
        "    ASSERT(c.subMethod() == 2);\n"
        "}"
        );
}
```

CyborgMaster:
The follow patch fixes it (it is similar to what I did on my branch).  What do you think?

``` diff
--- a/src/CBot/CBotInstr/CBotDefClass.cpp
+++ b/src/CBot/CBotInstr/CBotDefClass.cpp
@@ -174,7 +174,8 @@ CBotInstr* CBotDefClass::Compile(CBotToken* &p, CBotCStack* pStack, CBotClass* p
             CBotClass* result = pStk->GetClass();
             if ( !pStk->GetTypResult(CBotVar::GetTypeMode::CLASS_AS_POINTER).Eq(CBotTypNullPointer) &&
                ( !pStk->GetTypResult(CBotVar::GetTypeMode::CLASS_AS_POINTER).Eq(CBotTypPointer) ||
-                 ( result != nullptr && !result->IsChildOf(pClass) )))     // type compatible ?
+                 ( result != nullptr && !(pClass->IsChildOf(result) ||
+                                          result->IsChildOf(pClass)))))     // type compatible ?
             {
                 pStk->SetError(CBotErrBadType1, p->GetStart());
                 goto error;
@@ -268,9 +269,10 @@ bool CBotDefClass::Execute(CBotStack* &pj)
             // evaluates the expression for the assignment
             if (!m_expr->Execute(pile)) return false;
 
+            CBotVar* pv = pile->GetVar();
+
             if ( bIntrincic )
             {
-                CBotVar*    pv = pile->GetVar();
                 if ( pv == nullptr || pv->GetPointer() == nullptr )
                 {
                     pile->SetError(CBotErrNull, &m_token);
@@ -280,8 +282,17 @@ bool CBotDefClass::Execute(CBotStack* &pj)
             }
             else
             {
+                if ( !(pv == nullptr || pv->GetPointer() == nullptr) )
+                {
+                    if ( !pv->GetClass()->IsChildOf(pClass))
+                    {
+                        pile->SetError(CBotErrBadType1, &m_token);
+                        return pj->Return(pile);
+                    }
+                }
+
                 CBotVarClass* pInstance;
-                pInstance = (static_cast<CBotVarPointer*>(pile->GetVar()))->GetPointer();    // value for the assignment
+                pInstance = pile->GetVar()->GetPointer();    // value for the assignment
                 CBotTypResult type = pThis->GetTypResult();
                 pThis->SetPointer(pInstance);
                 pThis->SetType(type);        // keep pointer type
```

melex750:
Just my opinion, I propose we vote.

Implicit downcasting lets you write things like this:
SubClass s = new Base(); or funcNeedsSubClass( new Base() );
I think the compiler should laugh at such things.

A cast operator would help with casting.
This: ```s = cast<SubClass>(b);``` is easier to read than ```s = b;``` ,
also there can be```func( cast<SubClass>(b) );``` 
and the operator would have compile/runtime checking.

krzys-h:
I would rather not introduce a cast operator to CBot, the whole language is designed to be as simple as possible, that would overcomplicate things too much

CyborgMaster:
I agree with @krzys-h.  CBot is supposed to be a learning language, so the simpler the better.  Implicit downcasting is already implemented,  is works just fine for assignments like this 

``` cpp
Base b = new Child();
Child c;
c = b;
```

The patch I submitted just extends this to initialization as well so you can do this.

``` cpp
Base b = new Child();
Child c = b;
```

melex750:
I'm ok with no cast operator.

```funcNeedsSubClass( baseVar );```
This adds ```if..else, ->GetClass()->IsChildOf(...)``` overhead to every time you pass a pointer
even when upcasting, I think this should be up casting only, which is always safe. no need to check.

**Comment by CyborgMaster** [comment]
I'm fine if functions parameters are upcast only.  If you need to call a function with a downcast, you'll just have to use an intermediate child class variable.  I think that strikes a good balance.  Upcasting should always be implicit, and the only way to downcast would be to assign an intermediate variable.  Ex:

``` cpp
class Base {}
class Child extends Base {}
void needsChild(Child c) {}
Base b = new Child();
needsChild(b);  // Won't compile
Child c = b;  // Downcast assignment
needsChild(c) // This works
```

melex750:
@CyborgMaster , Your patch looks good.
I appreciate the input because, I am no expert.
I've had these questions for a while, thanks for taking the time.
If I'm unsure of something, I don't do it, I ask what others think.

Is downcasting used very often?
I was unsure about downcasting in the definition of a variable and
adding a little overhead when there may be no intent to downcast.
For example: inside of a (do, while, for) loop.

CyborgMaster:
If you're using polymorphism, then it can be used a lot.  The reason I found it is because I was trying to use it today.  I have code that looks like this:

``` cpp
ReserveResourceGoal g = subGoals[0];
bot.get(g.resource);
```

If we are going to support downcasting by assignment then we should support it for both normal assignment and variable initialization.  It would be very confusing if this worked (which it currently does):
``` cpp
Base b = new Child;
Child c;
c = b; 
```
But this did not (doesn't work currently, my patch fixes it):
``` cpp
Base b = new Child;
Child c = b; 
```
Anything that we support for assignment should also work for initialization.

melex750:
Applied @CyborgMaster 's patch with one small change.

I think that's the best I can do with the unit tests.

CyborgMaster:
@melex750, I finally found the difference between my patch and what you applied.  Looks great!  I'm glad you changed it.  

Here's all the tests that I've been using.  Your's probably covers, most if not all of this, I'm just including it incase you want to incorporate anything

``` cpp
TEST_F(CBotUT, ThisPointerInsideSuperCall)
{
    ExecuteTest(
        "public class Base {\n"
        "    Base GetThis() { return this; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    Base GetThis() { return super.GetThis(); }\n"
        "}\n"
        "extern void test() {\n"
        "    Child c();\n"
        "    ASSERT(c == c.GetThis());\n"
        "}\n"
    );
}

TEST_F(CBotUT, ThisPointerInsideSuperCallGrandparent)
{
    ExecuteTest(
        "public class Base {\n"
        "    Base GetThis() { return this; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    Base GetThis() { return super.GetThis(); }\n"
        "}\n"
        "public class Grandchild extends Child {\n"
        "    Base GetThis() { return super.GetThis(); }\n"
        "}\n"
        "extern void test() {\n"
        "    Grandchild g();\n"
        "    ASSERT(g == g.GetThis());\n"
        "}"
    );
}

TEST_F(CBotUT, DynamicDispatch)
{
    ExecuteTest(
        "public class Base {\n"
        "    int foo() { return 1; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    int foo() { return 2; }\n"
        "}\n"
        "extern void test() {\n"
        "    Base b();\n"
        "    ASSERT(b.foo() == 1);\n"
        "    Child c();\n"
        "    ASSERT(c.foo() == 2);\n"
        "    b = c;\n"
        "    ASSERT(b.foo() == 2);\n"
        "}"
    );
}

TEST_F(CBotUT, WrappedSuperMethod)
{
    ExecuteTest(
        "public class Base {\n"
        "    int wrapped() { this.foo() +  1; }\n"
        "    int foo() { throw 1; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    int foo() { return 2; }\n"
        "}\n"
        "extern void test() {\n"
        "    Child c();\n"
        "    ASSERT(c.wrapped() == 3);\n"
        "}"
    );
}

TEST_F(CBotUT, Temp)
{
    ExecuteTest(
        "public class Base {\n"
        "    int foo() { return 1; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    int foo() { return 2; }\n"
        "}\n"
        "public class Grandchild extends Child {\n"
        "    int foo() { return 3; }\n"
        "}\n"
        "extern void test() {\n"
        "    Base b = new Base();\n"
        "    ASSERT(b.foo() == 1);\n"
        "    Child c = new Child();\n"
        "    ASSERT(c.foo() == 2);\n"
        "    Grandchild g = new Grandchild();\n"
        "    ASSERT(g.foo() == 3);\n"
        "    b = c;\n"
        "    ASSERT(b.foo() == 2);\n"
        "    b = g;\n"
        "    ASSERT(b.foo() == 3);\n"
        "    c = b;\n"
        "    ASSERT(b.foo() == 3);\n"
        "}"
        );
}

TEST_F(CBotUT, Temp2)
{
    ExecuteTest(
        "public class Base {\n"
        "    int foo() { return 1; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    int foo() { return 2; }\n"
        "}\n"
        "public class OtherChild extends Base {\n"
        "    int foo() { return 3; }\n"
        "}\n"
        "extern void test() {\n"
        "    Base b = new Base();\n"
        "    ASSERT(b.foo() == 1);\n"
        "    Child c = new Child();\n"
        "    ASSERT(c.foo() == 2);\n"
        "    OtherChild o = new OtherChild();\n"
        "    ASSERT(o.foo() == 3);\n"
        "    b = c;\n"
        "    ASSERT(b.foo() == 2);\n"
        "    b = o;\n"
        "    ASSERT(b.foo() == 3);\n"
        "}"
        );
}

TEST_F(CBotUT, Temp3)
{
    ExecuteTest(
        "public class Base {\n"
        "    int foo() { return 1; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    int foo() { return 2; }\n"
        "}\n"
        "public int func(Base b) {\n"
        "    return b.foo();\n"
        "}\n"CO
        "extern void test() {\n"
        "    ASSERT(func(new Child()) == 2);\n"
        "}"
        );
}

TEST_F(CBotUT, Downcasting)
{
    ExecuteTest(
        "public class Base {\n"
        "    int foo() { return 1; }\n"
        "}\n"
        "public class Child extends Base {\n"
        "    int subMethod() { return 2; }\n"
        "}\n"
        "extern void test() {\n"
        "    Base b = new Child();\n"
        // "    Child c;\n"
        // "    c = b;\n"
        "    Child c = b;\n"
        "    ASSERT(c.subMethod() == 2);\n"
        "}"
        );
}
```

melex750:
Yep, the tests I did, cover pretty much all of those, and some other things that were fixed before.
I admit they need to be broken into smaller tests, but the test file too big already.
I'm not sure how the unit test file would be split into separate files, but i hope it happens soon.

P.S. I have some other bug fixes for classes, probably for a different PR though.

------------------------------------
@CyborgMaster , Have you tested these changes very much, in-game?
Some CBot programs that I have, using classes, inheritance, and linked-lists:
A*-like pathfinder
Code battle program
CBot Interpreter (very incomplete)
Whitespace Interpreter (done, needs error handling & debugged)

I will share them eventually, when they're more complete.
Some of them will be updated using polymorphism.

krzys-h:
> CBot Interpreter

CyborgMaster:
I've been writing CBot code off of this PR's changes (merged with your "sugar" pr as well) heavily for about a week now.  I've been working on a general goal based AI that can solve all of the campaign levels without any level specific logic.  It's been working great.  The only problem I have run into is the forward declaration of classes, but that's a different issue (#433).

CyborgMaster:
@melex750, What's a "Whitespace Interpreter"?

krzys-h:
@CyborgMaster https://en.wikipedia.org/wiki/Whitespace_(programming_language)

melex750:
I have a fix for "433".

CyborgMaster:
@krzys-h Wild!  That's a gnarly language.  

@melex750 put it up as another pull request!  

I think this PR is probably ready to be merged.  There might be other tweaks we can make, but we can do those as follow-on PRs as necessary.

melex750:
> put it up as another pull request!

@CyborgMaster  #814 :tada:

krzys-h:
Merged! :tada:

PR #814, opened 2016-08-09
Allow circular references from different classes

melex750:
Issue #433


CyborgMaster:
Awesome!  I'll test this out this evening.  

@melex750, you are a wizard!

krzys-h:
Could we get some unit tests for this too pleeeease? We should extend the test suite whenever possible.

melex750:
If I push unit tests, will it conflict?
Only two tests. I can paste them here if that's easier.

krzys-h:
@melex750 It shouldn't, even if it does I can fix it easily.

CyborgMaster:
I know this was already merged, so it's a bit of a moot point, but I can confirm that the forward declaration works great!  Yay for no more commenting lines out and back in every time I launch the program.

PR #817, opened 2016-08-13
radarall can return multiple objects with exactly the same distance

### LINKED ISSUE #816: radarall misses objects of equal distance
`radarall` uses a map of distance to object to store them in while searching, which used to work fine when you only needed to return one (because if two were exactly the same distance away, then ether one would be correct), but now that we have `radarall` in addition to `radar` this needs to be fixed to make sure it gets all the objects.  

This normally would be an extremely rare occurrence (trying to get two floating point distances to be exactly the same (especially with rounding errors) is almost impossible), except for at the very start of the level.  When the level first starts, I want to count the resources we brought with us on the spaceship.  At the beginning your robot is placed in one corner of the ship, and often 4 resources are placed in a square at the opposite corner.  This causes two of those resources to be exactly equidistant to your robot, so `radarall` returns 3, instead of 4.  

The layout ends up looking something like this:
```
o = PowerCell
x = Center of Spacehship
B = robot

o     o

o     o


            x



                     B   
```

The code responsible is in object_manager.cpp inside of `RadarAll`, line 285:
``` cpp
    std::map<float, CObject*> best;
```

I'll give upgrading it a shot and open a PR if I am successful.

CyborgMaster:
Simple needed to change a `map` to a `multimap` to allow for duplicates. Closes #816

### DIFF
--- a/src/object/object_manager.cpp
+++ b/src/object/object_manager.cpp
@@ -282,6 +282,8 @@ std::vector<CObject*> CObjectManager::RadarAll(CObject* pThis, Math::Vector this
     RadarFilter filter_flying = static_cast<RadarFilter>(filter & (FILTER_ONLYLANDING | FILTER_ONLYFLYING));
     RadarFilter filter_enemy = static_cast<RadarFilter>(filter & (FILTER_FRIENDLY | FILTER_ENEMY | FILTER_NEUTRAL));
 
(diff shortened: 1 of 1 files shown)


krzys-h:
Looks good :+1:, will merge when I have access to a computer to test

PR #819, opened 2016-08-14
Method chaining for constructor calls in CBOT

melex750:
```c++
AClass ac().SetA().SetB();

AClass a = new AClass().WithFeatureA().WithFeatureB();
```

### DIFF
--- a/src/CBot/CBotInstr/CBotDefClass.cpp
+++ b/src/CBot/CBotInstr/CBotDefClass.cpp
@@ -19,6 +19,7 @@
 
 #include "CBot/CBotInstr/CBotDefClass.h"
 
+#include "CBot/CBotInstr/CBotExprRetVar.h"
 #include "CBot/CBotInstr/CBotInstrUtils.h"
 
 #include "CBot/CBotInstr/CBotLeftExprVar.h"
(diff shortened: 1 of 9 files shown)


krzys-h:
I'm not sure if I like the first syntax. What would happen if you didn't return this in a method?
I think method chaining should work only in the second example.

And again, thanks for your awesome patience to work with the CBOT engine!

melex750:
In the first example, the return value only affects the next method in the chain,
the initial pointer doesn't change.

I'm ok to remove it though.

CyborgMaster:
As long as the first syntax wouldn't change the original pointer value, I think it's a great thing, and vote we keep it.  

In the second example I do think it's also important that something like this works:

```cpp
int ret = new AClass().performCalculations();
```

melex750:
@CyborgMaster , That works, trying to prevent it would complicate things a little bit.
The last return value is left on the stack, for whatever came before "new".

CyborgMaster:
@melex750, that's perfect.  That is what should happen.

krzys-h:
I'm still not sure whether I like the first syntax or not but I merged it anyway. I guess it's OK as long as it's behaviour is properly documented, because it looks a bit ambiguous to me.

PR #834, opened 2016-09-17
Implicit cast and null when passing arguments

melex750:
Plus error handling for "Ambiguous call to overloaded function" (see unit tests)

Fixed a bug that slows down finding in-class methods at runtime.
Fixes this bug:
```c++
public class TestClass
{
    public int Test() { // <-- note the public keyword
        message("TestClass");
        return 1;
    }
}

int Test() { message("Test"); return 2; } <-- should be called, but it's not.

extern void object::New()
{
    Test(); <------- tries to call the one in TestClass
                     but doesn't do anything.
    this.Test(); <--/   <----------------------- should be a compile error

    message(""+Test()); <-- failed assert
}
```
Also improved the way CBot handles access to
protected and private class member variables.

### DIFF
--- a/src/common/restext.cpp
+++ b/src/common/restext.cpp
@@ -717,6 +717,7 @@ void InitializeRestext()
     stringsCbot[CBot::CBotErrPrivate]       = TR("Private element");
     stringsCbot[CBot::CBotErrNoPublic]      = TR("Public required");
     stringsCbot[CBot::CBotErrNoExpression]  = TR("Expression expected after =");
+    stringsCbot[CBot::CBotErrAmbiguousCall] = TR("Ambiguous call to overloaded function");
 
     stringsCbot[CBot::CBotErrZeroDiv]       = TR("Dividing by zero");
     stringsCbot[CBot::CBotErrNotInit]       = TR("Variable not initialized");
--- a/test/unit/CBot/CBot_test.cpp
+++ b/test/unit/CBot/CBot_test.cpp
@@ -1904,3 +1904,152 @@ TEST_F(CBotUT, ClassMethodWithPublicKeyword)
         CBotErrUndefCall
     );
 }
+
+TEST_F(CBotUT, ClassTestProtectedMember)
+{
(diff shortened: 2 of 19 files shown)


src/common/restext.cpp
```diff
@@ -717,6 +717,7 @@ void InitializeRestext()
     stringsCbot[CBot::CBotErrPrivate]       = TR("Private element");
     stringsCbot[CBot::CBotErrNoPublic]      = TR("Public required");
     stringsCbot[CBot::CBotErrNoExpression]  = TR("Expression expected after =");
+    stringsCbot[CBot::CBotErrAmbiguousCall] = TR("Ambiguous call to overloaded function");
```
krzys-h: You should run ```make update-pot && make``` every time you edit restext.cpp in order to update translation files. I guess this should be documented somewhere in the file itself.

test/unit/CBot/CBot_test.cpp
```diff
@@ -1786,3 +1786,270 @@ TEST_F(CBotUT, ClassNewConstructorMethodChain)
...
+TEST_F(CBotUT, AmbiguousCallWithNumbers)
+{
+    ExecuteTest(
+        "void Test(int i, float f) {}\n"
+        "void Test(float f, int i) {}\n"
+        "\n"
+        "extern void AmbiguousCallNumbers() {\n"
+        "    Test(1, 2);\n"
+        "}\n",
+        CBotErrAmbiguousCall
```
melex750: In this case it would've used the first one it found, that being the one on top.

If both functions were public and in other programs, it's almost random which
function would be used, and that choice could change between save and reload,
which could easily crash the game.

melex750:
Thanks, the documentation looks good, sorry I took so long.
I was just about to push changes.  I found another bug in related code
and wrote some comments and filled in a bunch documentation.

krzys-h:
No problem, I just wanted to do something productive while I had a bit of free time

Issue #820, opened 2016-08-14
Editor freaks out with really long file


CyborgMaster:
When loading a really long file (>1000 lines), the editor freaks out.

![image](https://cloud.githubusercontent.com/assets/1882579/17652592/9e6df1e8-623d-11e6-9041-8278c2771032.png)

CyborgMaster:
I found the code that defines the max.  I propose we increase them by 10x.

``` cpp

//! maximum number of characters in CBOT edit
const int EDITSTUDIOMAX     = 200000;
//! maximum total number of lines
const int EDITLINEMAX       = 10000;
```

krzys-h:
I'm pretty sure this is a duplicate, but I don't have time to look for issue # right now. If you just increase the limit then at some point somebody will hit it again (and also, you are wasting RAM, very slightly but still). This whole class should be refactored to keep everything in variable-sized STL containers instead of fixed-size C arrays.
And while we're talking about it, I suggest we make this issue high priority because I saw people complaining about this limit for a long time (and I think it was even increased once already)

CyborgMaster:
I agree, STL containers would be much better.  Should we increase the hard limits temporarily while we fix it the right way?  I think 10,000 lines is a fine limit and can't imaging anything really getting past that in a reasonable manner.  That way it would be more usable until we got around to the correct fix.

MatiRg:
I already removed C like array in favor of STL.I need only make some adjustments and I cann make pull request.

B-CE:
ref : #300

Issue #300, opened 2014-04-09
Character limit in the code editor


CHmSID:
There is a character limit in the editor. The limit is about 10 full pages of characters.

PR #836, opened 2016-09-22
Remove character limit in CEdit

MatiRg:
Edit Control now can use infinite character limit and of course it can still be set to use certain amount of characters.
This issue:
https://github.com/colobot/colobot/issues/820
and I think also:
https://github.com/colobot/colobot/issues/300
In future You probally will migrate to CEGUI so all UI issues will be then resolved?


src/ui/controls/edit.h
```diff
@@ -232,51 +229,51 @@ class CEdit : public CControl
...
-    Math::Point     m_mouseFirstPos;
-    Math::Point     m_mouseLastPos;
-    float       m_column;
-
-    bool        m_bCapture;
-
-    bool        m_bUndoForce;
-    OperUndo    m_undoOper;
-    EditUndo    m_undo[EDITUNDOMAX];
+    std::unique_ptr<CScroll>                 m_scroll;           // vertical scrollbar on the right
```
krzys-h: Please avoid unnecessary formatting changes like these in pull requests, they make them harder to review. Besides, aligning to long names like `std::array<HyperHistory, EDITHISTORYMAX>` decreases readability a lot.

src/ui/controls/edit.h
```diff
@@ -132,8 +124,13 @@ class CEdit : public CControl
...
 
-    void        SetMaxChar(int max);
-    int         GetMaxChar();
+    void        SetMaxChar(int max);
+    //! Get max characters that can be hold before truncation or if is Infinite edit actual allocated characters in m_text
+    int         GetMaxChar();
+    //! Has no character limit for edit
+    bool        IsInfinite();
+    //! Set no character limit for edit
+    void        SetInfinite();
```
krzys-h: I don't think having a separate function for this is really necessary. Infinite should be the default, and if you ever need to revert SetMaxChar setting, you can always just do `SetMaxChar(std::numeric_limits<int>::max())`

src/ui/controls/edit.h
```diff
@@ -232,51 +229,51 @@ class CEdit : public CControl
...
-
-    bool        m_bCapture;
-
-    bool        m_bUndoForce;
-    OperUndo    m_undoOper;
-    EditUndo    m_undo[EDITUNDOMAX];
+    std::unique_ptr<CScroll>                 m_scroll;           // vertical scrollbar on the right
+
+    int                                      m_maxChar;          // max length of the buffer m_text or -1 to be infinite
+    std::vector<char>                        m_text;             // text (without zero terminator)
```
krzys-h: Wouldn't it be way easier to just make this (and m_format) an std::string?

src/object/implementation/program_storage_impl.cpp
```diff
@@ -131,7 +131,7 @@ Program* CProgramStorageObjectImpl::CloneProgram(Program* program)
 
     // TODO: Is there any reason CScript doesn't have a function to get the program code directly?
     auto edit = MakeUnique<Ui::CEdit>();
-    edit->SetMaxChar(Ui::EDITSTUDIOMAX);
+    edit->SetInfinite();
```
krzys-h: Not really necessary as infinite is now the default
(also applies to all other occurrences of this but I won't copy-paste this comment a billion times)

src/ui/controls/edit.cpp
```diff
@@ -47,19 +47,20 @@
...
-const float DELAY_SCROLL    = 0.1f;
+const float DELAY_SCROLL     = 0.1f;
 //! expansion for \b;
-const float BIG_FONT        = 1.6f;
-
-
-
+const float BIG_FONT         = 1.6f;
+//! start characters
+const int DEFAULT_CHARACTERS = 100;
```
krzys-h: This constant is not really needed, default is infinite anyway

src/ui/controls/edit.cpp
```diff
@@ -1256,14 +1260,25 @@ void CEdit::DrawColor(Math::Point pos, Math::Point dim, Gfx::Color color)
 // Give the text to edit.
 
 void CEdit::SetText(const char *text, bool bNew)
-{
+{
     int     i, j, font;
     bool    bBOL;
+
+    if ( text == nullptr )
```
krzys-h: For handling of wrong function arguments it's a lot better to use `assert()` than to fail silently.

src/ui/controls/edit.cpp
```diff
@@ -1256,14 +1260,25 @@ void CEdit::DrawColor(Math::Point pos, Math::Point dim, Gfx::Color color)
...
     bool    bBOL;
+
+    if ( text == nullptr )
+    {
+        return;
+    }
 
     if ( !bNew )  UndoMemorize(OPERUNDO_SPEC);
 
     m_len = strlen(text);
```
krzys-h: Try convert all C style string functions to std::string equivalents whenever possible (in this case, also change function argument to std::string)

src/ui/controls/edit.cpp
```diff
@@ -1439,414 +1455,420 @@ void CEdit::FreeImage()
 
 bool CEdit::ReadText(std::string filename, int addSize)
 {
-    int         len, i, j, n, font, iLines, iCount;
```
krzys-h: You did something weird to whitespace in this whole block (and a few other ones) but I can't quite figure out what. Indentation seems correct, line endings possibly?

src/ui/controls/edit.cpp
```diff
@@ -1939,10 +1963,23 @@ void CEdit::SetMaxChar(int max)
 }
 
 int CEdit::GetMaxChar()
-{
+{
+    if ( IsInfinite() )
+    {
+        return m_text.size();
```
krzys-h: Split this into a separate function

src/ui/controls/edit.cpp
```diff
@@ -2120,11 +2157,22 @@ void CEdit::SetMultiFont(bool bMulti)
     m_format.clear();
 
     if (bMulti)
-    {
-        m_format.reserve(m_maxChar+1);
-        for (int i = 0; i <= m_maxChar+1; i++)
-        {
-            m_format.push_back(m_fontType);
+    {
+        if ( !IsInfinite() )
```
krzys-h: It's not really good to depend on the value of IsInfinite() in code like this, using std::string like mentioned in other comments should help eliminate that.

src/ui/controls/edit.h
```diff
@@ -132,8 +124,13 @@ class CEdit : public CControl
     bool        ReadText(std::string filename, int addSize=0);
     bool        WriteText(std::string filename);
 
-    void        SetMaxChar(int max);
-    int         GetMaxChar();
+    void        SetMaxChar(int max);
+    //! Get max characters that can be hold before truncation or if is Infinite edit actual allocated characters in m_text
+    int         GetMaxChar();
+    //! Has no character limit for edit
+    bool        IsInfinite();
```
krzys-h: This should be removed along with SetInfinite(), external code doesn't really have any reason to care whether the the CEdit instance is infinite or not, if it does you can always use GetMaxChar() instead

src/ui/controls/edit.cpp
```diff
@@ -47,19 +47,20 @@
...
 //! expansion for \b;
-const float BIG_FONT        = 1.6f;
-
-
-
+const float BIG_FONT         = 1.6f;
+//! start characters
+const int DEFAULT_CHARACTERS = 100;
+//! infinite edit flag
+const int INFINITE_EDIT = -1;
```
krzys-h: You don't really need a separate flag, just use `std::numeric_limits<int>::max()`

MatiRg:
I will do changes when I will have time thanks.

MatiRg:
Nevermind  I will rewrite changes to CEdit from start

MatiRg:
std::numeric_limits<int>::max() gives bad_alloc exception so we will need lower number.

MatiRg:
I reworked code to use std::string as You told it was way easier it`s still not fully done.

src/ui/studio.cpp
```diff
@@ -1619,25 +1618,25 @@ bool CStudio::ReadProgram()
...
-    if ( p == nullptr || p != filename+strlen(filename)-4 )
-    {
-        strcat(filename, ".txt");
-    }
-    strcpy(dir, SearchDirectory(true).c_str());
-    strcat(dir, filename);
+    p = filename.find(".txt");
+    if ( p == std::string::npos )
+    {
+        filename += ".txr";
```
melex750: txt*

src/ui/studio.cpp
```diff
@@ -1657,32 +1656,32 @@ bool CStudio::WriteProgram()
...
-    }
-    strcpy(dir, SearchDirectory(true).c_str());
-    strcat(dir, filename);
+    filename = pe->GetText(100);
+    if ( filename.empty() )  return false;
+
+    p = filename.find(".txt");
+    if ( p == std::string::npos )
+    {
+        filename += ".txr";
```
melex750: txt*

MatiRg:
ok I`am on it

MatiRg:
I still need to deal with CRLF endings

src/ui/controls/edit.cpp
```diff
@@ -90,11 +89,10 @@ bool IsSep(int character)
 //! Object's constructor.
 CEdit::CEdit()
     : CControl(),
+      m_text( CHARACTERS_MAX + 1, '\0' ),
```
krzys-h: You are allocating 1MB of memory every time CEdit is created, this is effectively no better than an array with hardcoded size. Just leave it as an empty std::string object. Then you should be able to remove the CHARACTERS_MAX constant completly.

src/ui/controls/edit.cpp
```diff
@@ -1940,7 +1936,7 @@ void CEdit::SetMaxChar(int max)
 
 int CEdit::GetMaxChar()
 {
-    return m_maxChar;
+    return m_text.capacity() - 1;
```
krzys-h: m_text.capacity() can't really be used to determine the max amount of characters that CEdit is allowed to hold as it's value can be automatically adjusted by the STL implementation. This information should still be held as separate variable. See http://www.cplusplus.com/reference/string/string/capacity/

> Notice that this capacity does not suppose a limit on the length of the string. When this capacity is exhausted and more is needed, it is automatically expanded by the object (reallocating it storage space).
> 
> The capacity of a string can be altered any time the object is modified, even if this modification implies a reduction in size or if the capacity has not been exhausted (this is in contrast with the guarantees given to capacity in vector containers).

src/ui/studio.cpp
```diff
@@ -512,30 +512,30 @@ void CStudio::SearchToken(CEdit* edit)
             text = const_cast<char *>(GetObjectName(type));
             if ( text[0] != 0 )
             {
-                if ( strcmp(token, text) == 0 )
+                if ( token == text ) //strcmp(token, text) == 0 )
```
krzys-h: remove the commented out part

src/ui/studio.cpp
```diff
@@ -512,30 +512,30 @@ void CStudio::SearchToken(CEdit* edit)
...
-                    SetInfoText(std::string(token), true);
+                    SetInfoText(token, true);
                     return;
                 }
             }
             text = const_cast<char *>(GetObjectAlias(type));
             if ( text[0] != 0 )
             {
-                if ( strcmp(token, text) == 0 )
+                if ( token == text ) //strcmp(token, text) == 0 )
```
krzys-h: remove the commented out part

src/ui/controls/edit.cpp
```diff
@@ -1457,23 +1453,25 @@ bool CEdit::ReadText(std::string filename, int addSize)
...
 
-    m_maxChar = len+addSize+100;
     m_len = len;
     m_cursor1 = 0;
     m_cursor2 = 0;
 
     FreeImage();
 
-    m_text = std::vector<char>(m_maxChar+1, '\0');
+    m_text = std::string(len2+1, '\0');
```
krzys-h: You are wasting RAM again, see the comment in CEdit constructor

src/ui/controls/edit.cpp
```diff
@@ -2121,8 +2117,8 @@ void CEdit::SetMultiFont(bool bMulti)
 
     if (bMulti)
     {
-        m_format.reserve(m_maxChar+1);
-        for (int i = 0; i <= m_maxChar+1; i++)
+        m_format.reserve(GetMaxChar());
+        for (int i = 0; i <= static_cast<int>(GetMaxChar()); i++)
```
krzys-h: Since default character limit is to be assumed infinite (well, as close to infinite as possible in 32-bit numbers), you can't just start filling up infinite amount of memory. It should be dynamic like m_text, and their lengths should be kept in sync (which is probably the most complicated part of code required to fully remove the limit).

Maybe it would be a bit easier if you changed type of m_format to `std::basic_string<Gfx::FontMetaChar>`?

src/ui/controls/edit.cpp
```diff
@@ -89,7 +89,8 @@ bool IsSep(int character)
 //! Object's constructor.
 CEdit::CEdit()
     : CControl(),
-      m_text( CHARACTERS_MAX + 1, '\0' ),
+      m_maxChar(1000000),
```
krzys-h: Now that you fixed the memory allocation, you should be able to use `std::numeric_limits<int>::max()` here

MatiRg:
Alright now it will work.

krzys-h:
Thanks! Now I'll test it for a while to make sure there are no subtle bugs hidden somewhere as the editor is quite an important part of the game, if everything works correctly this is good to be merged.

MatiRg:
Thanks!

Issue #787, opened 2016-06-05
Allow ctrl + backspace in the editor.


VictorDenisov:
Current editors allow ctrl + arrow - move one word left or right, ctrl + del, ctrl + backspace, delete one word or backspace one word.

It would be convenient to have this feature in colobot.

RaptorParkowsky:
Definetely. :+1:

MatiRg:
I think I have something similar

MatiRg:
I think I can do it

PR #839, opened 2016-09-30
Allow ctrl + backspace in the editor.

MatiRg:
Deleting words in Edit by using Ctrl+Delete and Ctrl+Backspace.
Issue: https://github.com/colobot/colobot/issues/787

### DIFF
--- a/src/ui/controls/edit.cpp
+++ b/src/ui/controls/edit.cpp
@@ -2777,18 +2777,13 @@ void CEdit::DeleteWord(int dir)
 
     m_cursor1 = m_cursor2;
 
-    if( dir < 0)
+    if ( dir < 0)
     {
         if ( !m_cursor1 ) return;
-        if ( IsBreaker( m_text[m_cursor1] ) )
-        {
-            m_cursor2 = --m_cursor1;
-            GetLogger()->Info( "CASE 0\n" );
-        }
+        if ( IsBreaker( m_text[m_cursor1] ) ) m_cursor2 = --m_cursor1;
         else if ( !IsDelimiter( m_text[m_cursor1] ) && !IsDelimiter( m_text[m_cursor1-1] ) )
         {
             while ( m_cursor1 < m_len  && !IsDelimiter( m_text[m_cursor1] ) ) ++m_cursor1;
-            GetLogger()->Info( "CASE 1\n" );
         }
 
         if ( IsDelimiter( m_text[m_cursor2] ) )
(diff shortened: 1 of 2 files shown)


tomangelo2:
Few issues I've found.
1. If cursor "touches" second word (for example: lorem |ipsum, | is a cursor) and you press Ctrl+Backspace, second word (ipsum) is deleted instead of first one (lorem). Works correctly for Ctrl+Delete.
2. Ctrl+Backspacing multiple words leave all spaces that were between these words.
3. Both deleting and backspacing ignores punctuation marks (commas, dots, dashes, etc).

MatiRg:
1. Overlooked that.
2.So I will make all whitespaces erased
3.I will add special characters then

MatiRg:
Something like this?

tomangelo2:
These issues are fixed now, but I think that one space after deleting would be useful, now nearby words are merged into one. Also I noticed that if cursor is between word and comma (lorem|, ipsum) and press Ctrl+Delete, then lorem is deleted with comma, instead of only comma, which is default behavior in most editors.

MatiRg:
Ok, if You want this I will do it, tomorrow.

MatiRg:
I never used before deleting entire word in any editor, but I looked how in Notepad++ this is working nad I`am will do something like that.

MatiRg:
of course it wont work I just synced my local files with repository :)

MatiRg:
ok

HybridDog:
geany supports ctrl+backspace and lots of others, e.g. ctrl+b to jump to the opposite bracket
https://github.com/geany/geany

src/ui/controls/edit.cpp
```diff
@@ -84,7 +84,17 @@ bool IsSep(int character)
     return !IsWord(character);
 }
 
+bool IsBreaker(char c)
+{
+    return ( c == '.'  || c == '{' || c == '}' ||
+             c == ';' || c == ':' || c == '[' || c == ']' ||
+             c == '(' || c == ')' || c == '=' );
```
HybridDog: l think = shouldn't be a breaker, +,-,/ etc are also no breakers and using = without spaces is bad code style, isn't it?

MatiRg: Well I would add also +-*\%,? but I`am not sure how Colobot Team want to see it.

MatiRg:
Maybe we should expand delimiter list to: " ' , < > + - * / % ?

HybridDog:
MatiRg, no, by default a space is after , + - * / % ? < >
if the space is not present it can be considered bad code style

HybridDog:
krzys-h, l can't see the problem on your gif due to low image quality.

krzys-h:
```C++
extern void object::New()
{
	object item = radar(Titanium|);
}
```
| means where you should put the cursor. Then do ctrl+backspace and you'll be left with:
```C++
extern void object::New()
{
	object item = radar(|m);
}
```
It works correctly if you put the cursor anywhere else inside `Titanium`

Issue #316, opened 2014-08-08
Add command history for cheat console


krzys-h:
As mentioned in #315

PR #869, opened 2016-12-07
Command history for cheat console

Smok94:
Adds console command history that store only correctly spelled commands.
Browsable by up and down arrow keys.

Issue: #316

Nobody made i so far, so I rewited my old code using std::deque and std::string. Looks much clearner and works well.

### DIFF
--- a/src/level/robotmain.cpp
+++ b/src/level/robotmain.cpp
@@ -763,7 +763,7 @@ bool CRobotMain::ProcessEvent(Event &event)
                 m_interface->SetFocus(pe);
                 if (m_phase == PHASE_SIMUL) m_cmdEditPause = m_pause->ActivatePause(PAUSE_ENGINE);
                 m_cmdEdit = true;
-                m_commadHistoryIndex = -1;
+                m_commandHistoryIndex = -1; // no element selected in command history
             }
             return false;
         }
@@ -778,23 +778,25 @@ bool CRobotMain::ProcessEvent(Event &event)
         }
     }
 
+    // Browse forward command history with UP key
     if (event.type == EVENT_KEY_DOWN &&
         event.GetData<KeyEventData>()->key == KEY(UP) && m_cmdEdit)
     {
         Ui::CEdit* pe = static_cast<Ui::CEdit*>(m_interface->SearchControl(EVENT_CMD));
         if (pe == nullptr) return false;
         std::string cmd = GetNextFromCommandHistory();
-        if (cmd != "") pe->SetText(cmd);
+        if (!cmd.empty()) pe->SetText(cmd);
         return false;
     }
 
+    // Browse backward command history with DOWN key
     if (event.type == EVENT_KEY_DOWN &&
         event.GetData<KeyEventData>()->key == KEY(DOWN) && m_cmdEdit)
     {
         Ui::CEdit* pe = static_cast<Ui::CEdit*>(m_interface->SearchControl(EVENT_CMD));
         if (pe == nullptr) return false;
         std::string cmd = GetPreviousFromCommandHistory();
-        if (cmd != "") pe->SetText(cmd);
+        if (!cmd.empty()) pe->SetText(cmd);
         return false;
     }
 
@@ -814,7 +816,7 @@ bool CRobotMain::ProcessEvent(Event &event)
             m_cmdEditPause = nullptr;
         }
         ExecuteCmd(cmd);
-        if (m_isCommand) PushToCommandHistory(cmd);
+        PushToCommandHistory(cmd);
         m_cmdEdit = false;
         return false;
     }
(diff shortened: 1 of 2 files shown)


src/level/robotmain.cpp
```diff
@@ -1444,6 +1471,8 @@ void CRobotMain::ExecuteCmd(const std::string& cmd)
 
     if (m_phase == PHASE_SIMUL)
         m_displayText->DisplayError(ERR_CMD, Math::Vector(0.0f,0.0f,0.0f));
+
+    m_isCommand = false;
```
krzys-h: Pleeease don't introduce a global variable to `CRobotMain` for something that can be local to a function, it's already a big mess. There is a ton of better ways to handle this. For example you can just return it as return value from `ExecuteCmd`.

Actually, I personally don't think this is needed at all, keeping invalid commands in the history is a good way to fix accidental typos without having to retype the whole command.

src/level/robotmain.cpp
```diff
@@ -778,6 +779,26 @@ bool CRobotMain::ProcessEvent(Event &event)
     }
 
     if (event.type == EVENT_KEY_DOWN &&
+        event.GetData<KeyEventData>()->key == KEY(UP) && m_cmdEdit)
+    {
+        Ui::CEdit* pe = static_cast<Ui::CEdit*>(m_interface->SearchControl(EVENT_CMD));
+        if (pe == nullptr) return false;
+        std::string cmd = GetNextFromCommandHistory();
+        if (cmd != "") pe->SetText(cmd);
```
krzys-h: It would be nice to change this to `if (!cmd.empty())` for readability (and possibly performance but that's not that big of a deal here)

Smok94:
Now should be ok. Isn't there way to merge commits already pushed to github?

krzys-h:
By "merge" you probably mean squash:
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits
If you do that, you have to use a force-push to update the GitHub version. But there is no real need to do that here, I can do it while merging the changes and in pull requests it's usually better to keep a log of all changes that happened during the discussion.

src/level/robotmain.cpp
```diff
@@ -5866,3 +5890,28 @@ bool CRobotMain::GetDebugCrashSpheres()
...
+
+    m_commandHistory.push_front(str);
+
+    if (m_commandHistory.size() > 50) // to avoid infinite growth
+        m_commandHistory.pop_back();
+}
+
+std::string CRobotMain::GetNextFromCommandHistory()
+{
+    if (m_commandHistory.empty() || m_commandHistory.size() <= m_commandHistoryIndex + 1) // no next element
```
krzys-h:
```
/home/krzys_h/Pulpit/colobot/src/level/robotmain.cpp: In member function ‘std::__cxx11::string CRobotMain::GetNextFromCommandHistory()’:
/home/krzys_h/Pulpit/colobot/src/level/robotmain.cpp:5907:61: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     if (m_commandHistory.empty() || m_commandHistory.size() <= m_commandHistoryIndex + 1) // no next element
                                     ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Smok94:
Visual Studio don't shows such warning. I added cast to int. Maybe it will fix this warning for you.

krzys-h:
You can always check Jenkins output for the pull request. Right now introducing warnings or breaking code style doesn't automatically mark the PR as failing, only if the code doesn't compile at all, but we'll probably implement that in the future.
https://compiled.colobot.info/job/colobot/job/colobot/job/PR-869/lastBuild/console

src/level/robotmain.cpp
```diff
@@ -5866,3 +5890,28 @@ bool CRobotMain::GetDebugCrashSpheres()
...
+
+    m_commandHistory.push_front(str);
+
+    if (m_commandHistory.size() > 50) // to avoid infinite growth
+        m_commandHistory.pop_back();
+}
+
+std::string CRobotMain::GetNextFromCommandHistory()
+{
+    if (m_commandHistory.empty() || (int)m_commandHistory.size() <= m_commandHistoryIndex + 1) // no next element
```
krzys-h: And if you look in the Jenkins logs, you'll see this:
```
/var/lib/jenkins/workspace/colobot/colobot/PR-869/src/level/robotmain.cpp: In member function 'std::string CRobotMain::GetNextFromCommandHistory()':
/var/lib/jenkins/workspace/colobot/colobot/PR-869/src/level/robotmain.cpp:5906:64: warning: use of old-style cast [-Wold-style-cast]
     if (m_commandHistory.empty() || (int)m_commandHistory.size() <= m_commandHistoryIndex + 1) // no next element
                                                                ^
```
So you just introduced a warning while fixing another warning ;)

PR #873, opened 2016-12-22
Fix #797

### LINKED ISSUE #797: Robot programs list break after loading saved game.
When we load a savegame (both created when game crashed and normal saves), programs marked as read-only disappear. Also actually running programs cannot be aborted with Stop button.
screenshot_3

piotrwalkusz1:
I think it's little troublesome to contain string "ai/script_name.txt" in one variable and string "script_name.txt" in another.

### DIFF
--- a/po/ru.po
+++ b/po/ru.po
@@ -85,7 +85,7 @@ msgid "Alternative camera mode\\Move sideways instead of rotating (in free camer
 msgstr "Альтернативный режим камеры\\Движение в стороны вместо поворачивания (в режиме свободной камеры)"
 
 msgid "Ambiguous call to overloaded function"
-msgstr ""
+msgstr "Странный вызов перегруженной функции"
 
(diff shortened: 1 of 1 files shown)


krzys-h:
I agree that referring to the same file as *"script_name.txt"* or *"ai/script_name.txt"* depending on a context is a bit misleading, but it was done that way to maintain compatibility with old level file format, which always assumed that the file is to be loaded from "ai" directory (unless it started with *%user%*) while allowing userlevels, which were pretty much completely reimplemented, to use more complex directory scheme with *%lvl%* and *%chap%*. As you can see even I am getting lost in it myself.

As for the solution itself, that works but... perhaps it would actually be better to unify this convention a bit and allow level files to have absolute paths (well, actually relative to root of data directory) when prefixed with / or something. This would help prevent such problems in the future.

Issue #697, opened 2015-12-25
Non-public classes in CBot


krzys-h:
For some use cases it might be nice to have non-public classes in CBot. That would prevent naming conflicts between programs (or multiple instances of the same program).

immibis:
I would like to see public classes not exist actually, it makes no logical sense for multiple robots to share their RAM.

However that would break stuff from the original game.

cbot-coder:
Public classes have a use: to share info between bots

immibis:
@cbot-coder I don't care. It makes no logical sense for different robots to share their RAM.

melex750:
> It makes no logical sense for different robots to share their RAM.

By that logic, public functions also don't make sense.
It's a moot point since the game doesn't simulate RAM or a CPU.

In fact, there are distributed computing systems that operate as though physically
separate computers do share RAM with support from the OS and network.

I think the game wasn't even meant to be a literal simulation of robots.
It's a metaphorical representation of a programming environment, intended
for learning/teaching programming, where bugs(giant insects) are the enemy.
I actually read one of the original dev blogs that basically said as much.

immibis:
> > It makes no logical sense for different robots to share their RAM.
> 
> By that logic, public functions also don't make sense.

correct

> It's a moot point since the game doesn't simulate RAM or a CPU.

The game simulates separate robots, and you can easily see by looking at the game graphics, that there are no wires going between the robots.

You could say there is wireless communication, but there isn't, because we know what wireless communication looks like in this game (look at the ExchangePost send/receive animation) and there isn't that either.

cbot-coder:
Feel free to design your solutions so as not to use sharing of data between
bots. I typically don't, as I haven't found it to be terribly reliable or
helpful in the missions I've programmed so far at least. Instead I
typically give each bot a role to do, and certain parameters (such as,
wheeled bot only goes for resources within 50 meters of space ship, winged
grabber does things further away). I also tried having bots that need their
cells replaced by a grabber, tell the grabber so via public data
structures. But again, I found it not to work that well (for various
reasons) and decided rather to have the grabber periodically look for bots
that are standing still and have a near-empty cell. That seemed to be a bit
more reliable.

As for being semantically correct, I think it's really picking at minor
details. If someone wants to, they could find all kinds of unrealistic
things about the game and the programming. For example, the fact that bots
can use radar() to precisely pinpoint everything in the game (unless it's
magically hidden for the purpose of the level). In a more realistic world,
[...]

Issue #526, opened 2015-07-22
Make public functions in CBot not shared between teams


krzys-h:

CHmSID:
Research should also be separate, but I don't know if it needs its own issue?

krzys-h:
Research is already separate

hexagonrecursion:
AI also uses the same public function namespace: if you load the below program into one of your bots all AlienAnt's spawned by the AlienQueen will just stand in place and do nothing.

```
public void object::Attack() {	
}
```

melex750:
Already done in my WIP along with folding constants, some dead code elimination, infinite recursion via tail call optimization, and several other optimizations and fixes...

hexagonrecursion:
What WIP? Where?

hexagonrecursion:
> I've just finished proof reading after testing. It will be in a pull request soon.

soon?

hexagonrecursion:
> I've just finished proof reading after testing. It will be in a pull request soon.

@melex750 please push the code to a publicly visible branch in https://github.com/melex750/colobot. Let me help you

melex750:
> please push the code to a publicly visible branch...Let me help you

Thanks, I will probably do that.  After finding several conflicts with ```dev``` branch, I began to pair it down into smaller more reviewable units-of-work, the first of which was #1762 .  Another thing, it was a little messy with too many little fix-up commits and I tend to prefer more complete commits for the self-documenting nature of them.

hexagonrecursion:
> I began to pair it down into smaller more reviewable units-of-work

:+1: :100: :guitar:

PR #889, opened 2017-01-13
Fixed #703 and added private classes

### LINKED ISSUE #703: Defining two classes with the same name, crashes the game.
Should be a compile error, but it compiles once, and crashes the next time you click compile/OK.

Changing `false` to `true` on [line 497 CBotClass.cpp](https://github.com/colobot/colobot/blob/dev/src/CBot/CBotClass.cpp#L497)  fixes this issue.

The only place this value is tested, is just above that on [line 474](https://github.com/colobot/colobot/blob/dev/src/CBot/CBotClass.cpp#L474).

piotrwalkusz1:
COMMIT 1 :

Fix  #703
Before we checked if class exists in other programs
```
CBotClass* pOld = CBotClass::Find(name);
if ( pOld != nullptr && pOld->m_IsDef )
```
Now we also check if class was defined in precompilation phase (that was changed a little bit in second commit)
```
if ( (pOld != nullptr && pOld->m_IsDef) || /* public class exists in different program */
      program->ClassExists(name))           /* class exists in this program */
```

Additionally before CBot code
```
public class Something
{
```
threw exception. Now it works.

-------------------------------------------

COMMIT 2 :

Issue #697 

Changes:
-added private classes
-if CBotProgram is destroying, its classes are destroying also (before CBotClass destructor was never executed)
-~~new static variable CBotProgram* CBotProgram::m_currentProgram (it says which program is being compiled or run now)~~ In COMMIT 3. I deleted this.
-change working CBotClass::Find function, now it search class in current program and in list of public classes (before only in list of public classes) 
-change (in CBotClass::Compile1)
```
if ( (pOld != nullptr && pOld->m_IsDef) ||
      program->ClassExists(name))    
```
to
```
if ( (isPrivate && CBotProgram::GetCurrentProgram()->ClassExists(name)) ||
         (!isPrivate && CBotClass::Exists(name)) )
```
-set CBotErrNotClass as compile error (before runtime error)
-fixed one test (PublicClasses) and added few new tests

------------------------------------------------------------------

COMMIT 3 :

Look at `PublicFunctionUsesPrivateClass` and `PublicFunctionUsesPrivateClassMethod` to see what was fixed. I changed type of variable CBotFunction::m_MasterClass from string to CBotClass* as well (before every class had unique name, now with private classes it is impossible).

Also, I deleted CBotProgram* CBotProgram::m_currentProgram variable and now I take CBotProgram variable from CBotCStack::GetProgram().

-------------------------------------------------------

TODO :
-fix saving and restoring static variables from private classes

### DIFF
--- a/src/CBot/CBotClass.cpp
+++ b/src/CBot/CBotClass.cpp
@@ -442,7 +442,7 @@ bool CBotClass::CheckCall(CBotProgram* program, CBotDefParam* pParam, CBotToken*
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-CBotClass* CBotClass::Compile1(CBotToken* &p, CBotCStack* pStack)
+CBotClass* CBotClass::Compile1(CBotToken* &p, CBotCStack* pStack, CBotProgram* program)
 {
     if ( !IsOfType(p, ID_PUBLIC) )
     {
@@ -455,7 +455,8 @@ CBotClass* CBotClass::Compile1(CBotToken* &p, CBotCStack* pStack)
     std::string name = p->GetString();
 
     CBotClass* pOld = CBotClass::Find(name);
-    if ( pOld != nullptr && pOld->m_IsDef )
+    if ( (pOld != nullptr && pOld->m_IsDef) || /* public class exists in different program */
+         program->ClassExists(name))           /* class exists in this program */
     {
         pStack->SetError( CBotErrRedefClass, p );
         return nullptr;
@@ -491,6 +492,8 @@ CBotClass* CBotClass::Compile1(CBotToken* &p, CBotCStack* pStack)
         int level = 1;
         do                                          // skip over the definition
         {
+            if (p == nullptr) break; // end of code
+
             int type = p->GetType();
             p = p->GetNext();
             if (type == ID_OPBLK) level++;
--- a/src/CBot/CBotClass.cpp
+++ b/src/CBot/CBotClass.cpp
@@ -257,24 +257,19 @@ bool CBotClass::IsIntrinsic()
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-CBotClass* CBotClass::Find(CBotToken* &pToken, bool onlyPublic)
+CBotClass* CBotClass::Find(CBotToken* &pToken, CBotProgram* program)
 {
-    return Find(pToken->GetString(), onlyPublic);
+    return Find(pToken->GetString(), program);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-CBotClass* CBotClass::Find(const std::string& name, bool onlyPublic)
+CBotClass* CBotClass::Find(const std::string& name, CBotProgram* program)
 {
-    if (!onlyPublic)
+    if (program != nullptr)
     {
-        CBotProgram* program = CBotProgram::GetCurrentProgram();
+        CBotClass* classe = program->FindClass(name);
 
-        if (program != nullptr)
-        {
-            CBotClass* classe = program->FindClass(name);
-
-            if (classe != nullptr) return classe;
-        }
+        if (classe != nullptr) return classe;
     }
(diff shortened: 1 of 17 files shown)


src/CBot/CBotEnums.h
```diff
@@ -245,16 +248,15 @@ enum CBotError : int
     CBotErrNoRetVal      = 6003, //!< function did not return results
     CBotErrNoRun         = 6004, //!< Run() without active function
     CBotErrUndefFunc     = 6005, //!< calling a function that no longer exists
-    CBotErrNotClass      = 6006, //!< this class does not exist
```
krzys-h: I would prefer avoiding changing the associated numbers for these errors if possible. Usually it's just because it generates a lot of mess in the commit diff, but in this case it's also because of an undocumented feature which allows you to catch these errors using undocumented try...catch construct by their IDs. I guess we should finally decide what to do with try..catch in cbot at some point...

src/common/restext.cpp
```diff
@@ -719,6 +719,8 @@ void InitializeRestext()
     stringsCbot[CBot::CBotErrNoPublic]      = TR("Public required");
     stringsCbot[CBot::CBotErrNoExpression]  = TR("Expression expected after =");
     stringsCbot[CBot::CBotErrAmbiguousCall] = TR("Ambiguous call to overloaded function");
+    stringsCbot[CBot::CBotErrNoPrivilege]   = TR("Access privilege required");
```
krzys-h: When editing `restext.cpp`, you should run `make update-pot` to update the translation files. At some point I said I would properly document this as a comment somewhere in the file but I obviously forgot...

krzys-h:
> -~~new static variable CBotProgram* CBotProgram::m_currentProgram (it says which program is being compiled or run now)~~ In COMMIT 3. I deleted this.

It's very good that you removed it, we should actually work towards removing the static variables where possible. In the future it would be awesome to 1) make the lists of public things not static so that they can be separate for teams (see #526) and 2) possibly in the future make CBot programs run in a separate thread(s) to not block rendering when doing lots of computations and synchronize with main game thread only when needed.

krzys-h:
also poking @melex750 and @CyborgMaster for a review if they want to ;)

src/CBot/CBotClass.cpp
```diff
@@ -444,18 +459,29 @@ bool CBotClass::CheckCall(CBotProgram* program, CBotDefParam* pParam, CBotToken*
...
+    IsOfType(p, ID_CLASS); // skip class keyword
 
-    std::string name = p->GetString();
+    std::string name = p->GetString(); // name of class
 
-    CBotClass* pOld = CBotClass::Find(name);
-    if ( pOld != nullptr && pOld->m_IsDef )
+    if ( (isPrivate && program->ClassExists(name)) ||            // check only in current program
+         (!isPrivate && CBotClass::Exists(name, program)) )      // check in current program and list of public classes
     {
```
melex750: This looks like a public class and a local class may be allowed to have the same name,
that probably shouldn't be allowed.

I like your idea to check current program for duplicates. --> "703"

piotrwalkusz1: I thought it would be useful to be able to overwrite public class with the same name form other program by private class, but in fact it can be deceptive.

piotrwalkusz1: Though...
If we have local class in one program
```
class Example
{
}
```
and next we will add second program with public class
```
public class Example
{
}
```
first program will stop be correct. Maybe overwriting classes should be possible?

src/CBot/CBotEnums.h
piotrwalkusz1: CBotErrNotClass is compile error (no runtime) and has to have value less than 6000, otherwise tests won't work correct. However, I will preserve other error values (between 6005 and 6007 will be gap).

PR #890, opened 2017-01-15
Fixed #703

### LINKED ISSUE #703: Defining two classes with the same name, crashes the game.

piotrwalkusz1:
Ok, now I will be sending commits one by one.

This fixes #703 issue.

### DIFF
--- a/src/CBot/CBotProgram.h
+++ b/src/CBot/CBotProgram.h
@@ -334,6 +334,12 @@ public:
     const std::list<CBotFunction*>& GetFunctions();
 
     /**
+     * \brief Check if class with that name exists (public or local)
+     * \return True if exists, otherwise false
+     */
+    bool ClassExists(std::string name);
+
+    /**
      * \brief Returns static list of all registered external calls
      */
     static CBotExternalCallList* GetExternalCalls();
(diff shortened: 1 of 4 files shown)


src/CBot/CBotProgram.h
```diff
@@ -334,6 +334,12 @@ class CBotProgram
     const std::list<CBotFunction*>& GetFunctions();
 
     /**
+     * \brief Check if class with that name exists (public or local)
```
krzys-h: This explanation is ambiguous, it sounds like this function will check both private classes in this program and the global list of public classes, which is not the case.

krzys-h:
@melex750 any thoughts?

melex750:
Looks good to me :)

PR #891, opened 2017-01-16
Fix syntax errors and memory leaks

melex750:
Increment/decrement syntax.
```c++
extern void object::New()
{
    ++;
    ++5;
    ++WheeledGrabber;

    Test t();
    t.getInt()++;  <---- segfault on run
    ++t.getInt();  <----/
}
public class Test {
    int getInt() { return 1; }
}
```
This compiles :(
```c++
public class AClass
{
    // c-tors and d-tors should allow only 'void' return type.
    int AClass() {return 1;}

    int ~AClass(int i) // destructors don't have params
    {return 1;}

    void ~AClass(){} // should be only one destructor
    void ~AClass(){} // should be only one destructor

    int ~i = 1;      // bad character in var name
    int ~j[] = {};   // bad character in var name

    int AClass = 2; // same name as this class

    void v[];  // array with type 'void'

    int i = 1;
    int i = 2; // variable already declared
}
```
A memory leak with CBotStack::RestoreState
See also changes to RestoreState in 35d60aa

Reproduce:
Load a saved game with a running CBOT program.
Stop a CBOT program to leak 96kb or (sizeof(CBotStack) * 1000) bytes.

Will leak that amount for each running program in the scene,
if you abort the level or load another saved game.

The ~27 running programs in TrueLevels-Amber-1 cause a 2.5MB leak
after loading a saved game, then abort or load another saved game.

Fixed many other memory leaks in CBot engine.
```bash
$ valgrind -v --leak-check=full --show-leak-kinds=all ./colobot_ut --gtest_filter=CBotUT*
```
**Before:**
```
==13756== LEAK SUMMARY:
==13756==    definitely lost: 53,480 bytes in 362 blocks
==13756==    indirectly lost: 192,456 bytes in 1,799 blocks
==13756==      possibly lost: 0 bytes in 0 blocks
==13756==    still reachable: 72,760 bytes in 2 blocks
==13756==         suppressed: 0 bytes in 0 blocks
==13756== 
==13756== ERROR SUMMARY: 111 errors from 111 contexts (suppressed: 0 from 0)
```
**After:**
```
==14435== LEAK SUMMARY:
==14435==    definitely lost: 0 bytes in 0 blocks
==14435==    indirectly lost: 0 bytes in 0 blocks
==14435==      possibly lost: 0 bytes in 0 blocks
==14435==    still reachable: 72,760 bytes in 2 blocks
==14435==         suppressed: 0 bytes in 0 blocks
==14435== 
==14435== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```


krzys-h:
Great job! :+1:

In the future it would be awesome to introduce automatic memory management with C++11 style pointers, just like we did in the main game, that would mostly prevent any memory leaks in the future. That is going to be a loooot of work though.

PR #894, opened 2017-01-20
Move CRobotMain::CreateScene() to separate class

piotrwalkusz1:
`Split level loading code from CRobotMain`

I know that I made `CRobotMain` friend of `CLevelLoader`, but there were used too many private variables in `CRobotMain::CreateScene()`. I think, until we refactor the rest of `CRobotMain`, it's the best solution.


src/level/level_loader.cpp
```diff
@@ -0,0 +1,1201 @@
...
+        { "ProhibitedToken",    MakeCommand(&CLevelLoader::HandleCommand_ProhibitedToken) },
+        { "EnableBuild",        MakeCommand(&CLevelLoader::HandleCommand_EnableBuild) },
+        { "EnableResearch",     MakeCommand(&CLevelLoader::HandleCommand_EnableResearch) },
+        { "DoneResearch",       MakeCommand(&CLevelLoader::HandleCommand_DoneResearch, LevelLoadMode::Normal) },
+        { "NewScript",          MakeCommand(&CLevelLoader::HandleCommand_NewScript) }
+};
+
+CLevelLoader::CLevelLoader(bool solutionEnabled)
+{
+    m_main = CRobotMain::GetInstancePointer();
```
krzys-h: New code should really use dependency injection instead of the GetInstancePointer() functions whenever possible

src/level/level_loader.cpp
```diff
@@ -0,0 +1,1201 @@
...
+
+    levelLoader->BuildScene(fixScene, mode);
+}
+
+void CLevelLoader::ResetScene(LevelLoadMode mode)
+{
+    m_main->m_base = nullptr;
+
+    if (mode != LevelLoadMode::Reset)
+    {
```
krzys-h: I think the level loader is not really a good place for this function. It's related to resetting the engine state, not actually loading anything.

src/level/level_loader.cpp
```diff
@@ -0,0 +1,1201 @@
...
+
+    // NOTE: Reset timer always, even when only resetting object positions
+    m_main->m_missionTimerEnabled = false;
+    m_main->m_missionTimerStarted = false;
+    m_main->m_missionTimer = 0.0f;
+}
+
+void CLevelLoader::BuildScene(bool fixScene, LevelLoadMode mode)
+{
+    m_main->m_fixScene = fixScene;
```
krzys-h: Setting fixScene variable should not be part of loading either, this should be set in CRobotMain itself before calling the loader.

src/level/level_loader.cpp
piotrwalkusz1: We use specific dependency injection framework or simply I should pass CRobotMain instance as parameter?

src/level/level_loader.cpp
```diff
@@ -0,0 +1,1201 @@
...
+            default:
+                break;
+            }
+
+            m_main->m_short->SetMode(false);  // vehicles?
+        }
+
+        m_main->m_map->ShowMap(m_main->m_mapShow);
+        m_main->m_map->UpdateMap();
+        // TODO: m_main->m_engine->TimeInit(); ??
```
krzys-h: This is more resetting of engine state after loading, not loading itself.

src/level/level_loader.cpp
```diff
@@ -0,0 +1,1201 @@
...
+        m_main->m_sceneReadPath = "";
+        throw;
+    }
+    m_main->m_sceneReadPath = "";
+
+    if (m_main->m_app->GetSceneTestMode())
+        m_main->m_eventQueue->AddEvent(Event(EVENT_QUIT));
+
+    m_main->m_ui->ShowLoadingScreen(false);
+    if (m_main->m_missionType == MISSION_CODE_BATTLE)
```
krzys-h: If you are going to clean up level loading, do it right. Creating the UI really shouldn't be invoked from CreateScene, CRobotMain::ChangePhase would probably be a better place.

src/level/level_loader.h
```diff
@@ -0,0 +1,158 @@
...
+ * \brief Tool to loading and building scene from file
+ */
+class CLevelLoader
+{
+public:
+    /**
+     * \brief Constructor, should not be used from outside CLevelLoader class,
+     * use CreateScene() instead.
+     */
+    CLevelLoader(bool solutionEnabled);
```
krzys-h: You can probably make the constructor private if you don't want it to be invoked manually

src/level/level_loader.cpp
```diff
@@ -0,0 +1,1201 @@
...
+    m_backgroundPath = "";
+    m_backgroundUp = Gfx::Color(0.0f, 0.0f, 0.0f, 0.0f);
+    m_backgroundDown = Gfx::Color(0.0f, 0.0f, 0.0f, 0.0f);
+    m_backgroundCloudUp = Gfx::Color(0.0f, 0.0f, 0.0f, 0.0f);
+    m_backgroundCloudDown = Gfx::Color(0.0f, 0.0f, 0.0f, 0.0f);
+    m_backgroundFull = false;
+    m_selectedObject = nullptr;
+}
+
+CLevelLoader::Command::Command(CommandFunction function, int executionConditions)
```
krzys-h: This constructor could probably be inlined in the header file

src/level/level_loader.cpp
krzys-h: Just pass it as a parameter

krzys-h:
> there were used too many private variables in `CRobotMain::CreateScene()`

Which is exacly why nobody tried to move it yet - there is no good way to do it because this function has just too many dependencies

src/level/level_loader.h
piotrwalkusz1: It was my first idea, but std::make_shared() need public constructor.

Issue #642, opened 2015-09-24
Custom functions not accepting "nan".


Mrocza:
The folowing program will display "0" instead of "nan":

```
extern void object::Bug(){
message(echo(nan));
}
int object::echo(int i){
return i;
}
```

tomaszkax86:
NaN is technically only defined for floating-point formats. There's no formal way to specify integer NaN, so I don't know how could this work at all. CBot probably handles nan constant specially, just like categories, so doing any operations will likely remove it's meaning from assigned variable.

Mrocza:
For float variables the program I noted returns "0.00" instead of "nan"

krzys-h:
It probably works just the same as the categories, doing anything with them just converts them to a regular int. This will be likely fixed during CBot refactoring.

Mrocza:
One can check the category number by simply doing:

    message(WheeledGrabber+0);

Attempting to apply that trick to "nan" results in an error.

Issue #30, opened 2012-07-05
CBOT - not returning a value in a function that expects one crashes the game


CyborgMaster:
A function that sometimes returns a value and sometimes doesn't will crash the game.  It should display a compile or runtime error instead.

string test(int i) {
  if (i == 0) {
    return "works"
  } else {
    //no return, crashes colobot.exe
}

krzys-h:
This was fixed in #897, closing

PR #897, opened 2017-01-24
Fix game crashing related to syntax errors

melex750:
**Syntax related failed asserts/segfaults include:**
* Missing closing bracket in last function.
* Compiling single word "public".
* Compiling only "extern void New("
* Compiling only "public class"
* etc.

**Fixed issue #30**

Also implemented default values for parameters

Fixed issue #642

Fixed 'point' constructor not executing when called with new.

### DIFF
--- a/po/colobot.pot
+++ b/po/colobot.pot
@@ -1745,6 +1745,9 @@ msgstr ""
 msgid "Non-void function needs \"return;\""
 msgstr ""
 
+msgid "This parameter needs a default value"
+msgstr ""
+
 msgid "Dividing by zero"
 msgstr ""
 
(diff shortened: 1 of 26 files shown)


krzys-h:
> Also implemented default values for parameters

This is something I wanted for SO LONG. I think we should start calling this CBot++ :stuck_out_tongue:

melex750:
https://github.com/colobot/colobot/blob/master/src/CBot/CBotInstr/CBotFunction.cpp#L501-L502
Here, CBOT favors upgrading to float over downgrading to int.
```c++
void Test(float f, float ff) {}
void Test(int   i, int   ii) {}

void TestNumbers() {
    Test(1, 2.2f);  <-- CBOT chooses float, C++ says it's ambiguous,
}
```
Maybe these number conversions should have equal precedence
in this case, and throw an ambiguous call error like C++,
so you never have to guess about which function will be called.

krzys-h:
@MrSimbax This needs documentation in SatCom, will make a new issue for that

PR #898, opened 2017-01-26
Move part of CRobotMain::CreateScene() to separate class

piotrwalkusz1:
`Split level loading code from CRobotMain`

My second attempt. Now `CLevelLoader` doesn't build scene itself, but only reads data from parser and returns data in convenient form to `CRobotMain`. I also fixed bug "first frame is very slow". Function `CEngine::UpdateGroundSpotTextures()` (called in `CEngine::Render()`) while first execution loads some textures. I made this function public and invoked before end of scene loading. Few issues can be unclear so I will try explain they :

- I added new class `CLevelReadOnlyParser` and function `CLevelParser::LoadFromText()` because they can help in writing unit tests, data doesn't have to be read from file.
- I changed `RenderIfNeeded()` to `Render()` in  `CScreenLoading::SetProgress()` because loading screen doesn't display properly, e.g. :
```
m_ui->GetLoadingScreen()->SetProgress(0.25f, "Calculating 2 + 2"); // Display "Calculating 2 + 2"
int a = 2 + 2; // Fast calculation
m_ui->GetLoadingScreen()->SetProgress(0.5f, "Calculating 143^1000"); // Skip this because "render isn't needed", user still sees "Calculating 2 + 2"
int b = pow(143, 1000); // Long calculation
// Result? User thinks that 2 + 2 takes that much time.
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -224,6 +224,9 @@ set(BASE_SOURCES
     level/build_type.h
     level/level_category.cpp
     level/level_category.h
+    level/level_load_mode.h
+    level/level_loader.cpp
+    level/level_loader.h
     level/mainmovie.cpp
     level/mainmovie.h
     level/parser/parser.cpp
--- a/src/graphics/engine/engine.cpp
+++ b/src/graphics/engine/engine.cpp
@@ -1564,6 +1564,21 @@ int CEngine::CreateGroundSpot()
     return index;
 }
 
+int CEngine::CreateGroundSpot(Math::Vector position, float radius, Gfx::Color color, float min, float max, float smooth)
+{
+    int rank = CreateGroundSpot();
+    if (rank != -1)
+    {
+        SetObjectGroundSpotPos(rank, position);
+        SetObjectGroundSpotRadius(rank, radius);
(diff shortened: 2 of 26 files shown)


krzys-h:
Closing this for now - it's outdated, will need to be redone anyway after merging buzzingcars and blupimania which add a bunch of extra commands (which, knowing us, MAY take a long time) and I'm not sure if I really like the extra effort it takes to add new commands with this implementation.

---

> I also fixed bug "first frame is very slow". Function `CEngine::UpdateGroundSpotTextures()` (called in `CEngine::Render()`) while first execution loads some textures.

I will probably merge this bugfix separately

> I added new class CLevelReadOnlyParser and function CLevelParser::LoadFromText() because they can help in writing unit tests, data doesn't have to be read from file.

That's actually a really good idea, the level parser doesn't have any tests at all because I was a noob and didn't really know how that works when I rewrote it a few years ago :P

> I changed `RenderIfNeeded()` to `Render()` in `CScreenLoading::SetProgress()` because loading screen doesn't display properly

RenderIfNeeded() exists for a reason - if you just call Render(), it will wait for a vblank each time, slowing down the loading significantly. This was a major issue when implementing the loading screen.

PR #902, opened 2017-01-31
Regression testing

piotrwalkusz1:
`7. Look into possibility of testing through recorded event traces`

Description is in `test/recorder/main.cpp` file, but I put this here too :

 **A tool for recording, replaying and testing**
 
To record gameplay as list of event :
`./colobot_recorder record NAME_OF_RECORD`
This command starts game and begins recording all events. After exit game, file `./records/NAME_OF_RECORD.json` will be created.
 
If you want to run level immediately after starting game, add optional arguments :
`./colobot_recorder record NAME_OF_RECORD CATEGORY CHAPTER LEVEL`
For example :
`./colobot_recorder record NAME_OF_RECORD 2 1 3`
will run third mission on Earth.
 
To replay recorded game :
`./colobot_recorder replay NAME_OF_RECORD`
 
To run tests :
`./colobot_recorder test`
 
To add test :
  1) Record proper gameplay and give it suitable name
  2) Write new test in the end of this file. Test should have form :
```
TEST_F(CRecorderTest, NAME_OF_TEST)
{
    auto assertion = [](CApplicationMock* app)
    {
        // Write test of code that will be executed while every frame
        // NOTE : ASSERT_* from google test framework won't stop test, it doesn't work in nested functions
    };
 
    StartTest("NAME_OF_RECORD", assertion);
}
```
  3) If test works properly, copy record file from build directory to source directory (`test/recorder/records`). When other team's member builds project, records will copied from source directory to build directory.

**Comments**
 - I used third-party library (https://github.com/nlohmann/json) on MIT licence.
 - I am not master of cmake. If something is wrong, please let me know.
 - This is only prototype. If concept will be accepted, I will improve this.
 - This can be used in other purpose than tests, e.g. after discovery bug somebody can record part of gameplay, send `record_name.json` file and thanks to it developers can easy reproduce issue.


MrSimbax:
I'm closing this due to inactivity.

hexagonrecursion (changes requested):
I think we should close this pull request
1. This approach is too fragile for regression testing.
    1. The recording contains mouse x and y coordinates. The replay will be incorrect if the user interface changes. I doubt we want to lock ourselves into never being able to improve the interface.
    2. How about recording the trace without using a mouse? This prevents us from improving the keyboard-based user interface
    3. The recording also breaks if we modify the map of the level that was recorded
    4. There are probably many other things we may want to improve that too invalidate recordings
2. The pull request claims this may be useful for bug reports, but unfortunately the implementation is too deeply coupled to the unit testing infrastructure. We don't ship the tests, we ship the game. Most players will get a version of the game without the recording feature. Even if we solve this problem, I suspect differences in how the game was compiled, installed and configured may prevent a recording made on one machine from being replayed on another - a big waste of time for both the person reporting the bug and the person trying to fix it
    1. Edit: Several game mechanics use random numbers. CBot has rand(). It appears that the trace does not capture the random seed
    2. Edit: We use c++ `rand()`. Even if we make sure that the seed is the same, the numbers it returns may differ from one C++ implementation to another or from one version of the same implementation to another
    3. Edit: If recreating a bug requires loading a specific save file, the person reporting the bug may assume that the save file is embedded in the recording. This appears to not be the case. The save file has to be uploaded separately
    4. Edit: similarly to 2 the source code of any programs the player loads by opening a file via the in-game editor appears to be missing
    5. Edit: for player convenience when starting a level that the player has played before the game loads the source code that it automatically saved earlier - it does not start from the same fresh state for every player on every machine.
    6. Edit: [CBot has an API for reading files](https://colobot.info/SatComOnline/master/cbot/open)
    7. Edit: This is almost certainly not an exhaustive list

PR #911, opened 2017-02-15
Remove a dead link to the dev wiki

abenkovskii:
The wiki is dead now. See https://github.com/colobot/colobot/commit/c03d8beb8bc501f5c660ed843b1205d23484e6d4


abenkovskii:
@krzys-h I think this should be merged into master too.

Issue #828, opened 2016-08-29
SIGSEGV in CBOT code when loading levels


tomangelo2:
``` gdb
Thread 1 "colobot" received signal SIGSEGV, Segmentation fault.
0x00007ffff7bae034 in CBot::CBotStack::Delete() ()
   from /usr/lib/colobot/libCBot.so
(gdb) bt
#0  0x00007ffff7bae034 in CBot::CBotStack::Delete() ()
   from /usr/lib/colobot/libCBot.so
#1  0x00007ffff7bace1d in CBot::CBotProgram::Stop() ()
   from /usr/lib/colobot/libCBot.so
#2  0x00007ffff7bad961 in CBot::CBotProgram::Compile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, void*) () from /usr/lib/colobot/libCBot.so
#3  0x000000000068f526 in CScript::Compile() ()
#4  0x000000000069291f in CScript::GetScript(Ui::CEdit*) ()
#5  0x0000000000692b0c in CScript::ReadScript(char const*) ()
#6  0x0000000000640388 in CProgramStorageObjectImpl::LoadAllProgramsForLevel(CLevelParserLine*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) ()
#7  0x00000000005355e7 in CRobotMain::CreateScene(bool, bool, bool) ()
#8  0x0000000000539d68 in CRobotMain::ChangePhase(Phase) ()
#9  0x00000000005c6c40 in Ui::CScreenLevelList::EventProcess(Event const&) ()
#10 0x00000000005ba92d in Ui::CMainUserInterface::EventProcess(Event const&) ()
#11 0x000000000053d9b9 in CRobotMain::ProcessEvent(Event&) ()
#12 0x000000000048df20 in CApplication::Run() ()
#13 0x0000000000488226 in main ()
(gdb) 
```

Noticed on Debian testing from repository and Windows binaries compiled with MSYS2.

Some time ago I've noticed this crash, seems like problem occur when in cmake Release type != Debug (yes, even MinSizeRel and RelWithDebInfo) and there are any -O flags (except -O0). But -O flags with Debug build type works fine, even -O3.
[...]

PR #955, opened 2017-05-14
Fix all compiler warnings and enable -Werror

AbigailBuccaneer:
This removes all compilation warnings on:
* g++ 4.7.4, 4.8.5, 4.9.4, 5.4.1, 6.3.0, 7.0.1-svn246759
* clang++ 3.9.1, 4.0.0, 5.0.0-svn303007

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
 
     message(STATUS "Detected GCC version 4.7+")
 
-    set(NORMAL_CXX_FLAGS "-std=gnu++11 -Wall -Wold-style-cast -pedantic-errors")
+    set(NORMAL_CXX_FLAGS "-std=gnu++11 -Wall -Werror -Wold-style-cast -pedantic-errors")
     set(RELEASE_CXX_FLAGS "-O2")
     set(DEBUG_CXX_FLAGS "-g -O0")
     set(TEST_CXX_FLAGS "-pthread")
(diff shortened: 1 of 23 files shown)


krzys-h:
Are you absolutely sure all asserts for this!=nullptr are never going to happen? I think they shouldn't, I put them there just in case because the original code actually used if()s to do different things if function was called on nullptr but I wasn't sure if I managed to cleanup all of those calls.

AbigailBuccaneer:
The warning is: `warning: 'this' pointer cannot be null in well-defined C++ code; comparison may be assumed to always evaluate to true [-Wtautological-undefined-compare]`. So I haven't checked every callsite of every function ensure that we never call a member function on a null pointer, but the compiler is smart enough to realise that if the comparison *is* false then that's undefined behaviour, which means it's free to optimise out the assert entirely. And [gcc does this at -O1 and above](https://godbolt.org/g/usWzGU).

If you'd like, I can put those checks back in and disable that specific warning under Clang. Or I can enable `-fsanitize=null -fsanitize-undefined-trap-on-error` on debug builds (for both gcc and clang), so that *every* instance of calling methods on a null pointer is caught.

krzys-h:
I'm [fully](https://github.com/colobot/colobot/commit/bb9d1c82655af26f13cbf5dc094a8bd7f0f351a9) [aware](https://github.com/colobot/colobot/commit/8764d28e9e7d168e2b15245b75977b6141dec4b0) [it](https://github.com/colobot/colobot/commit/35d60aaae58b88d314dbe2501b5ee415490a1b96) [does](https://github.com/colobot/colobot/commit/b49fbf0cd6ada33a3e0a5850bd8b7b82efd76502) (#828). It caused us some problems in the past because whoever wrote the CBot engine really loves using undefined behaviour (and there were even worse things, like using bool variables to store three possible values - 0, 1 or 2). I don't think we really need that flag because we would have probably already noticed if anything was wrong, I just wanted to know if somebody double-checked me on that or not in case we run into it in the future.

PR #963, opened 2017-05-24
Fix -Wundefined-var-template warning

AbigailBuccaneer:
This avoids specializing `CSingleton<T>::m_instance` for each type, and instead just defines it once in the header. This is allowed by the standard, multiple definitions are merged in the same way that inline functions are.

This fixes the build under `clang-5.0`, and doesn't break it under any other Clang or GCC versions I tested. I haven't tested this under MSVC, but I've found references to people using this technique fine with MSVC.

### DIFF
--- a/src/app/app.cpp
+++ b/src/app/app.cpp
@@ -61,9 +61,6 @@
 #include <getopt.h>
 #include <localename.h>
 
-
-template<> CApplication* CSingleton<CApplication>::m_instance = nullptr;
-
 char CApplication::m_languageLocale[] = { 0 };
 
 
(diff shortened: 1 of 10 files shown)


krzys-h:
oh, and @tomaszkax86 was faster because I poked him a bit earlier on Slack too :stuck_out_tongue:

hlopetz:
i did not check the full patch, tbh. i only patched [this line](https://github.com/colobot/colobot/pull/963/files#diff-660bfb8efec239adf3bc708732f21498R76) in `src/common/singleton.h`.

krzys-h:
@hlopetz should still work, assuming @AbigailBuccaneer is correct on `multiple definitions are merged in the same way that inline functions are`

PR #982, opened 2017-06-24
Some Gameplay & UI Improvements

DavivaD:
This time I will not broke it D: Sorry.

### DIFF
--- a/src/graphics/engine/particle.cpp
+++ b/src/graphics/engine/particle.cpp
@@ -955,6 +955,14 @@ void CParticle::FrameParticle(float rTime)
             if (object != nullptr && object->Implements(ObjectInterfaceType::Damageable))
             {
                 dynamic_cast<CDamageableObject*>(object)->DamageObject(DamageType::Phazer, 0.002f, m_particle[i].objFather);
+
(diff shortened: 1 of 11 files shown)


krzys-h:
Please don't put multiple things in one pull request, it makes it a lot harder to discuss and review them.

The changes look good from code style perspective, but I'd like to hear some feedback from the community about each of these changes, especially the gameplay ones.

DavivaD:
d6dabfe is extension of 656ca91

rbcat:
> Increase OrgaShooter Damage to similar to Shooter Damage
IMHO this one isn't very sensible. Maybe it should be even greater to make orga shooters into truly cool weapons, or, better yet, just stay unchanged (their damage is very low compared to that of fireball shooters now). I personally prefer the later, because it makes more sense giving that they take less energy for shooting and it still allows fireball shooters to be a сompetitor when it comes to wars between robots.

And it looks like fire can actually do more damage than organic chemicals in real life.

immibis:
If you want to rebalance orgashooters for code battles, perhaps they should need alien matter or something.

MrSimbax:
I'm closing this pull request because of 2 reasons:
1) The changes should be split into separate pull requests.
2) There are conflicting files so it needs an update anyway.

Issue #320, opened 2014-08-19
Damage alarms


MrSimbax:
Moved from #291, CHmSID:

Also, there should be more damage indication alarms like:
- flashing/beeping icon in the top left corner of the interface to show bots which currently take damage
- beeping noise(similar to low energy level) when the bot we control has low health

krzys-h:
Beeping noise is now implemented, but I leave this issue open because of the second idea

PR #992, opened 2017-07-15
Damage Alarm Implementation (2th Stage)

DavivaD:
I no longer have the nerves to write the timer to off the Damage Alarm for Shortcut.

This is working but is after taking damage this isn't changing shortcut to normal state.

### DIFF
--- a/src/ui/controls/shortcut.cpp
+++ b/src/ui/controls/shortcut.cpp
@@ -109,6 +109,12 @@ void CShortcut::Draw()
         zoom = 1.0f;
         mode = Gfx::ENG_RSTATE_NORMAL;
     }
+    if ( m_state & STATE_DAMAGE  )
+    {
+        icon = 59;
+        zoom = 0.8f;
+        mode = Gfx::ENG_RSTATE_NORMAL;
+    }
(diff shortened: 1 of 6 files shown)


melex750:
**[Here's a diff](https://www.dropbox.com/s/anhvne16swiol5b/dmgblink.diff?dl=0)** for example.
damageblink
This would make the icon blink for 6 seconds then stop.

src/ui/controls/shortcut.cpp
```diff
@@ -209,6 +209,27 @@ void CShortcut::Draw()
 
         DrawIcon(m_pos, m_dim, uv1, uv2);
     }
+    if ( (m_state & STATE_DAMAGE) && Math::Mod(m_time, 0.7f) >= 0.3f )
+    {
+      Math::Point uv1, uv2;
+      float   dp;
+
+      m_engine->SetTexture("textures/interface/button2.png");
```
krzys-h: I don't like the fact that this uses a texture for just a solid color... but it should work for now

DavivaD:
Closes #320 .

PR #994, opened 2017-07-21
Thumper & Recycler Enhancements.

### LINKED ISSUE #280: Add to Recycler possibility to restore titanium from ruins, not only wrecks.

### LINKED ISSUE #658: Allow destroy some objects by Thumper's aparature
Just like that was with destroying a stone in CeeBot-Teen game.

I think those objects should be destroyable in this way:
TitaniumOre
UraniumOre
Titanium
PowerCell
NuclearCell
TNT - makes big damage to Thumper bot itself when destroyed (I think about 2/3 or 3/4 should be fine)
KeyA,B,C,D - well, they are indestructable, but only by Alien's fire or friendly fire
BlackBox - ^ same here
OrgaMatter
AlienNest

All wreckages and ruines too. For now there's no way to clean them somehow without Recycler (in wreckages case).

DavivaD:
Closes #658 .
Closes #280 .
But still to do is Recycler Animation Enhancement for recycling big Building Ruins (see #280).

### DIFF
--- a/src/object/task/taskrecover.cpp
+++ b/src/object/task/taskrecover.cpp
@@ -376,5 +376,5 @@ bool CTaskRecover::Abort()
 
 CObject* CTaskRecover::SearchRuin()
 {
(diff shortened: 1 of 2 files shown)


immibis:
@krzys-h What's the official policy on changing gameplay? This will have a slight impact on some missions involving recyclers because there'll be a bit more titanium available.

PR #1004, opened 2017-07-22
Escape codes for strings in CBOT

melex750:
```
\a           alert bell          0x07
\b           backspace           0x08
\t           horizontal tab      0x09
\n           new line            0x0A
\v           vertical tab        0x0B
\f           form feed           0x0C
\r           carriage return     0x0D
\"           double quote        0x22
\'           single quote        0x27
\\           backslash           0x5c
\nnn         octal value         byte nnn
\xhh         hexadecimal value   byte 0xhh
\uhhhh       UTF-8 character     code point U+hhhh
\Uhhhhhhhh   UTF-8 character     code point U+hhhhhhhh
```

Also fixed some token parsing bugs:
```c++
extern void BadCharacters"'?() <--- bad characters in function name
{
	int i? = 1;  <--- bad character in variable name
	int j" = i?; <---/
	int k' = j"; <--/
	
	BadCharacters"'?();
}
```

### DIFF
--- a/po/colobot.pot
+++ b/po/colobot.pot
@@ -1770,6 +1770,24 @@ msgstr ""
 msgid "This parameter needs a default value"
 msgstr ""
 
+msgid "Missing end quote"
+msgstr ""
+
+msgid "Unknown escape sequence"
+msgstr ""
+
+msgid "Octal value out of range"
+msgstr ""
+
+msgid "Missing hex digits after escape sequence"
+msgstr ""
+
+msgid "Hex value out of range"
+msgstr ""
+
+msgid "Invalid universal character name"
+msgstr ""
+
 msgid "Dividing by zero"
 msgstr ""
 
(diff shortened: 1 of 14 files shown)


a-v-o-dev:
There's a problem with backslash.

Example to reproduce:

    string s = "\\";
    // string s = "\\";
    
Close program editor and open it again. The code lines are then:

    string s = "\";
    // string s = "\";

One backslash is lost. Even remarks/comments are affected.

Colobot Version: `git-HEAD-rfabbdda` from `dev`

melex750:
That's a bug in ```UI::CEdit``` where it parses SatCom files. IIRC it is the last 3 characters on that line ```\";``` that cause it.
Workaround: ```string s = "\\" + "";``` or maybe this too: ```string s = '"';``` or add a space before semicolon;

PR #1005, opened 2017-07-22
Mod manager (and some random UI tweaking)

DavivaD:


src/ui/screen/screen_mod_manager.cpp
```diff
@@ -0,0 +1,351 @@
...
+            pl->SetSelect(-1);
+            pb->ClearState(STATE_ENABLE);
+            pb = static_cast<CButton*>(pw->SearchControl(EVENT_INTERFACE_UNLOAD));
+            if ( pb == nullptr )  break;
+            pb->SetState(STATE_ENABLE);
+            break;
+
+        case EVENT_INTERFACE_MODS_DIR:
+            modPathRaw = CResourceManager::GetSaveLocation() + "/" + "mods";
+            #ifdef _WIN32
```
krzys-h: Don't use these defines, use [our own PLATFORM_* defines](https://github.com/colobot/colobot/blob/master/src/common/config.h.cmake#L4-L8) instead. Or, even better, proxy these calls through CSystemUtils.

krzys-h:
And again - pleeeease don't mix multiple changes in one PR. At least the mod manager should really be it's own thing.

Like we've already discussed on Slack, the mod manager cannot depend on renaming the mod files as they may be located in a read-only system-wide location. The UI should really show the filesystem location of the mod (it can be either in 'data/mods' or 'saves/mods'), and it should include mods provided from the commandline (but without the ability to disable them). Enabling/disabling mods CAN be done without restarting the game in most cases, and we should implement that.

Waiting on feedback from community about 7b0b205189c7fa3e64dc44a164ba6b3c6389c715 and a92b3a1ff077a6d691c8045736a14610e7d0fdca

src/ui/screen/screen_mod_manager.cpp
DavivaD: Done in 3221b81 .

src/ui/screen/screen_mod_manager.cpp
```diff
@@ -203,19 +203,60 @@ bool CScreenModManager::EventProcess(const Event &event)
...
+            #if defined(PLATFORM_WINDOWS)
+                result = system(("start \""+modPathRaw+"\"").c_str());
+            #elif defined(PLATFORM_LINUX)
+                result = system(("xdg-open \""+modPathRaw+"\"").c_str());
+            #elif defined(PLATFORM_MACOSX)
+                result = system(("open \""+modPathRaw+"\"").c_str());
+            #endif
+            if (result == -1)
+            {
+                GetLogger()->Error("Failed to open Mods Directory! Is directory Exists?\n");
```
tomangelo2: Shouldn't it be "Does directory exists?"

DavivaD: Right...

DavivaD:
Closing because too random things in PR + Merge Conflicts.

PR #1022, opened 2017-08-05
Fix horizontal scrolling of planets and background images

### LINKED ISSUE #1021: Background planets/stars move in opposite direction to view
Latest dev build (commit 0de347d430106ac3bf9df005aedb826076255e44), compiled myself.

When I rotate the view the stars and planets in the sky rotate in the opposite direction. This is particularly noticeable in the moon flying practice missions, but seems to happen in every level (at least every level that has these objects). Clouds are not affected.

immibis:
Fixes #1021.

With this fix, both planets and background images move horizontally in exactly the way they should. They no longer zoom across the sky when you rotate the camera.

They still move weirdly in the Y direction. For background images there's no way to make it not weird - could we switch to using skyboxes instead? For planets we probably need to treat them as full 3D objects 
instead of 2D sprites.

I've tested the outer space cutscenes for the moon, Tropica, Crystallium and Saari to make sure the planets still line up. Note Tropica's cutscene was already broken - Tropica doesn't appear in it.

### DIFF
--- a/src/graphics/engine/engine.cpp
+++ b/src/graphics/engine/engine.cpp
@@ -2631,12 +2631,26 @@ void CEngine::SetFocus(float focus)
     float farPlane = m_deepView[0] * m_clippingDistance;
 
     float aspect = static_cast<float>(m_size.x) / static_cast<float>(m_size.y);
+
+    // Compute H-FoV from V-FoV and aspect ratio.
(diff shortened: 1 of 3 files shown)


immibis:
I tried again to properly fix the planet positions using proper 3D transforms, but I couldn't get it consistent with the original game in all cases - which makes sense because the original game is inconsistent with itself).

PR #1031, opened 2017-10-08
Added fonts configurability by a separate file

tomangelo2:
For now file fonts.ini is stored next to colobot.ini

### DIFF
--- a/src/common/font_config.cpp
+++ b/src/common/font_config.cpp
@@ -17,7 +17,7 @@
  * along with this program. If not, see http://gnu.org/licenses
  */
  
-#include "common/font_file.h"
+#include "common/font_config.h"
 
 #include "common/logger.h"
 #include "common/make_unique.h"
--- a/src/common/font_config.cpp
+++ b/src/common/font_config.cpp
@@ -42,12 +42,6 @@ CFontConfig::CFontConfig()
    : m_needsSave(false)
    , m_loaded(false)
 {
-    m_font[Gfx::FONT_COMMON] = "FontCommon";
-    m_font[Gfx::FONT_COMMON_BOLD] = "FontCommonBold";
-    m_font[Gfx::FONT_COMMON_ITALIC] = "FontCommonItalic";
-    m_font[Gfx::FONT_STUDIO] = "FontStudio";
-    m_font[Gfx::FONT_STUDIO_BOLD] = "FontStudioBold";
-    m_font[Gfx::FONT_SATCOM] = "FontSatCom";
 }
 
 CFontConfig::~CFontConfig()
@@ -85,10 +79,15 @@ bool CFontConfig::Init()
 
 std::string CFontConfig::GetFont(Gfx::FontType type)
 {
-    return std::string("/fonts/") + m_propertyTree.get<std::string>(m_font[type], GetDefaultFont(type));
+    return std::string("/fonts/") + m_propertyTree.get<std::string>(GetFontType(type), GetDefaultFont(type));
 }
 
 std::string CFontConfig::GetDefaultFont(Gfx::FontType type) const
 {
     return defaultFont.at(type);
+}
+
+std::string CFontConfig::GetFontType(Gfx::FontType type) const
+{
+    return fontType.at(type);
 }
--- a/src/common/font_config.cpp
+++ b/src/common/font_config.cpp
@@ -37,17 +37,17 @@
 
 namespace bp = boost::property_tree;
 
-CFontConfigFile::CFontConfigFile()
+CFontConfig::CFontConfig()
    : m_needsSave(false)
    , m_loaded(false)
 {
 }
 
-CFontConfigFile::~CFontConfigFile()
+CFontConfig::~CFontConfig()
 {
 }
 
-bool CFontConfigFile::Init()
+bool CFontConfig::Init()
 {
     try
     {
@@ -76,7 +76,7 @@ bool CFontConfigFile::Init()
     return true;
 }
 
-std::string CFontConfigFile::GetFont(Gfx::FontType type)
+std::string CFontConfig::GetFont(Gfx::FontType type)
 {
     switch(type)
     {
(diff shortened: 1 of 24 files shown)


src/common/font_file.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+    catch (std::exception & e)
+    {
+        GetLogger()->Error("Error on parsing config file: %s\n", e.what());
+        return false;
+    }
+    return true;
+}
+
+std::string CFontConfigFile::GetCommonFont()
```
krzys-h: Why so much repetition? You could have just used a single method that takes FontType as an argument.

src/common/font_file.h
```diff
@@ -0,0 +1,92 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://gnu.org/licenses
+ */
+ 
+ /**
+ * \file common/font_file.h
```
krzys-h: This file name may be confusing, it should clearly state that this is just for a config file

src/common/font_file.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <utility>
+#include <cstring>
+#include <boost/property_tree/ini_parser.hpp>
+#include <boost/regex.hpp>
+
+namespace bp = boost::property_tree;
+
+CFontConfigFile::CFontConfigFile()
+   : m_needsSave(false)
+   , m_useCurrentDirectory(false)
```
krzys-h: useCurrentDirectory was only used for unit tests, so either add unit tests or remove it from here

src/common/font_file.h
```diff
@@ -0,0 +1,92 @@
...
+    std::string GetSatComFont();
+    
+private:
+    boost::property_tree::ptree m_propertyTree;
+    bool m_needsSave;
+    bool m_useCurrentDirectory;
+    bool m_loaded;
+};
+
+//! Global function to get config file instance
```
krzys-h: We were supposed to decrease the amount of global singletons, not add new ones. You could have easily done without.

src/common/font_config.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+    catch (std::exception & e)
+    {
+        GetLogger()->Error("Error on parsing config file: %s\n", e.what());
+        return false;
+    }
+    return true;
+}
+
+std::string CFontConfig::GetFont(Gfx::FontType type)
```
krzys-h: The problem was not that it was multiple methods, the problem was that you are copy-pasting a lot of code ;) You are repeating (almost) the same 13 lines of code 6 times.
In this exact case, I'd make a method GetDefaultFont(Gfx::FontType) that just retuns a value from constant std::map, and GetFont(Gfx::FontType) that first converts font type to name in ini file using some other map and then executes all the code you are copy-pasting and falls back to GetDefaultFont in case of an error. That makes it a lot easier to expand in the future if needed.

src/graphics/engine/text.cpp
```diff
@@ -116,19 +117,26 @@ CText::~CText()
...
-
-    m_fonts[FONT_COLOBOT]        = MakeUnique<MultisizeFont>("fonts/dvu_sans.ttf");
-    m_fonts[FONT_COLOBOT_BOLD]   = MakeUnique<MultisizeFont>("fonts/dvu_sans_bold.ttf");
-    m_fonts[FONT_COLOBOT_ITALIC] = MakeUnique<MultisizeFont>("fonts/dvu_sans_italic.ttf");
-
-    m_fonts[FONT_COURIER]        = MakeUnique<MultisizeFont>("fonts/dvu_sans_mono.ttf");
-    m_fonts[FONT_COURIER_BOLD]   = MakeUnique<MultisizeFont>("fonts/dvu_sans_mono_bold.ttf");
-
+    
+    m_fonts[FONT_COMMON]        = MakeUnique<MultisizeFont>(fontConfig.GetFont(FONT_COMMON));
```
krzys-h: Some ugly repetition again, but in this case I have no idea how to write it better as the values are bitmasks so there is no easy way to go through all of them.

tomangelo2: At first thought my idea was to get only primary font type (common, studio or satcom), and for every font type match regular, bold and italic variant. This way we would have 9 font types available in game. But I couldn't figure out how to do this without messing in text code.

krzys-h: Untested pseudocode:
```C++
for(auto type : {FONT_COMMON, FONT_EDITOR, FONT_SATCOM})
{
    Load(type);
    Load(type|FONT_BOLD);
    Load(type|FONT_ITALIC);
}
```

You can check subtype when choosing filename with (font&FONT_BOLD)!=0 etc. Other types would be unused until we add something hat needs them.

src/common/font_config.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include <boost/property_tree/ini_parser.hpp>
+#include <boost/regex.hpp>
+
+namespace bp = boost::property_tree;
+
+CFontConfig::CFontConfig()
+   : m_needsSave(false)
+   , m_loaded(false)
+{
+    m_font[Gfx::FONT_COMMON] = "FontCommon";
```
krzys-h: This could also be a const map

src/common/font_config.h
```diff
@@ -0,0 +1,98 @@
...
+#include <string>
+
+/**
+* \class CFontConfig
+*
+* \brief Class for loading config file
+*
+*/
+
+const std::map<Gfx::FontType, std::string> defaultFont =
```
krzys-h: Is there any reason to keep these in the header file? Why not just in the cpp file?

Issue #354, opened 2014-11-01
Cannot load programs from long file names


gergelypolonkai:
I have saved a program as `GotoSpaceShip` (which eventually got renamed to `GotoSpaceShip.txt`). When I try to load my program in another bot, the text field fills up with `GotoSpaceShip.tx`, and when I try to hit the OK button, according to the console, it tries to load `GotoSpaceShip.tx.txt`. If I manually remove the `.tx` part, it loads fine. This is annoying, especially that the only message about the failure goes to the console only, which, in a usual case, isn’t visible to the user.

PR #1048, opened 2017-11-15
Program sub-folders in CBOT editor open/save dialog

melex750:
I've had this for a while, thought it might be useful.

* Moved open/save dialog from CStudio to CFileDialog.
* CFileDialog is not dependent on CStudio or CScript, and
may be used as a file selector elsewhere in the game.

**Folders**
* All folders are displayed at the top of the file list.
* Folder navigation by double-click on list item.
* Ability to create sub-folders.

**Files**
* File list only shows files with specified extension(s).
* "Overwrite existing file?" confirmation.
* Typing a file name searches the file list.
* CStudio defines extensions **.cbot**(default) and **.txt**(optional).

**Other**
* Fixed wrong size calculation for the list-box.

**TODO**
* ~~More documentation for CFileDialog.~~
* ~~Allow CFileDialog to operate as a "select folder" dialog.~~
* Allow **.cbot** extension in object.factory() string argument.
* ~~Rename "programs" to "files" in CFileDialog::PopulateList()~~
* Anything else?
filedialog1
filedialog2
filedialog3


krzys-h:
Awesome, that'll definitely be useful! Will do a more complete review soon.

krzys-h:
While you're working on it, would it be possible to fix #354 more easiky with these changes?

melex750:
I think the best solution for #354 would be a lightweight version of CEdit which scrolls left/right,
I don't know how easy that would be though.

The character limit is increased when you make the dialog wider,
maybe that could be done automatically when there is not enough space.

src/ui/studio.cpp
```diff
@@ -976,6 +979,124 @@ void CStudio::SetInfoText(std::string text, bool bClickable)
     }
 }
 
+// Beginning of the display of a dialogue.
+
+void CStudio::StartDialog(const Event &event)
+{
+    if ( event.type == EVENT_STUDIO_OPEN )
+    {
+        m_dialog = MakeUnique<CFileDialog>();
```
krzys-h: I'd rather call this `m_fileDialog`, as `m_dialog` is used in other places referring to `CMainDialog`

krzys-h:
I tested the functionality, it works great! But I'll have to do a deeper code review before I merge this, I don't feel comfortable merging 1600 lines of code I didn't even look at.

One minor thing I noticed while we're working on improving the file dialog: if you manually type a nonexistent filename in the "Open" dialog, you get only a message in the log file, no feedback in the game. Perhaps we should add at least a sound feedback in that case? (I'm pretty sure something else uses SOUND_TZOING for something like that, but I don't remember what exactly)

MrSimbax:
I assume this works based on the comments above so please merge dev so that the checks are green and I'll give a positive review and merge this.

PR #1049, opened 2017-11-17
Rework new functions so they follow established syntax rules

Emxx52:
1. reverted factory() to its original design: 
    - factory(cat[, program, object]);
2. reworked syntax of other object.functions() accordingly:
    - object.busy(); is now isbusy(object);
    - takeoff(object);
    - research(type[, object]);
    - destroy(object);
3. improved compatibility with CeeBot4 SCHOOL


krzys-h:
Like I said multiple times, I think we should stay with the object-oriented version. Waiting on opinions from others.

Emxx52:
The purpose of this change is to unify CBOT syntax and ensure compatibility with CeeBot4 SCHOOL.
As we discussed earlier there is no point in changing already established functions such as factory(); to utilize a forced object-oriented solution.

An object-oriented version of the language would potentially be developed in the future and will include drastic syntax changes (i.e CBOT++).

krzys-h:
So here we go with another changes vs backwards compatibility discussion:

I think I already said it once on Slack, but I'll repeat it here: To me, CBOT was meant to be object oriented (it's based on C++/Java overall), but it was too hard to implement in the original version of CBOT library that didn't fully support classes, so it was just 'faked' with the `object::` syntax in function declarations, and with a limitation that you can't explicitly specify `this.` and it has to be implied. *(oh, where did I recently see implementing something differently because it was too hard at the moment...)* The structural versions look to me like:

```C++
extern void object::SomeProgram()
{
    this.factory(WheeledGrabber, "test.txt", radar(BotFactory));
}
```

Which makes completely no sense. It's much more logical to do:

```C++
extern void object::SomeProgram()
{
    radar(BotFactory).factory(WheeledGrabber, "test.txt");
}
```

The only functions in original game that we could argue about are `send`/`receive`/`testinfo`/`deleteinfo`. They operate on ExchangePosts, not robots, BUT it still makes sense, as in reality you are controlling the transmitter on the robot, not the ExchangePost itself.

Argument about compatibility with CeeBot is good, but I don't think we should be forced to switch to it - I don't even remember if I knew about the existence of these functions in CeeBot when implementing factory(), which was the 'first object-oriented function' (or 'first function to operate on something else than this' if you agree with my theory). It would be nice to keep compatibility between them, but we went in different directions with no good way to connect them again (and we don't want to have both to avoid mess in the language)

src/script/cbottoken.cpp
```diff
@@ -99,6 +99,10 @@ const char* GetObjectName(ObjectType type)
     if ( type == OBJECT_MOBILEts    )  return "TrackedSniffer";
     if ( type == OBJECT_MOBILEfs    )  return "WingedSniffer";
     if ( type == OBJECT_MOBILEis    )  return "LeggedSniffer";
+    if ( type == OBJECT_MOBILEwb    )  return "WheeledBuilder";
```
krzys-h: Oh, and this PR doesn't even compile :stuck_out_tongue:

Emxx52:
>send/receive/testinfo/deleteinfo. They operate on ExchangePosts, not robots, BUT it still makes sense, as in reality you are controlling the transmitter on the robot, not the ExchangePost itself.

That's right, because buildings are not programmable, factory() and any other function reflects that now, as you are telling the bot to use it's transmitter to do sth (i.e. to tell BotFactory to start building a bot).

krzys-h:
ExchangePosts work using radio transmission. And the radio transmission technology in the game is clearly limited to one transmission per second, aa you can see in send/receive. Everything else works by some undefined magical force.
And what are static variables in public classes? Quantum-entangled RAM modules?
Not every game feature needs to have a logical explanation in in-game world. I'm perfectly fine with this being just asking the object to do something using the same magical force that makes the button in the UI start the factory process.

And just so you know, changing pull request title to suggest that it fixes something that was clearly wrong is just stupid.

Emxx52:
You're the one who changed the title of _my_ pull request first. I merely changed it back to sth that conveys _my_ feelings and _my_ opinion on the subject.

tomangelo2:
> You're the one who changed the title of my pull request first

Because originally it said only `dev-cbot-factory to dev `, which isn't much explaining what this PR is doing.
Or at least GitHub doesn't show me true previous title

krzys-h:
I changed it from something that was meaningless without reading the description to something that describes what exactly this PR does in as neutral way as possible. You changed it to say what you want to force it to be. And saying 'established syntax' when we are still using object-oriented approach is the exact opposite of what you really want to say...

Are we seriously going to argue about title of a pull request now?

immibis:
Well he's not wrong, none of CBOT is object-oriented in the vanilla game.

tomangelo2:
So, there are 3 solutions:

1. We imply that it was from the beginning structural (even if Cbot was based on OOP languages) and just merge this. 
2. We imply that it was from the beginning objected with hidden `this.` prefix as a workaround for only partial support of classes in previous versions, fix detection of `this.*` with default methods and go further with OOP methods. Such taking `this.` prefix by default works already for variables: `id == this.id;`, `position == this.position;` etc., so `grab();` might be also equal to `this.grab();`.
3. Make a language switcher (either in UI or by inserting/removing some word/command, divide Cbot into Cbot and Cbot++ and support 2 similar languages with different ideas (and have fun with testing and debugging 2 languages, unless current code would somehow support such schism without having to take care about 2 separate parts of code for every issue).

Whatever of which option wins, there is also question about current `this.` behaviour. In first case will we just delete that? And in second force it (or just keep optional like now, unless it won't cause too much mess)?

krzys-h:
@immibis But it didn't have commands operating on something else than the robot itself either

immibis:
@krzys-h 

That is true, so your way also makes sense (as new syntax for a new feature).

I'm not a fan of magical invisible communication in this game, so I'd like if eventually `factory` etc. worked like `send` (it can even reuse the same animation).

"Controlling the object" vs "controlling the transmitter" is a pointless classification - of course every piece of code running on the robot *should be* controlling the robot - nothing else makes logical sense. Cheats, and level scripts (when they're added) can control objects directly.

immibis:
I would also point out that the base game's ExchangePost functions (send/receive/testinfo/deleteinfo) are not object-oriented. If we follow this precedent then `factory` would have to operate on the closest factory to the robot.

src/script/cbottoken.cpp
MrSimbax: Now it compiles.

PR #1053, opened 2017-11-22
added a thing

krzys-h:
Hey, I added a thing. Plz accept.


krzys-h:
https://slack-files.com/T2T2124TU-F840DTB6V-f6972619c6
(by @DavivaD)

immibis:
It adds `jestemPAWIEM` as a synonym for `iPAW` and `jestemPISTOLETEM` as a synonym for `iGUN` in `EnableResearch`.

The base64 block is a low-resolution JPEG of a Slack message in Polish.

immibis:
Yeah, I don't think it's sensible to add joke stuff into the map file format @krzys-h 

Imagine that we want to get rid of it 2 years later but we can't because maps are using it.

PR #1058, opened 2017-12-21
Update Trainer (Practice) bots, trainer=1 objects, and TargetBot

Emxx52:
1. Changed _Trainer_ and _trainer=1_ bots:
   - _TrackedTrainer_, _WingedTrainer_, _LeggedTrainer_ full implementation;
   - _PracticeBot_ is now an alias for _WheeledTrainer_, _TrackedTrainer_, _WingedTrainer_, or _LeggedTrainer_;
   - new models
2. Added _trainer=1_ buildings (i.e. _BotFactory_, _ResearchCenter_);
3. _TargetBot_ research;
4. _Amphibious_ and _Heavy_ trainer bots;


krzys-h:
Sorry that I didn't have time to follow what was goimg on with rhis on Slack, but you gave a general :+1: from me on these changes. A few questions though:
1. What exactly changed about trainer=1? I see you are using GetTrainer() in a lot of places. From a quick look it seems that it changes something in the models, if so some screenshots in a PR like that would be nice.
2. You should at least mention that these changes require a different branch from data (or even make a PR there and link it here)
3. trainer=1 for buildings - :+1:
4. What do we need Amphibious and BigTracked for? I can see uses for the other ones, but not for these. Amphibious is just Subber and BigTracked is essentialy jest Tracked, there is no reason to use the big version.
5. TargetBot research - I'm pretty sure I saw a mention of what this could be used for but I can't remember exactly. Isn't a research for that an overkill though?

src/level/parser/parserparam.cpp
```diff
@@ -339,6 +339,10 @@ ObjectType CLevelParserParam::ToObjectType(std::string value)
     if (value == "Portico"           ) return OBJECT_PORTICO;
     if (value == "SpaceShip"         ) return OBJECT_BASE;
     if (value == "PracticeBot"       ) return OBJECT_MOBILEwt;
```
krzys-h: This should seriously change to be the 'alias', but it seems impossible without breaking levels. How about at least a warning for the sake of 'unification'?

src/object/object_type.h
```diff
@@ -133,6 +133,7 @@ enum ObjectType
     OBJECT_MOBILEtb         = 221,  //!< TrackedBuilder
     OBJECT_MOBILEfb         = 222,  //!< WingedBuilder
     OBJECT_MOBILEib         = 223,  //!< LeggedBuilder
+    OBJECT_MOBILEpr         = 224,  //!< PracticeBot (alias)
```
krzys-h: If you seriously have to add an object type that does nothing and is used just as an alias (and I see no better option of doing it), at least make it stand out so that nobody tries to use it as object type. Maybe add ALIAS to the name and give it some big (or negative even) numeric value?

src/script/cbottoken.cpp
```diff
@@ -210,6 +214,9 @@ std::string GetHelpFilename(ObjectType type)
     if ( type == OBJECT_MOBILErs    )  helpfile = "object/botshld";
     if ( type == OBJECT_MOBILEsa    )  helpfile = "object/botsub";
     if ( type == OBJECT_MOBILEwt    )  helpfile = "object/bottr";
+    if ( type == OBJECT_MOBILEtt    )  helpfile = "object/bottr";
+    if ( type == OBJECT_MOBILEft    )  helpfile = "object/bottr";
+    if ( type == OBJECT_MOBILEit    )  helpfile = "object/bottr";
```
krzys-h: Missing a help filename for OBJECT_MOBILEpr

src/script/scriptfunc.cpp
```diff
@@ -984,16 +984,30 @@ bool runRadar(CBotVar* var, std::function<bool(std::vector<ObjectType>, float, f
     {
         while ( array != nullptr )
         {
+            if (array->GetValInt() == OBJECT_MOBILEpr)
+            {
+                type_v.push_back(OBJECT_MOBILEwt);
```
krzys-h: Wrong place, this should go in CObjectManager::Radar so that it also works in other functions like search() and level files

src/object/object_type.h
krzys-h: Note that all other aliases use one of the existing values as a 'base' value

src/level/parser/parserparam.cpp
Emxx52: Keeping this in Scene is intentional, you can use both PracticeBot and WheeledTrainer for the sake of compatibility

src/script/cbottoken.cpp
Emxx52: There is no way anyone could access such helpfile, as this is not really an object.

src/script/scriptfunc.cpp
Emxx52: I've been trying _stuff_ in object_manager. Just wanted to get out of there, heh. You're right, this is not a perfect solution, gotta work on that.

Emxx52:
1. trainer=1 changes chassis now, simple as that. The chassis model and its attachments have been separated, trainer.mod is the main thing, trainera.mod is the tool holder, trainerw.mod etc. are drive-specific attachments (such as wings and whatnot).
2. ~
3. ~
4. There has been a Slack discussion about that. I assumed we'd make some driving-licence-typeofadeal Excercises in the future or something.
5. TNT from m6.1 is supposed to be analysed by AutoLab in the next mission. I may elaborate on that once the feature is done.

src/script/scriptfunc.cpp
Emxx52: Yeah, CObjectManager doesn't want to work with that. I'll just add the same code for select() and detect() - it's simpler this way, and the game won't need to go through the same vector multiple times. I think this is the better solution. Give me a minute.

src/level/parser/parserparam.cpp
krzys-h: I know, I'm just suggesting to deprecate one of them with a warning because it's confuzing

src/script/cbottoken.cpp
krzys-h: Type PracticeBot in the editor and click on the name on status bar - I'm pretty sure that opens files from this list

src/script/scriptfunc.cpp
Emxx52: Yeah, I'm sorry about that, but I can't seem to make it work without greater changes in both scriptfunc and object_manager. We should probably rework this whole system in the future tho.

src/script/cbottoken.cpp
Emxx52: Ah, yes, of course. Forgot about that.

src/common/restext.cpp
```diff
@@ -626,7 +628,7 @@ void InitializeRestext()
     stringsErr[ERR_BASE_DLOCK]      = TR("Doors blocked by a robot or another object");
     stringsErr[ERR_BASE_DHUMAN]     = TR("You must get on the spaceship to take off");
     stringsErr[ERR_LABO_NULL]       = TR("Nothing to analyze");
-    stringsErr[ERR_LABO_BAD]        = TR("Analyzes only organic matter");
+    stringsErr[ERR_LABO_BAD]        = TR("Inappropriate sample");
```
krzys-h: Just looking through old pull requests, you didn't run `make update-pot` after changing restext.cpp and I still didn't add a comment in this file explaining that you have to to that :P

src/script/scriptfunc.cpp
MrSimbax: I'm resolving this conversation since this pull request is already old and the code can be always be refactored later.

src/level/parser/parserparam.cpp
MrSimbax: I don't think deprecation is needed and I don't really see it as confusing. Besides, a warning can always be added later and since the pull request is already too old, I'm resolving this conversation.

src/common/restext.cpp
MrSimbax: It's now done so I'm resolving this.

src/object/object_type.h
MrSimbax: Seems that nobody knows what do you mean by other aliases and base value. Since this PR is already too old and the issue seems minor, I'm resolving this conversation.

Issue #1113, opened 2018-02-15
[CLang build error] 1st compilation error (dev branch, release) - missing "override" keyword


B-CE:
Hi,

on my first compilation attempt, i got a compilation error : 
 "override" keyword seams mandatory even if specified as optional
http://en.cppreference.com/w/cpp/language/override 

 but its because Wall option turned it on.

I don't know if you plan to fix it, in the meantime, 
workaround is to specify "--Wno-inconsistent-missing-override" to compiler.

L148 of master  CMakeLists.txt  (also doable on L157 that ignore some other warnings)
=>    set(NORMAL_CXX_FLAGS "-std=c++11 -Wall -Werror -Wold-style-cast -pedantic-errors -Wno-inconsistent-missing-override")

config : xUbuntu 17.10 - CLang (1:4.0-37) 

Details : 
$ make
```

[ ... ]
[ 29%] Building CXX object src/CMakeFiles/colobotbase.dir/graphics/engine/particle.cpp.o
In file included from ~~~/src/graphics/engine/particle.cpp:43:
In file included from ~~~/src/object/subclass/shielder.h:22:
In file included from ~~~/src/object/subclass/base_robot.h:22:
In file included from ~~~/src/object/subclass/base_vehicle.h:22:
~~~/src/object/old_object.h:240:17: error: 'SetDamaging' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
     void        SetDamaging(bool damaging);
                ^
~~~/src/object/interface/damageable_object.h:65:25: note: overridden virtual function is here
    virtual void        SetDamaging(bool damaging) = 0;
                        ^
1 error generated.
src/CMakeFiles/colobotbase.dir/build.make:1070 : « src/CMakeFiles/colobotbase.dir/graphics/engine/particle.cpp.o » failed
make[2]: *** [src/CMakeFiles/colobotbase.dir/graphics/engine/particle.cpp.o] Erreur 1
CMakeFiles/Makefile2:186 : « src/CMakeFiles/colobotbase.dir/all » failed
make[1]: *** [src/CMakeFiles/colobotbase.dir/all] Error 2
Makefile:151 : « all » failed
make: *** [all] Error 2
```

Issue #958, opened 2017-05-22
Unable to compile Gold with -Werror flag


tomangelo2:
```gcc
$ make
[  0%] Built target localename
[  1%] Building CXX object src/CMakeFiles/colobotbase.dir/common/resources/inputstreambuffer.cpp.obj
C:/Users/student/repo/colobot/src/common/resources/inputstreambuffer.cpp: In member function 'virtual std::basic_streambuf<char>::int_type CInputStreamBuffer::underflow()':
C:/Users/student/repo/colobot/src/common/resources/inputstreambuffer.cpp:82:32: error: 'PHYSFS_sint64 PHYSFS_read(PHYSFS_File*, void*, PHYSFS_uint32, PHYSFS_uint32)' is deprecated [-Werror=deprecated-declarations]
     PHYSFS_sint64 read_count = PHYSFS_read(m_file, m_buffer.get(), sizeof(char), m_bufferSize);
                                ^~~~~~~~~~~
In file included from C:/Users/student/repo/colobot/src/common/resources/inputstreambuffer.h:27:0,
                 from C:/Users/student/repo/colobot/src/common/resources/inputstreambuffer.cpp:20:
C:/msys64/mingw32/include/physfs.h:1326:27: note: declared here
 PHYSFS_DECL PHYSFS_sint64 PHYSFS_read(PHYSFS_File *handle,
                           ^~~~~~~~~~~
C:/Users/student/repo/colobot/src/common/resources/inputstreambuffer.cpp:82:94: error: 'PHYSFS_sint64 PHYSFS_read(PHYSFS_File*, void*, PHYSFS_uint32, PHYSFS_uint32)' is deprecated [-Werror=deprecated-declarations]
     PHYSFS_sint64 read_count = PHYSFS_read(m_file, m_buffer.get(), sizeof(char), m_bufferSize);
                                                                                              ^
In file included from C:/Users/student/repo/colobot/src/common/resources/inputstreambuffer.h:27:0,
                 from C:/Users/student/repo/colobot/src/common/resources/inputstreambuffer.cpp:20:
C:/msys64/mingw32/include/physfs.h:1326:27: note: declared here
 PHYSFS_DECL PHYSFS_sint64 PHYSFS_read(PHYSFS_File *handle,
                           ^~~~~~~~~~~
cc1plus.exe: all warnings being treated as errors
[...]

PR #1114, opened 2018-02-18
Fix#1113 - a missing override keyword

B-CE:
just activated compilation option  -Wno-inconsistent-missing-override as described in #1113, like #958 fix.

### DIFF
--- a/src/object/old_object.h
+++ b/src/object/old_object.h
@@ -237,7 +237,7 @@ public:
     void        SetMagnifyDamage(float factor) override;
     float       GetMagnifyDamage() override;
 
-    void        SetDamaging(bool damaging);
+    void        SetDamaging(bool damaging) override;
     bool        IsDamaging()  override;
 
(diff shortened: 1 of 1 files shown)


melex750:
What happens if you add the missing override keyword to the declaration at [line 240 old_object.h](https://github.com/colobot/colobot/blob/master/src/object/old_object.h#L240) ?

B-CE:
Oh, I didn't tested it in fact ^^ Very good remark !!
Due to this omission, I guessed there were lots of similar anywhere, so as I was afraid to modify lots of codes everywhere before having compiled it...
Effectively all other "override" are well present, I didn't noticed at that moment... 
So, my apologies.  It works with this little modification (just re-tested after make clean on my local). I'll resubmit it as you  suggested.

Erihel:
See my comment from #1123. You can close this issue or change it to have just single line fix in this pull request.

PR #1129, opened 2018-03-16
Add realtime scoreboard sorting

tomangelo2:


src/level/scoreboard.h
```diff
@@ -32,6 +32,22 @@
...
+ * \struct Score
+ * \brief Struct containing score of individual team and additional variables to allow sorting teams through different criteria
+*/
+struct Score
+{
+    int points = 0; //! Team score
+    float time = 0; //! Time when points were scored 
+};
+
+enum SortType
```
krzys-h: New code should probably prefer using strongly-typed enums (`enum class`) over traditional C enums, but I think it can stay like that for now

src/level/parser/parserparam.h
```diff
@@ -108,6 +109,7 @@ class CLevelParserParam
     Gfx::EngineObjectType AsTerrainType(Gfx::EngineObjectType def);
     int AsBuildFlag(int def);
     int AsResearchFlag(int def);
+    int AsSortType(int def);
```
krzys-h: It would be probably better to return an enum rather than a flat int here

PR #1136, opened 2018-04-06
Some CBot code optimizations

krzys-h:
I just had a random idea to run CBot under a profiler to see what would happen and managed to reduce compilation times a bit with some trivial changes. It's probably not really noticeable unless you have some crazy huge programs or crazy slow CPU that would take a few seconds to compile anyway, but it's an improvement nevertheless. And there is still room for more if anybody has time to dig deeper into that.

Before:
```
╭─krzys_h@krzysh-laptop ~/Pulpit/colobot/build  ‹dev› 
╰─➤  for i in 64 128 256 512 1024 2048 4096; do echo -n "$i: "; python gentest.py $i | time ./CBot_console >/dev/null 2>&1; done
64: ./CBot_console > /dev/null 2>&1  0,02s user 0,00s system 87% cpu 0,027 total
128: ./CBot_console > /dev/null 2>&1  0,04s user 0,00s system 94% cpu 0,042 total
256: ./CBot_console > /dev/null 2>&1  0,08s user 0,01s system 96% cpu 0,089 total
512: ./CBot_console > /dev/null 2>&1  0,22s user 0,01s system 98% cpu 0,235 total
1024: ./CBot_console > /dev/null 2>&1  0,82s user 0,01s system 99% cpu 0,834 total
2048: ./CBot_console > /dev/null 2>&1  3,65s user 0,01s system 99% cpu 3,664 total
4096: ./CBot_console > /dev/null 2>&1  14,24s user 0,06s system 99% cpu 14,294 total
```

After:
```
╭─krzys_h@krzysh-laptop ~/Pulpit/colobot/build  ‹dev-cbot-optimizations› 
╰─➤  for i in 64 128 256 512 1024 2048 4096; do echo -n "$i: "; python gentest.py $i | time ./CBot_console >/dev/null 2>&1; done
64: ./CBot_console > /dev/null 2>&1  0,02s user 0,00s system 86% cpu 0,023 total
128: ./CBot_console > /dev/null 2>&1  0,03s user 0,00s system 91% cpu 0,038 total
256: ./CBot_console > /dev/null 2>&1  0,06s user 0,00s system 96% cpu 0,071 total
512: ./CBot_console > /dev/null 2>&1  0,19s user 0,00s system 99% cpu 0,196 total
1024: ./CBot_console > /dev/null 2>&1  0,71s user 0,02s system 99% cpu 0,731 total
2048: ./CBot_console > /dev/null 2>&1  2,86s user 0,03s system 99% cpu 2,893 total
4096: ./CBot_console > /dev/null 2>&1  11,10s user 0,06s system 99% cpu 11,158 total
```

Testing on Release build with the following input:
```python
import sys
import re

template = """
int fact_iter(?)(int n) {
	int r = 1;
	for(int i = 2; i <= n; i++) {
		r *= i;
	}
	return r;
}

int fact_rec(?)(int n) {
	if (n == 0) {
		return 1;
	} else {
		return n * fact_rec(?)(n-1);
	}
}

extern void test(?)() {
	message(""+(fact_iter(?)(10) == fact_rec(?)(10)));
}
"""

for i in range(int(sys.argv[1])):
	print(re.sub(r'\(\?\)', str(i), template))
```

### DIFF
--- a/src/CBot/CBotVar/CBotVar.cpp
+++ b/src/CBot/CBotVar/CBotVar.cpp
@@ -378,7 +378,7 @@ void CBotVar::SetInit(CBotVar::InitType initType)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-std::string CBotVar::GetName()
+const std::string& CBotVar::GetName()
 {
     return    m_token->GetString();
 }
--- a/src/CBot/CBotVar/CBotVar.cpp
+++ b/src/CBot/CBotVar/CBotVar.cpp
@@ -49,7 +49,7 @@ namespace CBot
 long CBotVar::m_identcpt = 0;
 
 ////////////////////////////////////////////////////////////////////////////////
-CBotVar::CBotVar( )
+CBotVar::CBotVar( ) : m_token(nullptr)
 {
     m_pMyThis = nullptr;
     m_pUserPtr = nullptr;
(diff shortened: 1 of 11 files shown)


src/CBot/CBotVar/CBotVar.cpp
```diff
@@ -378,7 +378,7 @@ void CBotVar::SetInit(CBotVar::InitType initType)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-std::string CBotVar::GetName()
+const std::string& CBotVar::GetName()
 {
     return    m_token->GetString();
```
AbigailBuccaneer: This change could cause dangling reference bugs in the future. It's currently fine as `m_token` is never reassigned to, but you have to read the rest of the code to know that. Perhaps declare `CBotVar::m_token` as `CBotToken* const` instead of `CBotToken*`?

PR #1139, opened 2018-04-19
Fix building under clang-7

AbigailBuccaneer:
Currently the build fails because of `-Wdelete-non-virtual-dtor` warnings.
This catches when an object is destructed, has a non-virtual destructor,
and is an abstract base class or a non-final class with virtual
functions. The warning happens inside `unique_ptr<T>::~unique_ptr`.

The warning is to prevent somebody writing code like this:

```C++
class MySceneEndCondition : public CSceneEndCondition {
    ~MySceneEndCondition() { /* some complex logic */ }
};
// this won't call MySceneEndCondition's destructor, potentially
// leading to leaks or segfaults:
std::unique_ptr<CSceneEndCondition> p{new MySceneEndCondition()};
```

### DIFF
--- a/src/level/scene_conditions.h
+++ b/src/level/scene_conditions.h
@@ -85,7 +85,7 @@ protected:
  * \class CSceneEndCondition
  * \brief Scene end condition
  */
-class CSceneEndCondition : public CSceneCondition
+class CSceneEndCondition final : public CSceneCondition
 {
 public:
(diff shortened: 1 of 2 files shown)


AbigailBuccaneer:
This is ready for merge but I'm unable to merge this myself; you (or another developer) will have to merge this.

PR #1140, opened 2018-04-19
Compile with -Wsuggest-override under GCC

AbigailBuccaneer:
Clang by default compiles with -Winconsistent-missing-override, which
warns when a class declares virtual functions that override those in the
base class, and some but not all of them are explicitly declared
`override`.

GCC doesn't support this option, but has a stronger version,
-Wsuggest-override. In combination with -Werror, this means that any
virtual function that overrides another *must* be explicitly declared as
`override`.

This commit enables -Wsuggest-override where available. This means that
GCC users can't break the Clang build with inconsistent overrides (see
 #1113 and #1114) and consequently that any build that passes the pull
request CI build on Jenkins won't break because of inconsistent
overrides.

### DIFF
--- a/test/unit/CBot/CBotToken_test.cpp
+++ b/test/unit/CBot/CBotToken_test.cpp
@@ -27,12 +27,12 @@ using namespace CBot;
 class CBotTokenUT : public testing::Test
 {
 public:
-    void SetUp()
+    CBotTokenUT()
     {
         CBotProgram::Init();
     }
 
-    void TearDown()
+    ~CBotTokenUT()
     {
         CBotProgram::Free();
     }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -134,6 +134,11 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
 
     set(NORMAL_CXX_FLAGS "-std=gnu++11 -Wall -Werror -Wold-style-cast -pedantic-errors")
     set(NORMAL_CXX_FLAGS "${NORMAL_CXX_FLAGS} -Wno-error=deprecated-declarations") # updated version of physfs is not available on some platforms so we keep using deprecated functions, see #958
+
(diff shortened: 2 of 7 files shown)


test/unit/CBot/CBotToken_test.cpp
```diff
@@ -27,12 +27,12 @@ using namespace CBot;
 class CBotTokenUT : public testing::Test
 {
 public:
-    void SetUp()
```
AbigailBuccaneer: I chose to switch from `SetUp/TearDown` to constructor/destructor, as Google's official advice is that [the latter is preferred](https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#should-i-use-the-constructordestructor-of-the-test-fixture-or-the-set-uptear-down-function).

AbigailBuccaneer:
The Jenkins build seems to have failed due to `colobot-lint` having a segfault. Should I be worried about that?

krzys-h:
I noticed. I was trying to reenable colobot-lint because it's based on clang and it should help detect these errors, but it fails for some reason. No need to worry, I'll look into that later.

krzys-h:
I think I figured out what was wrong with my scripts causing colobot-lint to fail, it was trying to pass the gcc version of the options instead of clang ones and there is apparently a bug which causes it to segfault if you do that. We'll see after Jenkins rebuilds it.

Which version of clang did you see these warnings on? I was trying to reproduce that to make sure colobot-lint would warn us if we break clang builds again (because it uses it for parsing anyway), but that warning doesn't appear even if I pass `-Winconsistent-missing-override` manually, probably because the version is too old (colobot-lint uses Clang 3.6 iirc, and probably won't be updated unless @piotrdz has time to do that)

AbigailBuccaneer:
This was found using Clang 7, which isn't yet released - though I seem to remember that the warning exists in previous versions too. If you'd like to test with it yourself on Debian or Ubuntu, https://apt.llvm.org/ provides a repository with packages that let you easily install multiple versions of Clang alongside each other (and alongside your system compiler).

PR #1141, opened 2018-04-20
Use C++11 threading and timing constructs

AbigailBuccaneer:
This replaces various platform-specific bits of code (whether implemented by colobot or SDL) with cross-platform C++11 code:

* Use `std::thread` instead of `SDL_Thread`
* Use `std::mutex` instead of `SDL_Mutex`
* Use `std::condition_variable` instead of `SDL_Cond`
* Use `std::chrono::high_resolution_clock` instead of platform-specific timer functions


krzys-h:
So, according to Jenkins MXE builds don't like std::thread... I think this might have been the reason we didn't use it in the first place. There may be a way to enable it in MXE somehow though, I haven't updated that thing in a while.

immibis:
+1, also I think the CSystemUtils timestamp functions (other than GetCurrentTimestamp, because of unit tests) should be moved outside of CSystemUtils as they are now stateless.

MrSimbax:
There are conflicting files unfortunately.

If it doesn't work for MXE then we should wait before merging this until the process of updating the building process https://github.com/colobot/colobot/issues/1306 is complete.  Therefore I'm not asking to update this PR yet because it may need more updates later after those changes anyway.

We could solve it in 2 ways: fix the issue with MXE, or drop MXE support and build the Windows binaries without cross compilation in our CI. I believe the second option may be a real possibility if we manage to move to GitHub Actions.

tomangelo2:
IIRC it was possible to fix MXE errors by rebuilding it. By default MXE is build with flag `MXE_TARGETS=i686-w64-mingw32.static`, what we need to do to make it work is to rebuild MXE with `MXE_TARGETS=i686-w64-mingw32.static.posix`.

MrSimbax:
Hmm, if that's true then maybe you could try to make it build @tomangelo2 ? There are still conflicting files in this PR unfortunately which need to be fixed though, but the sooner we can merge this the better.

tomangelo2:
`Experimental support for GCC with posix threads was added in November 2015. Since January 2019 it is used by default.`
So it seems MXE supports C++ threads even without rebuilding it with custom `MXE_TARGETS` flag. After resolving merge conflicts it should be ready to go.

Issue #1104, opened 2018-02-03
A lot of text in editor cause FPS drop


tomangelo2:
Noticed when investigating issue #1065 
![lorem_pusto](https://user-images.githubusercontent.com/4140449/35767970-7dc42858-08f5-11e8-8419-358d64fb652f.png)
![lorem_pelno](https://user-images.githubusercontent.com/4140449/35768060-a13347fa-08f6-11e8-987b-e276c40964b4.png)

Resizing editor window helps a bit, it looks like rendering text is heavier than rendering 3D world in the background. It occurs on all graphic engines.

RaptorParkowsky:
It works well on my computer until I minimize the font to the smallest size possible. From stable 60 fps it drops to 30fps or lower.

tomangelo2:
Do you have enabled VSync, or another FPS limiter?

RaptorParkowsky:
I don't think so. My GeForce GTX 750 Ti just doesn't render more than 60fps in any game. My monitors doesn't support any VSync technologies I guess.

tomangelo2:
I think you have enabled VSync globally in Nvidia Control Panel. GTX 750Ti isn't fastest card in the world, but it shouldn't be slow enough to display >60 fps in Colobot. And basic VSync should work on all displays.
Of course you don't need to disable it globally, there are reasons why GPU don't need to work on 100% power all the time, pointing it just in case.

RaptorParkowsky:
It's just default auto settings depends of 3D application. When I switched it off, GOLD rendered ~150-300 or ~700fps in menu. But I can't see those frames on my monitors, because they're all 60-75Hz. This is overkill and I doubt it have anything to do with this issue. Framedrops are the same with or without this feature.

![schowek01](https://user-images.githubusercontent.com/4019345/37297897-2a82cea0-261f-11e8-87f3-7564292bcb45.png)

tomaszkax86:
OpenGL devices have DrawPrimitives() methods which use glMultiDrawArrays(). With DrawPrimitives() you can render multiple characters as triangle fans with a single draw call, as long as they are all in one texture. This is almost the same as using quads, except that quads are not supported in OpenGL 3.3. I added it to speed things up but I never got to rewriting text renderer.

PR #1148, opened 2018-04-27
Improve text rendering performance

AbigailBuccaneer:
This pull request significantly improves the performance of text rendering, fixing #1104.

Previously, each character rendered would update the world, view and projection matrices twice each, and make a draw call for a rectangle to draw that individual character.

Now, we only update the matrices once per call to `CText::DrawText`, and we batch as many rectangles as possible into a single draw call. If the text has no highlights and no icons, the entire string will be drawn in a single draw call.

On my computer, previously, a full screen of text in the program editor would drop the framerate to under 30. Now the framerate isn't affected, and stays at vsync at 60.

Thanks to @tomaszkax86 for pointing out the existence of `CDevice::DrawPrimitives` to do this.

### DIFF
--- a/src/graphics/engine/text.cpp
+++ b/src/graphics/engine/text.cpp
@@ -691,6 +691,7 @@ void CText::DrawString(const std::string &text, std::vector<FontMetaChar>::itera
                        float size, Math::IntPoint pos, int width, int eol, Color color)
 {
     m_engine->SetState(ENG_RSTATE_TEXT);
+    m_engine->SetWindowCoordinates();
 
     int start = pos.x;
 
@@ -776,6 +777,7 @@ void CText::DrawString(const std::string &text, std::vector<FontMetaChar>::itera
         color = Color(1.0f, 0.0f, 0.0f);
         DrawCharAndAdjustPos(ch, font, size, pos, color);
     }
+    m_engine->SetInterfaceCoordinates();
 }
 
 void CText::StringToUTFCharList(const std::string &text, std::vector<UTF8Char> &chars)
@@ -847,10 +849,12 @@ void CText::DrawString(const std::string &text, FontType font,
 
     std::vector<UTF8Char> chars;
     StringToUTFCharList(text, chars);
+    m_engine->SetWindowCoordinates();
     for (auto it = chars.begin(); it != chars.end(); ++it)
     {
         DrawCharAndAdjustPos(*it, font, size, pos, color);
     }
+    m_engine->SetInterfaceCoordinates();
 }
 
 void CText::DrawHighlight(FontMetaChar hl, Math::IntPoint pos, Math::IntPoint size)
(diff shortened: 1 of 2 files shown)


tomangelo2:
obraz
There is indeed a difference (left one is with this PR, probably capped with VSync). GL calls dropped by half. Finally 2D text rendering isn't significantly heavier than 3D world rendering.

src/graphics/engine/text.cpp
```diff
@@ -94,6 +94,76 @@ const Math::IntPoint REFERENCE_SIZE(800, 600);
 const Math::IntPoint FONT_TEXTURE_SIZE(256, 256);
 } // anonymous namespace
 
+/// The QuadBatch is responsible for collecting as many quad (aka rectangle) draws as possible and
+/// sending them to the CDevice in one big batch. This avoids making one CDevice::DrawPrimitive call
+/// for every CText::DrawCharAndAdjustPos call, which makes text rendering much faster.
+/// Currently we only collect textured quads (ie. ones using Vertex), not untextured quads (which
+/// use VertexCol). Untextured quads are only drawn via DrawHighlight, which happens much less often
+/// than drawing textured quads.
+class CText::QuadBatch
```
krzys-h: Clang error: `'QuadBatch' defined as a class here but previously declared as a struct`

src/graphics/engine/text.cpp
```diff
@@ -94,6 +94,76 @@ const Math::IntPoint REFERENCE_SIZE(800, 600);
...
+    void Add(Vertex vertices[4], unsigned int texID, EngineRenderState renderState, Color color)
+    {
+        if (texID != m_texID || renderState != m_renderState || color != m_color)
+        {
+            Flush();
+            m_texID = texID;
+            m_renderState = renderState;
+            m_color = color;
+        }
+        m_quads.emplace_back(Quad{vertices[0], vertices[1], vertices[2], vertices[3]});
```
krzys-h: Clang error: `suggest braces around initialization of subobject`

src/graphics/engine/text.cpp
krzys-h: Code style: `Class 'QuadBatch' should be named in a style like CUpperCamelCase`

src/graphics/engine/text.cpp
```diff
@@ -94,6 +94,76 @@ const Math::IntPoint REFERENCE_SIZE(800, 600);
...
+    /// Draw all pending quads immediately.
+    void Flush()
+    {
+        if (m_quads.empty()) return;
+
+        m_engine.SetState(m_renderState);
+        m_engine.GetDevice()->SetTexture(0, m_texID);
+
+        assert(m_firsts.size() == m_counts.size());
+        if (m_firsts.size() < m_quads.size()) {
```
krzys-h: Code style: `Body of declaration or statement begins in a style that is not allowed`

PR #1150, opened 2018-04-29
Add generic debug rendering functions

AbigailBuccaneer:
Currently the engine can draw debug spheres to show crash sphere
positions. This extends this to draw arbitrary spheres and cuboids,
transformed arbitrarily. With these in place it's now very quick and
easy to create a debug visualisation - for example, it's a one-line code
change to render the bounding box or sphere of every EngineObject.

### DIFF
--- a/src/graphics/engine/engine.cpp
+++ b/src/graphics/engine/engine.cpp
@@ -3147,24 +3147,6 @@ void CEngine::ApplyChange()
     }
 }
 
-void CEngine::ClearDisplayCrashSpheres()
-{
-    m_displayCrashSpheres.clear();
-
-    m_debugCrashSpheres = false;
-}
-
-void CEngine::AddDisplayCrashSpheres(const std::vector<Math::Sphere>& crashSpheres)
-{
-    for (const auto& crashSphere : crashSpheres)
-    {
-        m_displayCrashSpheres.push_back(crashSphere);
-    }
-
-    m_debugCrashSpheres = true;
-}
-
-
 /*******************************************************
                       Rendering
  *******************************************************/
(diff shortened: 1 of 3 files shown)


src/graphics/engine/engine.cpp
```diff
@@ -3651,76 +3632,138 @@ void CEngine::DrawCaptured3DScene()
...
-    std::vector<VertexCol> lines((2 + LONGITUDE_DIVISIONS + LATITUDE_DIVISIONS) * LINE_SEGMENTS);
-    std::vector<int> firsts(2 + LONGITUDE_DIVISIONS + LATITUDE_DIVISIONS);
-    std::vector<int> counts(2 + LONGITUDE_DIVISIONS + LATITUDE_DIVISIONS);
+    static constexpr int LONGITUDE_DIVISIONS = 16;
+    static constexpr int LATITUDE_DIVISIONS = 8;
+    static constexpr int NUM_LINE_STRIPS = 2 + LONGITUDE_DIVISIONS + LATITUDE_DIVISIONS;
+    static constexpr int VERTS_IN_LINE_STRIP = 32;
 
-    Color color(0.0f, 0.0f, 1.0f);
+    static std::array<Math::Vector, NUM_LINE_STRIPS * VERTS_IN_LINE_STRIP> verticesTemplate = []{
```
krzys-h: Code style: the linter complains that this { should be on a new line

src/graphics/engine/engine.h
```diff
@@ -1406,6 +1406,13 @@ class CEngine : public CSingleton<CEngine>
 
     Texture         m_shadowMap;
 
+    struct
```
krzys-h: The linter complains about the anonymous struct too, although I personally think it's cleaner that way... Maybe just make it `struct PendingDebugDraw {...} m_pendingDebugDraws;`

PR #1155, opened 2018-05-04
Use one static buffer per EngineBaseObject

AbigailBuccaneer:
Each EngineBaseObject contains multiple EngineBaseObjTexTiers, which
each contain multiple EngineBaseObjDataTiers. Each of those DataTiers
currently maintains their own static buffer. This pull request
restructures things so now all of the DataTiers' vertices are in a
single static buffer owned by the EngineBaseObject, and each DataTier
knows its own range of vertices in this buffer.

This decreases CPU-side rendering overhead, as it requires fewer buffer
binds (and in the case of gl14 and gl21, fewer client state changes and
attribute pointer setting).

It also includes various bits of cleanup along the way that were more convenient to do than not to do. 

Here's some performance numbers from my computer: this is the framerate for the entire scene, not just object rendering. It's also not a particularly complicated scene; this was tested on the very first mission from the default camera angle.

As you can see, gl33 doesn't improve much, because most of the work that this cuts out is already saved by the existence of vertex array objects.

Renderer | Framerate before | Framerate after | Frame time before | Frame time after | Improvement
--- | --- | --- | --- | --- | ---
`gl14` | 568fps | 662fps | 1.76ms | 1.51ms | +16.5%
`gl21` | 458fps | 507fps | 2.18ms | 1.97ms | +10.7%
`gl33` | 306fps | 309fps | 3.27ms | 3.24ms | +1.0%


src/graphics/engine/engine.cpp
```diff
@@ -1687,20 +1672,25 @@ void CEngine::UpdateGeometry()
     m_updateGeometry = false;
 }
 
-void CEngine::UpdateStaticBuffer(EngineBaseObjDataTier& p4)
+void CEngine::UpdateStaticBuffer(EngineBaseObject& p1)
```
krzys-h: colobot-lint warning: ```Function 'Gfx::CEngine::UpdateStaticBuffer' has other declaration(s) with inconsistently named parameter(s)```

AbigailBuccaneer: Fixed

Issue #505, opened 2015-07-15
Add missing options from colobot.ini to UI


krzys-h:

RaptorParkowsky:
What options there will be available in 0.1.6?

krzys-h:
* [ ] Setup.FilterMode
* [ ] Setup.MipmapLevel
* [ ] Setup.Anisotropy
* [ ] Setup.MSAA
* [ ] Setup.ShadowMapping
* [ ] Setup.PerPixelLighting (if the new engines will be stable enough)
* [ ] Setup.OffscreenBuffer (automatically adjusts ShadowRange)
* [ ] Setup.JoystickAxisBinding[0-2], Setup.JoystickAxisInvert[0-2], Setup.JoystickDeadzone (see #449)
* [ ] graphics engines

Settings that will be still available only in colobot.ini:
* Setup.ShadowColor (this doesn't really need to be configurable for the user)
* Setup.ShadowRange (but we need to make sure the autodetection works well enough)

RaptorParkowsky:
If the Setup.ShadowRange autodetection will be working not enough good, then we can just add 3 presets or something:
Low=64
High=128
Very Hight=192

krzys-h:
First three options added - 90756f8b344022bd2ff4e2f30820698d60dfa2af

krzys-h:
MSAA finished - e2ebcb769e6dffac4261d89a64d3db6531f929f7

krzys-h:
Shadow mapping settings - f03734d17fc4005acf42c339be5f629f19f4c16e

Erihel:
Some options like FilterMode, MipmapLevel or Anisotropy cause minimap texture to disappear. Minimap is back when you restart level. This happens only if you change settings while in play.

krzys-h:
@erihel Should be fixed now

PR #1168, opened 2018-05-16
Add support for game-specific data directories

krzys-h:
As most of you probably know, we've recieved source code of BuzzingCars and Blupimania2 which were based on the same engine as Colobot and we're slowly working towards bringing them up to date with the Gold Edition engine. The BuzzingCars is already mostly done in the dev-buzzingcars branch of the repo.

This means that soon, we'll have to figure out a way to merge these codebases. And in the future, we may even port levels from old Ceebot games and release them as well.

As most of you probably know as well, the Colobot engine is... not really designed in a way that would allow for easy separation of the games into separate binaries dependent on some common engine library file. This means that for now, there is really no other way than either making a super-binary that can handle all of these games depending on the data dir, or going back to the old ugly #defines scattered across the source code that we've seen in the original ceebot.zip source code package.

My proposal is to, for now, go with the first approach. This is possible with BuzzingCars as the gameplay is mostly the same (some parts of the game are even coded in CBot!) - but we'll definitely have to rethink this when we're finished with Blupimania. The additional huge benefit of that is the possibilities this opens for modders and userlevel makers - I'm imagining levels with BuzzingCars races on Colobot planets and stuff like that. This would mean that we have some data files that are common for the games (models, textures) and some that is separate (levels, help files). This pull request is my attempt at separating these.

From now on, the game looks for two data directories - one is the 'engine common' (`/usr/share/games/colobot/common` or `data/common`) files, and one is the 'game specific' (`/usr/share/games/colobot/<gamename>` or `data/<gamename>`) files, which are overlayed on top of each other exactly like mods normally do. The game directory path is determined based on the executable name - I think this is what some Linux programs do and it would make it easy to handle for maintainers as you just have to add a symlink from gamename to main-engine-binary, but correct me if I'm wrong. This should also translate nicely into package dependencies, where e.g. colobot would depend on colobot-data, colobot-engine-data and colobot-engine-binary.

I would like to get some input from people on this whole thing, especially the package maintainers (ping @OdyX @erihel). We're also open to any suggestions on better ways to handle this.

The colobot-data part of this pull request is at <fill this in when i send the other pull request>

PS. The exact path scheme is still subject to change, I don't like the repeating in `/usr/share/games/colobot/colobot` but I have no better ideas right now


tomangelo2:
There was discussions about engine name few years ago, but we couldn't made a name that isn't already used somehow (to avoid trademark issues), yet still be Colobot-ish enough, so we dropped that topic. Looks like we now need to back in time, it might be indeed confusing if someone will say that `Colobot just got an update`, when game and engine could be separate enough to be developed and updated independently.

Erihel:
Correct me if I'm wrong but animations, collisions and sound effects (like what to play when moving parts of robot arm) are hard coded currently? I think @piotrdz was working on new model format but 
that was long ago. I guess game mods are quite limited right now.

When I was adding PhysicsFS to the game code I didn't want to limit user to *data dir* + *mod dir* but to have a list of modules to load. Colobot data would be a **core** module and user could load how many other modules he wants. A module could add new files or "replace" a file from **core** module.

tomangelo2:
Yes, they're still hard coded into engine. We either need to finish new model format, or move that code to external scripts.

krzys-h:
@Erihel This is almost exactly how the modding support works currently. You have the main data files which are the core module (now split into two separate ones, one loaded always and the other one dependent on the game you launch) and the mods are additional modules on top of that (can be more than one). It's just that the main data dir is hardcoded so that it's required to be valid before the game attempts to launch.

If I remember correctly, the new model format can already load static object along with all the collisions, collision sounds etc. It also has multiple mesh support for purposes of animations in code, but the model format itself has no animation support. Fully moving all the object behavior would probably mean we have to reimplement the objects in CBot, which isn't feasible right now because of how badly slow and unoptimized the CBot interpreter is. The old models are not converted yet because I wanted to wait with that until the other games are ready, because it's easier to port old code before we do a major refactoring like that.

Erihel:
Oh I didn't know that. Maybe you should extend list in #505 to add mod manager in game or in some launcher that was being made? Maybe there can be a roadmap made out of this PR? Current milestones are quite old and misleading like CEGUI porting saying it's 42% complete when there's no CEGUI code (there's a branch but it's 4 years old).

Issue #1124, opened 2018-03-06
Translate to Brazilian Portuguese


badernageral:
Hi, i'm gonna to translate the game to Brazilian Portuguese, i've created the file pt_br.po and started translating it, what's the other files i'm need to translate?

DavivaD:
All other translation files are in colobot-data repo.

badernageral:
I already cloned the colobot-data repository, only the .po extensions files must be translate, or exists another files?

Erihel:
There are multiple po files. You can check pull request for Czech translation https://github.com/colobot/colobot-data/pull/34/files There are some other small changes needed to add support for new language. You can see them in https://github.com/colobot/colobot/pull/1118/files

badernageral:
Ok, thanks.

B-CE:
[Here are the syntax helper & location explanation](https://github.com/colobot/colobot-data/blob/master/README.i18n.md)

badernageral:
Hi, i'm having a problem, I have created the pt_br.po file and added/changed the other files to add language support, but after compile, open the game and change the language in options, the renderized language is english, not my translation.
Currently i'm translating the colobot-data files, but i need to test my translations.
In this [commit](https://github.com/badernageral/colobot/commit/f24fc20d107c2c69629733337da5b839743a2939) are all the modifications i've make in my forked colobot repository, someone please give-me a help!

B-CE:
Can you check the traces of CApplication::SetLanguage  function ?

Launch with:
`colobot -loglevel trace
`

badernageral:
Hi, the output is:
`SetLanguage: Set LANGUAGE=pt_BR.utf8 in environment
Default system locale: pt_BR.UTF-8
Setting locale: LC_CTYPE=pt_BR.utf8;LC_NUMERIC=C;LC_TIME=pt_BR.utf8;LC_COLLATE=pt_BR.utf8;LC_MONETARY=pt_BR.utf8;LC_MESSAGES=pt_BR.utf8;LC_PAPER=pt_BR.utf8;LC_NAME=pt_BR.utf8;LC_ADDRESS=pt_BR.utf8;LC_TELEPHONE=pt_BR.utf8;LC_MEASUREMENT=pt_BR.utf8;LC_IDENTIFICATION=pt_BR.utf8
`
I discover that if I rename the folder in gameFolder/share/locale/pt_br generated by the **make** command to gameFolder/share/locale/pt, the translation works!
[...]

PR #1172, opened 2018-05-28
Expand language tags

tomangelo2:
Now language might be tagged with more chars, eliminating characters collision like in #1124

### DIFF
--- a/src/level/parser/parser.cpp
+++ b/src/level/parser/parser.cpp
@@ -160,7 +160,7 @@ void CLevelParser::Load()
     if (!file.is_open())
         throw CLevelParserException("Failed to open file: " + m_filename);
 
-    char lang = CApplication::GetInstancePointer()->GetLanguageChar();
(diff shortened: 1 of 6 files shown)


src/level/parser/parser.cpp
```diff
@@ -196,17 +196,27 @@ void CLevelParser::Load()
         auto parserLine = MakeUnique<CLevelParserLine>(lineNumber, command);
         parserLine->SetLevel(this);
 
-        if (command.length() > 2 && command[command.length() - 2] == '.')
+        std::size_t separator = command.find('.');
+        if (command.length() > 3 && separator != std::string::npos)
```
krzys-h: Why `command.length() > 3` needs to be here?

tomangelo2:
I'm aware that scripts also needs to be changed, I'm working on that.
I'll fix these after I'll finish data changes.

tomangelo2:
Ok, I'm done with these translations scripts in data. Not gonna waste my time for this anymore.

PR #1174, opened 2018-06-02
Add support for VS CMake/vcpkg compilation

MrSimbax:
Needed a few changes for Colobot to compile in VS 2017.

vcpkg is now probably the easiest way to set up Colobot dev environment on Windows. I'll add the details on the forum, I just need someone to check if my changes haven't broken anything on other systems.

With this commit I managed to compile 64-bit both release/debug and static/dynamic configurations, although the debug one fails at runtime because of [this](https://developercommunity.visualstudio.com/content/problem/256866/assert-in-mbrtowc-when-initializing-stdlocale.html), but that's another issue which will probably be fixed in future versions of VS. The release static build works perfectly.


src/CMakeLists.txt
```diff
@@ -44,8 +44,51 @@ if(MXE) # MXE requires special treatment
 elseif(PLATFORM_WINDOWS)
     # because it isn't included in standard linking libraries
     if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
-        find_library(LIBINTL_LIBRARY NAMES intl.lib)
-        set(PLATFORM_LIBS ${LIBINTL_LIBRARY})
+        find_library(LIBINTL_LIBRARY NAMES intl.lib libintl)
+
+        if(${MSVC_STATIC})
+            if (${OPENAL_SOUND})
+            find_library(FLAC_LIBRARY NAMES flac.lib)
```
krzys-h: Something seems wrong with the indentation here

src/common/system/system_windows.cpp
```diff
@@ -119,7 +119,7 @@ std::string CSystemUtilsWindows::GetSaveDir()
     if (envUSERPROFILE == nullptr)
     {
         GetLogger()->Warn("Unable to find directory for saves - using current directory");
-        savegameDir = CSystemUtils::GetSaveDir(false);
+        savegameDir = CSystemUtils::GetSaveDir();
```
krzys-h: Whoops, my bad, I changed my mind about this code so many times that I left this here accidentaly :P This should be `"./saves"` to keep it the same as in other system_* files

PR #1181, opened 2018-07-05
Add VSync switch to Device tab

tomangelo2:


src/app/app.cpp
```diff
@@ -817,9 +817,15 @@ bool CApplication::CreateVideoSurface()
     int vsync = 0;
     if (GetConfigFile().GetIntProperty("Experimental", "VSync", vsync))
     {
-        SDL_GL_SetSwapInterval(vsync);
+        if (SDL_GL_SetSwapInterval(vsync) == -1)
+        {
+            GetLogger()->Warn("Adaptive sync not supported.\n");
```
krzys-h: This could also be called when setting something else than adaptive sync and it's not supported

src/common/settings.cpp
```diff
@@ -112,6 +112,7 @@ void CSettings::SaveSettings()
 
     // Experimental settings
     GetConfigFile().SetBoolProperty("Experimental", "TerrainShadows", engine->GetTerrainShadows());
+    GetConfigFile().SetIntProperty("Experimental", "VSync", engine->GetVSync());
```
krzys-h: If this gets a aGUI setting now, it should be moved out of Experimental

B-CE:
Just for trace, i'd improoved VSync integration during [this commit](https://github.com/colobot/colobot/pull/1134/commits/1bf0759bfb82369e637f23e2c321e19bfc82dd37).

PR #1189, opened 2018-07-25
Add some Visual Studio specific fixes

MrSimbax:
Changes are explained in commit messages.


src/graphics/engine/engine.cpp
```diff
@@ -431,7 +431,7 @@ bool CEngine::ProcessEvent(const Event &event)
     {
         auto data = event.GetData<KeyEventData>();
 
-        if (data->key == KEY(F12))
+        if (data->key == KEY(F11))
```
krzys-h: I would make this handle both cases, so that people who are used to pressing F12 can keep doing it

src/app/app.cpp
```diff
@@ -1832,7 +1832,12 @@ void CApplication::SetLanguage(Language language)
     // Update C++ locale
     try
     {
+#if defined(_MSC_VER) && defined(_DEBUG)
+        // Avoids failed assertion in VS debugger
+        throw -1;
```
krzys-h: Did you check that translations work correctly with this code?

MrSimbax: No, I did not :(

src/graphics/engine/engine.cpp
MrSimbax: You mean using #ifdef? I'm not sure if keys should change depending on the build, it would seem confusing to most people in my opinion.

krzys-h: No, I mean just `data->key == KEY(F11) || data->key == KEY(F12)` 😛

src/app/app.cpp
MrSimbax: Nothing seems broken, the game starts in Polish on my system. And even if translations were broken, I don't think it's really important for debugging.

PR #1191, opened 2018-08-02
Implement Mod Manager

DavivaD:
Small changes:
1. Moved Mod Manager to Options Menu.
2. Removed Global Mods submenu (because it requires Administrator Privileges).


src/app/app.cpp
```diff
@@ -699,6 +699,26 @@ bool CApplication::Create()
     return true;
 }
 
+void CApplication::Reload()
+{
+    m_sound->Create();
```
krzys-h: You are calling Create() on reload, this function is supposed to be only called once. You are probably reallocating resources without freeing old ones.

src/app/app.cpp
```diff
@@ -699,6 +699,26 @@ bool CApplication::Create()
...
+    m_engine->ReloadAllTextures();
+    CThread musicLoadThread([this]()
```
krzys-h: You just copy-pasted like 15 lines of code from CApplication::Create(), how about replacing it with a call to this function? Also, I'd rename it to `ReloadResources()` to avoid confusion, this is not reloading the whole application

src/app/pathman.cpp
```diff
@@ -72,7 +72,28 @@ void CPathManager::AddModAutoloadDir(const std::string &modAutoloadDirPath)
 
 void CPathManager::AddMod(const std::string &modPath)
 {
-    m_mods.push_back(modPath);
+    std::string::size_type ON;
```
krzys-h: I don't need to wait for Jenkins to finish running the linter to tell you that it's not going to like naming a variable with uppercase letters :P

src/graphics/engine/text.cpp
```diff
@@ -261,6 +261,8 @@ void CText::FlushCache()
     m_lastCachedFont = nullptr;
     m_lastFontType = FONT_COMMON;
     m_lastFontSize = 0;
+
+    Create();
```
krzys-h: Again, you are calling Create() multiple times - this is bad approach and most likely does not free all resources of the previous instance

src/app/pathman.cpp
```diff
@@ -150,15 +171,33 @@ void CPathManager::InitPaths()
...
+                GetLogger()->Info("Found excluded mod: '%s'\n", modPath.c_str());
+            }
         }
     }
 
     for (const std::string& modPath : m_mods)
     {
-        GetLogger()->Info("Loading mod: '%s'\n", modPath.c_str());
-        CResourceManager::AddLocation(modPath);
+      std::string::size_type ON;
```
krzys-h: Something seems wrong with indentation here

src/ui/screen/screen_setup_mods.cpp
```diff
@@ -0,0 +1,281 @@
...
+            pl->SetSelect(-1);
+            pb->ClearState(STATE_ENABLE);
+
+            pb = static_cast<CButton*>(pw->SearchControl(EVENT_INTERFACE_UNLOAD));
+            if ( pb == nullptr )  break;
+            pb->SetState(STATE_ENABLE);
+            break;
+
+        case EVENT_INTERFACE_WORKSHOP:
+            #if defined(PLATFORM_WINDOWS)
```
krzys-h: Files other than system_*.cpp should not contain any platfrom-dependent #ifs, please move this to a function in the system module

src/ui/screen/screen_setup_mods.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+    if (event.type == EVENT_INTERFACE_LOAD)
+    {
+        pl = static_cast<CList*>(pw->SearchControl(EVENT_INTERFACE_MODS_UNLOADED));
+        if ( pl == nullptr )  return false;
+        modName = pl->GetItemName(pl->GetSelect());
+
+        modPathRaw = CResourceManager::GetSaveLocation() + "/" + "mods" + "/";
+        modPath = modPathRaw.c_str();
+        boost::filesystem::rename(modPath+OFF+modName, modPath+modName);
```
krzys-h: UI classes should not contain any logic other than the display code. Move all of the enable/disable/mod listing/check if enabled logic to a new CModManager class.

src/ui/screen/screen_setup.cpp
```diff
@@ -221,6 +239,9 @@ bool CScreenSetup::EventProcess(const Event &event)
                 m_main->ChangePhase(PHASE_SETUPss);
                 return false;
 
+            case EVENT_INTERFACE_SETUPm:
```
krzys-h: This shouldn't ever happen. Add `assert(false)` to make it more explicit?

src/common/restext.cpp
```diff
@@ -89,6 +89,9 @@ void InitializeRestext()
     stringsText[RT_SETUP_KEY1]       = TR("1) First click on the key you want to redefine.");
     stringsText[RT_SETUP_KEY2]       = TR("2) Then press the key you want to use instead.");
 
+    stringsText[RT_MODS_UNLOADED]     = TR("Unloaded Mods:");
```
krzys-h: After modifying restext.cpp you should always run `make update-pot` to update translation files (I keep promising myself that I'll add a comment describing that but I always forget)

krzys-h:
And for the future: modify pull requests instead of closing and reopening them, this way we won't lose discussion and commit history
#1005

src/common/restext.cpp
DavivaD: Oh... Okay...

src/ui/screen/screen_setup.cpp
krzys-h: (or maybe I'm stupid and it should - either way, add a comment why is it different than the others)

tomangelo2:
It seems it doesn't like mods loaded in /data/mods/ directory. It won't load them even though they're present in "loaded" list, and an attempt to disable them cause game to crash due to being unable to find it in /saves/mods/. Due to changing name of directory, they're impossible to be disabled anyway (might be handful in some cases).

DavivaD:
Global Mods (`/data/mods`) will be hidden in mod manager because of root-protected write but I don't know how to do it.

MrSimbax:
The work on this is continued on the [`dev-mod-manager`](https://github.com/colobot/colobot/tree/dev-mod-manager) branch, so I'm closing this PR.

PR #1192, opened 2018-08-04
Fix GCC 8 warning about sprintf overflow

CHmSID:
Fixes a GCC 8 warning about sprintf overflow.

This is my first pull request ever so let me know if you need something done differently.


krzys-h:
This not fixing the core issue, either replace sprintf with snprintf or, even better, StrUtils::Format

krzys-h:
There are apparently more cases of gcc8 errors, would you want to take a look at that? See #1195

Issue #1180, opened 2018-06-28
Build failure on gcc8


amreo:
Hi. I've tryied to setup colobot from AUR (archlinux) but it has failed in building stage. Which maybe is the error?

 `[ 45%] Building CXX object src/CMakeFiles/colobotbase.dir/graphics/model/model_output.cpp.o
/home/amreo/temp/yaourt-tmp-amreo/aur-colobot-gold/src/colobot-colobot-gold-0.1.11.1-alpha/src/graphics/model/model_output.cpp: In function ‘void Gfx::ModelOutput::WriteOldModel(const Gfx::CModel&, std::ostream&)’:
/home/amreo/temp/yaourt-tmp-amreo/aur-colobot-gold/src/colobot-colobot-gold-0.1.11.1-alpha/src/graphics/model/model_output.cpp:331:16: error: ‘char* strncpy(char*, const char*, size_t)’ specified bound 20 equals destination size [-Werror=stringop-truncation]
         strncpy(t.texName, triangle.tex1Name.c_str(), 20);
         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [src/CMakeFiles/colobotbase.dir/build.make:752: src/CMakeFiles/colobotbase.dir/graphics/model/model_output.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:219: src/CMakeFiles/colobotbase.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
`

Erihel:
What compiler do you use? Do you have some compiler flags set in `/etc/makepkg.conf` that treat warnings as errors?

This warning comes from code that's marked as deprecated. Quick workaround would be to change 20 to 19 in `src/graphics/model/model_output.cpp:331`

amreo:
I'm using GCC 8.1.1 on GNU/Linux. These are the default flags present in /etc/makepkg
`#########################################################################
# ARCHITECTURE, COMPILE FLAGS
#########################################################################
#
CARCH="x86_64"
CHOST="x86_64-pc-linux-gnu"

#-- Compiler and Linker Flags
CPPFLAGS="-D_FORTIFY_SOURCE=2"
CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt"
[...]

PR #1199, opened 2018-08-18
Fix another GCC 8.2.0 warning

### LINKED ISSUE #1195: Cannot build from source nor build from AUR
Hello everyone!
I am trying to install the game on my computer. so I tried installing from the AUR and it failed it gave me an error message similar to the same one here https://github.com/colobot/colobot/issues/1180.
After that I have tried to install from source which also failed(both stable branch and dev branch), I even ran cmake with ignore errors flag which did let it compile but the game couldn't launch.

```
/tmp/pamac-build-oren/colobot-gold/src/colobot-colobot-gold-0.1.11.1-alpha/src/graphics/model/model_output.cpp: In function ‘void Gfx::ModelOutput::WriteOldModel(const Gfx::CModel&, std::ostream&)’:
/tmp/pamac-build-oren/colobot-gold/src/colobot-colobot-gold-0.1.11.1-alpha/src/graphics/model/model_output.cpp:331:16: error: ‘char* strncpy(char*, const char*, size_t)’ specified bound 20 equals destination size [-Werror=stringop-truncation]
         strncpy(t.texName, triangle.tex1Name.c_str(), 20);
         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

cc1plus: all warnings being treated as errors
make[2]: *** [src/CMakeFiles/colobotbase.dir/build.make:752: src/CMakeFiles/colobotbase.dir/graphics/model/model_output.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:187: src/CMakeFiles/colobotbase.dir/all] Error 2
```
I also tried to change this part to `strncpy(t.texName, triangle.tex1Name.c_str(), 20);` writing 19 instead of 20 but then I ran to another error message.

I have cmake 3.11.4-2
gcc 8.2.0-2

when compiling the game with errors:
```
Error in CEngine::Init()

Please see the console output or log file
to get more information on the source of error


```
what should I do ?
thanks for the help

CHmSID:
Fixes #1195 and #1180 
Since the code is marked as deprecated I tried to change as little as possible.

What changed:
- Texture name buffers are now size 21 instead of 20 to prevent a possible lack of null character.
- Buffers are now initialized to a null character.
- All `strcpy` behaviour was left unchanged. They still copy 20 characters, therefore the 21st one is guaranteed to be always null.

Are there any tests I can run locally to make sure I didn't break anything (the game itself runs)? It's hard to find information about that.

### DIFF
--- a/src/graphics/model/model_io_structs.h
+++ b/src/graphics/model/model_io_structs.h
@@ -188,7 +188,7 @@ struct OldModelTriangleV1
     Vertex p2;
     Vertex p3;
     Material material;
-    char texName[20] = {};
+    char texName[21] = {'\0'};
     float min = 0;
     float max = 0;
 };
@@ -207,7 +207,7 @@ struct OldModelTriangleV2
     Vertex p2;
     Vertex p3;
     Material material;
-    char texName[20] = {};
+    char texName[21] = {'\0'};
     float min = 0.0f;
     float max = 0.0f;
     long state = 0;
@@ -231,7 +231,7 @@ struct OldModelTriangleV3
     VertexTex2 p2;
     VertexTex2 p3;
     Material   material;
-    char texName[20] = {};
+    char texName[21] = {'\0'};
     float min = 0.0f;
     float max = 0.0f;
     long state = 0;


tomangelo2:
There is a mods folder, you can unpack it inside data directory, it contains parts of Me model with textures with names from 18 to 22 chars. Gloves and thighs should be untextures due to 21 and 22 character long texture names, while arms should be textured as they have 18-20 character long texture names.
[mods.zip](https://github.com/colobot/colobot/files/2300407/mods.zip)

CHmSID:
Thank you for the test case.
I can confirm the gloves and thighs are untextured, arms are different colours, and the console output indicates that the two textures with 21 and 22 characters are blacklisted.

Issue #1204, opened 2018-10-03
Add an appdata XML for Linux packaging 


ananace:
Most of the modern Linux packaging methods and stores expect applications to provide metainfo in the form of AppStream metadata formatted XML.

The format is described at;
https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

Erihel:
This has been merged.

PR #1205, opened 2018-10-03
Add an appdata XML for Linux packaging (#1204)

ananace:
This is just an example for how the required appdata XML could be built, and OARS and additional information might need cleaning up.

The releases can either be handled manually by adding release objects for every release, or just made to track the latest one by generating the file with CMake and using the current date or the tag date if the build is on a release tag.


ananace:
Indeed, I first wrote it to generate into the binary dir, but didn't finish that as I wasn't sure how said generation would be wanted. Must've forgotten to update the source path as well.

Issue #1203, opened 2018-09-16
SIGSEGV on Destroying Power Cell Factory


Willyanto39:
Destroying power cell factory while an object is placed on it crashes the game.

Btw, I am using version 0.1.11-alpha.

tomangelo2:
Confirmed. Placed battery cell in it and then shoot with Phazer.
```gdb
Thread 1 received signal SIGSEGV, Segmentation fault.
0x004a54ba in COldObject::UpdateTransformObject (this=0x1322e590, part=0,
    bForceUpdate=false)
    at C:/msys64/home/user/colobot/src/object/old_object.cpp:1803
1803                matWorldTransporter = m_transporter->GetWorldMatrix(m_transporterLink);
(gdb) bt
#0  0x004a54ba in COldObject::UpdateTransformObject (this=0x1322e590, part=0,
    bForceUpdate=false)
    at C:/msys64/home/user/colobot/src/object/old_object.cpp:1803
#1  0x004a5866 in COldObject::UpdateTransformObject (this=0x1322e590)
    at C:/msys64/home/user/colobot/src/object/old_object.cpp:1855
#2  0x004a6ad0 in COldObject::EventFrame (this=0x1322e590, event=...)
    at C:/msys64/home/user/colobot/src/object/old_object.cpp:2171
#3  0x004a6969 in COldObject::EventProcess (this=0x1322e590, event=...)
    at C:/msys64/home/user/colobot/src/object/old_object.cpp:2141
#4  0x0046baf5 in CRobotMain::EventFrame (this=0xc4e4ba8, event=...)
    at C:/msys64/home/user/colobot/src/level/robotmain.cpp:2369
#5  0x00465885 in CRobotMain::ProcessEvent (this=0xc4e4ba8, event=...)
    at C:/msys64/home/user/colobot/src/level/robotmain.cpp:684
#6  0x004090c8 in CController::ProcessEvent (this=0xc4ddd48, event=...)
    at C:/msys64/home/user/colobot/src/app/controller.cpp:56
#7  0x00406620 in CApplication::Run (this=0x28fc30)
    at C:/msys64/home/user/colobot/src/app/app.cpp:1143
#8  0x00401b5e in SDL_main (argc=1, argv=0x4f7eec8)
    at C:/msys64/home/user/colobot/src/app/main.cpp:197
#9  0x0064c344 in main_getcmdline ()
#10 0x0064c455 in WinMain@16 ()
[...]

Issue #1194, opened 2018-08-11
SIGSEGV, segmentation fault


Defi1X:
Hello, I have this problem with starting Colobot (Compiled from master branch today)

```
leb@quantum:~$ cd /some/install/path/games/
leb@quantum:/some/install/path/games$ ./colobot -datadir /some/install/path/share/games/colobot/
[INFO]: Colobot: Gold Edition 0.1.11.1-alpha starting
[INFO]: Using data dir: '/some/install/path/share/games/colobot/'
[INFO]: Creating CApplication
[INFO]: Data path: /some/install/path/share/games/colobot/
[INFO]: Save path: /home/leb/.local/share/colobot
[ERROR]: Error on parsing config file: failed to open file
[WARN]: Config could not be loaded. Default values will be used!
[INFO]: Opening audio device...
[INFO]: Done.
[INFO]: CApplication created successfully
[INFO]: No joysticks detected
[INFO]: Creating CDevice - OpenGL 1.4
[INFO]: OpenGL 3.0 Mesa 10.1.3
[INFO]: Mesa DRI Intel(R) Cherryview 
[INFO]: Shadow mapping available (core)
[INFO]: Anisotropic filtering available
[INFO]: Maximum anisotropy: 16
[INFO]: Multisampling supported, max samples: 8
[INFO]: Auto-detecting VBO support
[INFO]: Core VBO supported
[INFO]: Maximum texture units: 8
[INFO]: Maximum texture size: 8192
[INFO]: Framebuffer supported (ARB)
[INFO]: Maximum renderbuffer size: 8192
[INFO]: CDevice created successfully
[ERROR]: Error on storing config file: failed to open file
[ERROR]: Error on storing config file: failed to open file
[ERROR]: Error on storing config file: failed to open file
[INFO]: Created shadow map texture: 1024x1024, depth 32
[ERROR]: Framebuffer incomplete: [ERROR]: attachment point incomplete[ERROR]: 
[ERROR]: Could not create MSAA framebuffer, disabling MSAA

Unhandled exception occurred!
==============================
SIGSEGV, segmentation fault
==============================

This is usually caused by a bug. Please report this on http://github.com/colobot/colobot/issues
including information on what you were doing before this happened and all the information below.
==============================
You seem to be running a custom compilation of version 0.1.11.1-alpha, but please verify that.

The game was in phase PHASE_WELCOME1 (ID=0)
Last started level was: category=exercises chap=0 rank=0
==============================

[...]

PR #1206, opened 2018-10-10
Fixes for #1203 and #1194

iceslab:
Set `m_transporter` to `nullptr` before deletion to avoid SIGSEGV.

### DIFF
--- a/src/graphics/engine/engine.cpp
+++ b/src/graphics/engine/engine.cpp
@@ -4048,7 +4048,9 @@ void CEngine::UseMSAA(bool enable)
                 }
             }
 
-            framebuffer->Bind();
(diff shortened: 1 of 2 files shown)


tomaszkax86:
I checked the changes and it mostly works. Mostly because the crash still occurs, but only when I put titanium into power plant and the plant doesn't convert it to power cell, which can only happen if you build one where there's no energy in the ground. I did more code reading and I believe this can be corrected by adding `m_object->SetPower(nullptr);` in `DestroyObject()` in two places right before `CObjectManager::GetInstancePointer()->DeleteObject(cargo);` in autopowerplant.cpp. Could you verify this?

PR #1211, opened 2018-11-18
Add a minimum & maximum game speed limit

suve:
Currently it is possible to use the in-game keybinds to set the game speed to ludicrous values, like x65536. On my machine, the game crashes once x1048576 is reached.

This patch puts a minimum and maximum limit on what speed can the player set the game to.


MrSimbax:
Thanks for contributing!

PR #1239, opened 2019-02-02
Tweak strncpy() usage in ModelOutput::WriteOldModel

suve:
The code in said function uses `strncpy(dest, src, size)`, with the `size` being equal to the destination buffer size. If the source buffer is larger than the dest buffer, the resulting string won't be properly NUL-terminated. This changeset fixes that.

### DIFF
--- a/src/graphics/model/model_output.cpp
+++ b/src/graphics/model/model_output.cpp
@@ -328,7 +328,10 @@ void ModelOutput::WriteOldModel(const CModel& model, std::ostream &stream)
         t.material.ambient = triangle.ambient;
         t.material.diffuse = triangle.diffuse;
         t.material.specular = triangle.specular;
-        strncpy(t.texName, triangle.tex1Name.c_str(), 20);
+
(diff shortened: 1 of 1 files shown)


src/graphics/model/model_output.cpp
```diff
@@ -328,7 +328,10 @@ void ModelOutput::WriteOldModel(const CModel& model, std::ostream &stream)
         t.material.ambient = triangle.ambient;
         t.material.diffuse = triangle.diffuse;
         t.material.specular = triangle.specular;
-        strncpy(t.texName, triangle.tex1Name.c_str(), 20);
+
+        strncpy(t.texName, triangle.tex1Name.c_str(), sizeof(t.texName)-1);
```
tomangelo2: Linter complains here about `Implicit cast bool -> 'char'`.

suve: This is line 332. The warning is for line 322 (10 lines above this one), for this little bit:
```
    t.used = true;
```

tomangelo2: Oh, my bad.

Issue #1246, opened 2019-02-24
Please provide AppImage for download


probonopd:
https://colobot.info/download-colobot-gold/ has an AppImage but it is contained inside an archive.

Please __DO NOT__ put an AppImage into another archive like a `.zip` or `.tar.gz`. While it may be tempting to avoid users having to set permission, this breaks desktop integration with the optional `appimaged` daemon, among other things. Besides, the beauty of the AppImage format is that you never need to unpack anything. Furthermore, packing an AppImage into some form of archive prevents the AppImage from being added to the central catalog of available AppImages at https://github.com/AppImage/AppImageHub.

Reference:
https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages#common-mistake

MrSimbax:
We didn't do that because the AppImage requires `data` and `lang` folders to work. We have yet to figure out how to embed them inside the AppImage. We'll probably update our download page once we do that.

I personally didn't see any problems with putting it inside archive but I'm more used to Windows way of distributing software and treat AppImage more like a static executable than a container. Thanks for admonishing us though.

probonopd:
Are the data and lang folders changing or do they contain static (non-changing) data?

* If the data is not changing, then you can simply put them inside the AppImage.
* If the data is changing, then it should be copied from the AppImage to somewhere into `$HOME` upon the first launch of the application, and then used from there.

MrSimbax:
They contain static data. We tried putting them inside AppImage but Colobot refuses to load them. I don't know the details as I haven't tried it myself so I'll ping @tomangelo2 here.

MrSimbax:
By the way, should AppImage by default create user-specific save files in the directory it's in or in other directory? Currently they reside in `$XDG_DATA_HOME`. Is there a way to create a directory like `Colobot-x86_64.AppImage.data` for `$XDG_DATA_HOME` like it is possible for `$HOME` (`Colobot-x86_64.AppImage.home`)?

probonopd:
`$XDG_DATA_HOME` is correct. No need to change.

tomangelo2:
I've already created an issue about that https://github.com/colobot/colobot/issues/1245.
The problem is that engine does looks for data and lang directories in folder where AppImage is stored, not inside of directories mounted by AppImage. That's why we couldn't make it to work in proper way. I've made some research, turns out SDL2 might come handy here, if not then there is `binreloc` library to help https://github.com/limbahq/binreloc.

[...]

Issue #1245, opened 2019-02-23
Better support for binary relocation


tomangelo2:
From 0.1.12 release we've decided to deliver AppImage for Linux builds, but it's not in proper way this image should be delivered. Instead of one file with everything inside of it, there are `data` and `lang` directories next to it. This is because Colobot - when set to relative paths - is searching for files in directory where AppImage file is placed. Instead it should look inside directories mounted by AppImage.

PR #1247, opened 2019-02-24
Add support for binary relocation

MrSimbax:
The changes are required for AppImage to work (#1246).

The changes could break the standard building process for Linux (the paths hardcoded in binary are now relative) so probably package maintainers should try them before we merge them.

I have no ability to test the OS X build either.

And I can confirm that the changes work since I created AppImage for release 0.1.12-alpha with them patched. AppImage from Jenkins (DEV_BUILD=1) does not have embedded data files and still uses work directory which should be more comfortable for testers and developers in my opinion.

Haven't yet did any other tests, so building process for Windows should also be tried.

**tl;dr:** I haven't tested it except on AppImage. Please try it so we know that it can be safely merged.

Related issue: #1245


krzys-h:
I think it would be better to keep the old behavior as a fallback and for backwards compatibility. That is e.g. how nmap does things. And while we're at it, most software also allows you to override the data dir path using environment variables, we could add that as well.

Using local dir rather than base dir in dev builds makes no sense at all... 

Also, how about generating the relative paths in cmake rather than hardcoding them? That would allow for easier customization by just changing the INSTALL path without having to remember about the relative variants. In that case, maybe USE_RELATIVE_PATHS (defaulting to true) rather than leaving the old behavior would be a better idea. That would also probably deprecate PORTABLE since it would be equivalent to customizing the install paths with relative paths enabled (which is the only reason you couldn't do that earlier).

MrSimbax:
> And while we're at it, most software also allows you to override the data dir path using environment variables, we could add that as well.

Could give an example how should it work in Colobot?

> Using local dir rather than base dir in dev builds makes no sense at all... the point was to use something else than system-wide install directory, with relative paths there is no point of doing that.

I didn't know what was the original purpose of this flag. I agree it's ugly workaround though, I used it only to keep the debug AppImage in Jenkins working.

So let me see if I understand correctly since I have some doubts. Working directory only matters to the `saves` folder, nothing else. Assume `CMAKE_INSTALL_PREFIX=install` for simplicity.

When `USE_RELATIVE_PATHS=1` then CMake generates relative paths automatically (somehow) for `data` and `lang`. CMake of course installs `data` to `install/data` or `install/share/games/colobot` depending on platform and flags. But the binary has got hard coded the **generated** relative paths: `data` or `../share/games/colobot`. It uses basedir to determine the real paths at runtime.

Assume `USE_RELATIVE_PATHS=0` then we have old behaviour and the binary has got hard coded paths `absolutePath(install)/data` or `absolutePath(install)/share/games/colobot`.

So, you still need the `PORTABLE` flag to distinguish between `install/data` and `install/share/games/colobot`. Unless you want people to manually specify all the `COLOBOT_INSTALL_*` variables. Or Windows has relative paths anyway to the workdir and we assume that no one ever will want the `PORTABLE` behaviour on Linux. Maybe then this flag could be removed but it would only make things harder if for example I actually wanted the `PORTABLE` behaviour on Linux with paths other than the share slash games slash something thing, like when I'm working locally and want to create either the AppImage or have it in my `/usr/local`. I would have to specify all the `COLOBOT_INSTALL_*` variables in the command line or modify `CMakeLists.txt`.

> USE_RELATIVE_PATHS (defaulting to true) rather than leaving the old behavior

I don't understand this part - how would it remove the old behaviour? Isn't the whole point of this variable to have both behaviours?
And if we want to keep backwards compatibility, so I don't know, package maintainers don't have to update their scripts, then the USE_RELATIVE_PATHS defaulting to true is opposite to the compatibility since it introduces new behaviour by default which *could* potentially break something (although probably nothing will break since the final paths at run-time will probably be the same as always in this case but that's not the point).

I want to make sure we have the same idea before I start working on it.

krzys-h:
Assume \<basedir\> means the directory where the main application binary is placed

---

> > And while we're at it, most software also allows you to override the data dir path using environment variables, we could add that as well.
>
> Could you give an example how should it work in Colobot?

The search path would be as follows:
1. If -datadir is specified on the commandline, use that
2. If not, check if the COLOBOT_DATA_DIR environment variable exists. If it does, use that.
3. Finally, use the default of ../share/colobot or whatever

This lets you change the data dir by modifying your environment variables without having to specify explicit -datadir every time (and without recompilation of course), something like this is supported in basically every Linux program that needs addditional data files that I looked at.

>> Using local dir rather than base dir in dev builds makes no sense at all... the point was to use something else than system-wide install directory, with relative paths there is no point of doing that.
>
> I didn't know what was the original purpose of this flag. I agree it's ugly workaround though, I used it only to keep the debug AppImage in Jenkins working.

The point of `./data` was just to have a relative path rather than an absolute one, and I just never thought about making it `<basedir>/data` before. There is no point in keeping it relative to the work dir now that we have a better solution.

Ah... I think I get it now. For appimage, `<basedir>` would be inside the image and `.` is outside, right? That breaks the whole idea of having binary and data as separate packages. The perfect thing would be to somehow make it appimage-file-relative for Jenkins builds so that `./some/random/path/colobot.AppImage` works, but I'm not sure anything like that exists. In ether case, we may need yet another flag to switch this behaviour - not DEV_BUILD though, you don't want your local dev builds randomly switching to using workdir rather than basedir.

> Working directory only matters to the `saves` folder, nothing else.

... in PORTABLE_SAVES builds, I assume?

No. I didn't know what I was doing when I initially implemented it / was too lazy to figure out how to find the basedir. Typing `colobot` in the terminal shouldn't start dropping random things in whatever directory I happen to be in (assuming you have it on PATH). If colobot is installed system-wide then it should use the system user directory (`~/whatever`), if this is a fully portable build, it should be basedir-relative `<basedir>/saves`. I'm not sure what the convention is for appimage, but my guess would be the user directory - that is how every single precompiled Linux binary I ever used worked (which is probably why I never complained about our portable Windows builds doing that).

---

> When `USE_RELATIVE_PATHS=1` then CMake generates relative paths automatically (somehow) for `data` and `lang`. CMake of course installs `data` to `install/data` or `install/share/games/colobot` depending on platform and flags. But the binary has got hard coded the **generated** relative paths: `data` or `../share/games/colobot`. It uses basedir to determine the real paths at runtime.

Exactly. For standard Linux setup, you have:
INSTALL_PREFIX: /usr/local
BINARY_INSTALL_DIR: ${INSTALL_PREFIX}/games/
LIB_INSTALL_DIR ${INSTALL_PREFIX}/lib/colobot/
DATA_INSTALL_DIR: ${INSTALL_PREFIX}/share/games/colobot/
LANG_INSTALL_DIR: ${INSTALL_PREFIX}/share/locale/
but CMake transforms these to the following `#define`s in the binary (basedir-relative, NOT workdir-relative):
DEFAULT_DATA_DIR ../share/games/colobot
DEFAULT_LANG_DIR ../share/locale

(Also, I just noticed: this setup is still not really relocatable since on Linux libraries often use hardcoded paths so it will always load libCBot.so from /usr/local/lib/colobot/libCBot.so, but I guess appimage deals with that already.)

For Windows, you have
INSTALL_PREFIX: ./some_output_dir
BINARY_INSTALL_DIR: ${INSTALL_PREFIX}/
LIB_INSTALL_DIR ${INSTALL_PREFIX}/
DATA_INSTALL_DIR: ${INSTALL_PREFIX}/data/
LANG_INSTALL_DIR: ${INSTALL_PREFIX}/lang/
which would transform into
DEFAULT_DATA_DIR ./data
DEFAULT_LANG_DIR ./lang

> Assume `USE_RELATIVE_PATHS=0` then we have old behaviour and the binary has got hard coded paths `absolutePath(install)/data` or `absolutePath(install)/share/games/colobot`.

Exactly. This is probably the right setup to use when building non-appimage Linux builds, for OS packages etc. - but I guess we should ask package maintainers about that.

That makes me doubt what the default should be... On Linux, the default should probably be USE_RELATIVE_PATHS=0 unless building appimage or portable build, on Windows the default should always be USE_RELATIVE_PATHS=1 because this is what is usually expected of Windows programs.

> So, you still need the `PORTABLE` flag to distinguish between `install/data` and `install/share/games/colobot`. Unless you want people to manually specify all the `COLOBOT_INSTALL_*` variables. Or Windows has relative paths anyway to the workdir and we assume that no one ever will want the `PORTABLE` behaviour on Linux. Maybe then this flag could be removed but it would only make things harder if for example I actually wanted the `PORTABLE` behaviour on Linux with paths other than the share slash games slash something thing, like when I'm working locally and want to create either the AppImage or have it in my `/usr/local`. I would have to specify all the `COLOBOT_INSTALL_*` variables in the command line or modify `CMakeLists.txt`.

That is basically what I was trying to say. We don't *need* PORTABLE anymore, but I guess it can be used as a shortcut. It just doesn't do anything that can't be achieved by modifying other flags now.

> > USE_RELATIVE_PATHS (defaulting to true) rather than leaving the old behavior
>
> I don't understand this part - how would it remove the old behaviour? Isn't the whole point of this variable to have both behaviours?

I was just referring to the first part of my comment where I suggested having both of them and using the hardcoded one if we can't find the files in the relative one. Now we only have one or the other, chosen at compile time.

MrSimbax:
> Ah... I think I get it now. For appimage, `<basedir>` would be inside the image and `.` is outside, right? That breaks the whole idea of having binary and data as separate packages. The perfect thing would be to somehow make it appimage-file-relative for Jenkins builds so that `./some/random/path/colobot.AppImage` works, but I'm not sure anything like that exists. In ether case, we may need yet another flag to switch this behaviour - not DEV_BUILD though, you don't want your local dev builds randomly switching to using workdir rather than basedir.

Yes, when you run AppImage the contents is mounted in some directory in `/tmp` and the binary is run from there. If you look at the logs when you run Colobot AppImage you can clearly see from where the `data` content is loaded from. That was the whole point of issue #1246 and it's the main purpose of this pull request to make it work.

Of course it breaks the idea of separate packages, AppImage is supposed to be the whole package with all the static data it needs to work. I'm not sure what the flag should be then though, `LOAD_DATA_FROM_WORKING_DIR`? `DEBUG_APPIMAGE_BUILD`? `USE_WORKING_DIR_PATHS`?

>For Windows, you have
INSTALL_PREFIX: ./some_output_dir
BINARY_INSTALL_DIR: ${INSTALL_PREFIX}/
LIB_INSTALL_DIR ${INSTALL_PREFIX}/
DATA_INSTALL_DIR: ${INSTALL_PREFIX}/data/
LANG_INSTALL_DIR: ${INSTALL_PREFIX}/lang/
which would transform into
DEFAULT_DATA_DIR ./data
DEFAULT_LANG_DIR ./lang

Should it be `./data` or `data`? `<basedir>./data` looks weird.

MrSimbax:
I think I misunderstood the whole purpose of AppImage and we shouldn't use it for debug builds for Jenkins after all. We should just compile it statically in Jenkins with portable paths. I am not sure how to do that though, I have never managed to compile anything statically in Linux. So I think that AppImage from Jenkins should be treated as temporary solution (it just simulates static linking by putting all the dynamically linked variables inside AppImage). It should probably have portable paths.

krzys-h:
> Of course it breaks the idea of separate packages, AppImage is supposed to be the whole package with all the static data it needs to work. I'm not sure what the flag should be then though, `LOAD_DATA_FROM_WORKING_DIR`? `DEBUG_APPIMAGE_BUILD`? `USE_WORKING_DIR_PATHS`?

I had no idea for a name either, that''s why I didn't mention any.

>>For Windows, you have
INSTALL_PREFIX: ./some_output_dir
BINARY_INSTALL_DIR: ${INSTALL_PREFIX}/
LIB_INSTALL_DIR ${INSTALL_PREFIX}/
DATA_INSTALL_DIR: ${INSTALL_PREFIX}/data/
LANG_INSTALL_DIR: ${INSTALL_PREFIX}/lang/
which would transform into
DEFAULT_DATA_DIR ./data
DEFAULT_LANG_DIR ./lang
>
>Should it be `./data` or `data`? `<basedir>./data` looks weird.

Doesn't matter really, I just added it for clarity - it resolves to the same path

> We should just compile it statically in Jenkins with portable paths. I am not sure how to do that though, I have never managed to compile anything statically in Linux.

I would have done that a long time ago if it was that easy :P You need a version of each library that was made for static linking (.a rather than .so) and these are not provided anywhere with the standard packages afaik, so you would need to build all libraries from source with proper flags to generate these. And you would also need to build all their dependencies. And the dependencies of their dependencies. Ugh. And the binary after that would probably be >100 MB big like the MXE one.

tomangelo2:
`and these are not provided anywhere with the standard packages afaik`
https://askubuntu.com/questions/1052434/is-it-possible-to-install-static-libraries-from-apt-in-18-04-x86-64
Actually in `apt` based distributions there are `.a` libraries in `/usr/lib/x86_64-linux-gnu/` directory. It just requires to override `-lsomelib` flag with `-l:somelib.a` for every library used in project. Also there are already few flags in CMake to link some libs statically, but I've never checked them if they works. There is also `-static` flag.

krzys-h:
I don't remember what the exact problem was then, maybe it changed or it was just a few dependencies missing. I remember just that it would involve more manual compilation than I was willing to do at the time.

MrSimbax:
This pull request is ready for code review and testing.

Changes:

- Saves are now loaded (as fallback or with `PORTABLE_SAVES` flag) from the base directory instead of the working directory.
- Added a new option to CMakeLists: `USE_RELATIVE_PATHS`.
    - It's turned off by default for backwards compatibility on Linux. On Windows it should be on by default, but I don't know how to do that, so it must be enabled by hand (yeah, another flag to remember for Windows builds).
    - It's always turned on with the `PORTABLE` flag, regardless of the operating system. Note that for Windows the `PORTABLE` flag doesn't really change anything.
    - It only matters for `data` and `lang` directories.
    - If it's off, then the **absolute** paths to `data` and `lang` are hardcoded into the binary. This is meant for Linux non-portable installations.
    - If it's on, then the **relative to the binary** paths to `data` and `lang` are hardcoded into the binary. This is meant for AppImage, Windows and portable installations. **Note:** This flag must be specified by hand for Windows, I don't know how to change the default based on the platform used.
    - The relative paths are generated from the `COLOBOT_INSTALL_*` variables automatically.
- Generally, all the working directory paths in the game are now binary relative. At least the ones considered in the pull request.

The changes are not supported for OS X as I don't have it, so I can't test them. Also I have no idea how things should work on OS X so I'd rather not blindly touch it.

Another note: since lib paths in Linux binaries are apparently hardcoded, then, if my understanding is correct, this means that in order to make the Linux not-AppImage build truly portable, `CBOT_STATIC` should also be on.

Issue #1008, opened 2017-07-24
switch..case not accept negative values


Arano-kai:
Ver: 0.1.10 (from sid)
OS: Debian (stretch)
#920 Fixed default args but not `switch..case` statement.
```
extern void object::TestSwitch()
{
	test(-2);
	test(-1,-1);
	test(-1);
}

void test(int a, int b = -2) {
	message("a="+a+" b="+b);
	switch(a){
		case b: message("a=b"); break; // completely ignores, see #998
		//case -1: message("a=-1"); break; // can't find ":" <== fixed by #920... or not (:
		case 0: message("a=0"); break;
		case 1: message("a=1"); break;
		default: message("woops!");
	}
}
```

Arano-kai:
Please note, that is current Linux release is from sid (unstable), so need confirmation on other platforms.

tomaszkax86:
I can confirm that there's a problem on Windows as well. Negative numbers cause compilation error and none of those tests worked right.

However, `switch...case` in most languages doesn't normally accept variables for `case` clauses, only constants, so `case b:` is probably incorrect on the language level. I think we could keep this, but we need to fix the bug first.

melex750:
This compiles with no error:(
```c++
extern void CaseBug()
{
	switch(-1)
	{
		case var   :
		case case  :
		case while :
		case "str" :
		case ?     :
		case int   :
		case point :
		case &&    :
		default:
	}
}
```
The fact that you can use a variable name with```case```is a bug,
in such cases, the equivalent of```case 0:```is compiled.

It would be nice if unary operations **```+  -  ~```**, and other
arithmetic and bitwise operations, could be used with```case```.

I think only numerical constants should be supported, including character constants and enumerators.

Arano-kai:
Then we has to have the way to define const and enum in scripts.

Issue #993, opened 2017-07-19
Math near border values return wrong results.


Arano-kai:
Ver: 0.1.9
OS: Debian (stretch)
Issue:
```
i = abs(-2147483606); // -2147483648
i = -2147483600 * 1; // -2147483648
i = 2147483600 * 1; // -2147483648
i = 2147483646 * 1; // -2147483648
i = 2147483646 * -1; // -2147483648
i = 2147483000 * -1; // -2147483008
i = 2147483000 * 1; // 2147483008
```

tomaszkax86:
Looks like these operations are performed on `float` rather than `int`, hence it loses precision after 7 decimal places. This definitely needs to be checked.

MrSimbax:
That's right, I tested it once. http://colobot.info/SatComOnline/master/cbot/expr

MrSimbax:
Looks like all of those examples are fixed except
```
i = abs(-2147483606); // -2147483648
```

melex750:
Oops, I forgot about that, the other math functions also need attention. I'll have a look this evening.

hexagonrecursion:
This issue has been fixed and should be closed.

PR #1253, opened 2019-04-11
Add types byte, short, char, long, and double

melex750:
**Refactor SaveState and RestoreState to use streams**
* Added ability to **test save/restore** in CBotUT.
* Improved handling errors when loading a saved game.
* More compact and flexible file format.

Fixed bug:
```c++
extern void object::Array_Of_Class_Bug()
{
    object obj[] = {};   <----- declare array of object (or any other class)

    message("waiting"); wait(10); <------ save/reload here

    obj[0] = this; <----------run-time error "Wrong type for the assignment"
    obj[1] = radar(Any); <------------/
}
```

**Fix arithmetic operations with integers (#993)**
* Also fixed unsigned right shift operator ">>>"

**Add missing primitive data types in CBOT**
* char is 32-bit unsigned integer for UTF-32 values
* Assigning or adding char to string converts to UTF-8

**Add sizeof() operator for numeric data types**
* E.g. ```sizeof(int)```, ```sizeof(long)```, etc.

**Add binary and character literals in CBOT**

**Fix and improve switch...case (#1008)**
* Case labels can use unary minus and other arithmetic operations.
* Added compile error for duplicate case/default labels.
* Default label can be anywhere, not only the bottom.

**TODO**
* More doxygen documentation for changes in this PR.
* ? Do we need to throw a compile error if we detect a bad right
  operand for bit shift? E.g. negative or number too large.

**New issues for the bug tracker ?**
* Unit tests and SatCom docs for new data types.
* More unit tests and SatCom docs for bitwise operations.


tomangelo2:
I've got some time to check that pull request. Didn't found anything that would break after merging this code. I'm not experienced enough to do proper code review, but I don't see anything bad here.

hexagonrecursion:
> Add missing primitive data types in CBOT

1. I would like to know why we need byte,and short. Just because C++ has a real need for those datatypes (frugal use of memory, cpu cache efficiency, ram bandwidth efficiency) does not make them "missing". In CBOT they appear to be "just like int, but worse". What is their use-case in CBOT?
2. I do not see why we need both float and double in CBOT. If the issue is that float is too small for some programs we can just make CBOT float be C++ double. This is what the [javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_encoding), [python](https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex) and [lua (by default)](https://www.lua.org/manual/5.3/manual.html#2.1) do.
3. Likewise if the only reason to have long in CBOT is that int is only 4 bytes on some platforms (technically can be as small as 2), we can just have CBOT int be C++ long long (guaranteed to be at least 8 bytes)
 
> Add sizeof() operator for numeric data types

4. Why? In C we need sizeof() to calculate malloc() sizes. What is the use-case in CBOT?

melex750:
These new types were not actually "missing", just a tiny bit of code that enabled them. sizeof() is obviously useful if someone can use bitwise operators.

tomangelo2:
1. Educational purpose. CBot is a language with syntax similar to C++/Java. This design allows players to smooth transition from one language to another. Existence of C++/Java types in CBot smooths down transition further by making them more alike. Also it makes it possible for new players to learn in-game the difference between them.
2. File I/O operations. To properly load some files you might need a wider palette of types. Particularly when dealing with files in binary mode instead of text mode. You can do it already with `read_byte()` and `write_byte()`, but it's harder to use.

hexagonrecursion:
I do not understand the File I/O argument
1. CBOT is not a general-purpose programming language. It is a scripting language used by one game. How does being able to write a .zip parser help you complete a level?
2. The types are not very useful for advanced binary I/O because:
    1. Their sizes are not guaranteed to be the same on a different version of the game or on a different platform
    2. There are no unsigned types
    3. There are no big endian/little endian variants
    4. There is no guarantee that we get 1, 2, 4 and 8 byte types. e.g int might be the same size as short
    5. basically you need int8_t, int16_t, ..., uint64_t
3. There are no read_byte() or write_byte() in CBOT. Are you thinking of another language?
4. You only need one integer data type to be able to read any sized integer (two if you want 64 bit unsigned integers without doing what [python](https://stackoverflow.com/a/20980470) does):
```
int read_i8be(...)
int read_i16be(...)
int read_i32be(...)
int read_i64be(...)
```

PR #1256, opened 2019-04-28
Fix compilation on MSYS2 after boost libraries update

tomangelo2:
As mentioned in linked issue, recently Boost libraries in MSYS2 were updated in a way that broke old CMake scripts, making it unable to link them. This should fix it, but I want to be sure that this won't break older versions of Boost library and/or other compilers.


tomangelo2:
Seems like it isn't a thing since some time, MSYS2 builds are going good, so there is no point of merging that anymore.

PR #1258, opened 2019-05-12
Allow negative numbers in 'case' and prohibit non-numbers.

### LINKED ISSUE #1008: switch..case not accept negative values

immibis:
Fixes #1008

### DIFF
--- a/src/CBot/CBotInstr/CBotExprLitNum.cpp
+++ b/src/CBot/CBotInstr/CBotExprLitNum.cpp
@@ -48,7 +48,14 @@ CBotInstr* CBotExprLitNum::Compile(CBotToken* &p, CBotCStack* pStack)
     CBotExprLitNum* inst = new CBotExprLitNum();
 
     inst->SetToken(p);
-    std::string    s = p->GetString();
+
+    // Parsing a 'case' expression calls this. The token is whatever the player typed after 'case'
+    // so we need to check that it's actually a number.
+    // For some reason defines are also hacked on here so they can be treated as numbers.
+
+    // We also handle negative numbers which appear as two tokens: ID_SUB followed by the number.
+    // Note: - followed by a define is not supported, only - followed by a number.
+    // The case of - followed by a define will be handled more generically the future if we implement constant folding.
 
     inst->m_numtype = CBotTypInt;
     if (p->GetType() == TokenTypDef)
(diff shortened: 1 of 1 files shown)


immibis:
I've made a comment about constant folding. Looking through the compiler code, I think the whole compiler could use an overhaul. But I could be wrong.

src/CBot/CBotInstr/CBotExprLitNum.cpp
```diff
@@ -75,6 +108,8 @@ CBotInstr* CBotExprLitNum::Compile(CBotToken* &p, CBotCStack* pStack)
 
         return pStack->Return(inst, pStk);
     }
+    // Program ended after the number. Return error indication.
+    // XXX: Why are we returning an error indication *here*, when the end-of-program doesn't actually affect the number parsing?
```
melex750: In fact, the code after your comment will never execute because the next token will never be nullptr.

immibis: NextToken returns true (and updates p by reference) if there was a next token and otherwise returns false. How do you know there will always be a next token?

melex750: Currently the only places where numbers or other expressions will be compiled are within the definition of a function or class, which are verified by structure before any such expressions are compiled. Which is to say, always somewhere between curly braces or parentheses.

And beyond that, CBotToken::CompileTokens() appends a dummy token at the end, so the program will always have a first and last token.  The first token, by the way, is only used to consume any comments or white-space that may be at the top of the program script.

immibis: I suppose this sort of thing is why the compiler needs a rewrite. (Does it though? Sometimes I jump into code and think it needs a rewrite, because I don't understand it yet)

MrSimbax:
> This and a few other issues with switch...case are fixed in #1253.

Indeed, the issue seems fixed. Then I guess I should close this.

PR #1263, opened 2019-07-02
Fix OpenGL specular lighting

nextghost:
I've taken a look at GL shaders in Colobot recently and I've noticed that the specular lighting calculation is completely broken. Every single fragment gets specular light component equal to 1 regardless of light direction and camera position. Fortunately, the problem is hidden for now because every single material in the game has specular reflection of (0; 0; 0).

This pull request fixes specular lighting calculation in OpenGL 2.1 renderer. The same fix needs to be applied to OpenGL 3.3 renderer but unfortunately it's not a trivial copy-paste and my Intel GPU only supports OpenGL 3.0 so I wouldn't even be able to compile the shaders to check for syntax errors.

You can force enable specular lighting for testing by editing `src/graphics/opengl/shaders/gl21/fs_normal.glsl` and temporarily deleting the multiplication by `uni_Material.specular` on line 118. There are some outstanding issues with light sources in game data which sometimes have vertical direction of 0 (so they're effectively in the horizontal plane relative to camera position). This produces visual artifacts like below:

colobot-specular

While this would be a perfectly valid specular effect for a light source close above horizon (like a sunrise or sunset), the problem is that the light source here is actually below horizon and it's shining through the terrain. The light source is located infinitely far away at the tip of the 2D "light cone". Changing the vertical direction of the light source in level definition file fixes these artifacts.

### DIFF
--- a/src/graphics/opengl/gl21device.cpp
+++ b/src/graphics/opengl/gl21device.cpp
@@ -388,6 +388,7 @@ bool CGL21Device::Create()
         uni.modelMatrix = glGetUniformLocation(m_normalProgram, "uni_ModelMatrix");
         uni.normalMatrix = glGetUniformLocation(m_normalProgram, "uni_NormalMatrix");
         uni.shadowMatrix = glGetUniformLocation(m_normalProgram, "uni_ShadowMatrix");
+        uni.cameraPosition = glGetUniformLocation(m_normalProgram, "uni_CameraPosition");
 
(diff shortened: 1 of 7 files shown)


nextghost:
Actually, after a forced hardware upgrade earlier this month (old laptop died), I now have a GPU that supports OpenGL 4.6. I'll fix the OpenGL 3.3 engine tomorrow.

nextghost:
Specular lighting fix for OpenGL 3.3 added. I've also added initialization of camera position to (0;0;0) to the OpenGL 2.1 commit.

PR #1274, opened 2019-09-24
Fix manpage build rules to work with Ninja generator

mgorny:
Specify outputs for po4a command explicitly rather than via implicit
target, in order to clearly establish the dependency chain between
manpages and input files generated by po4a.  This is necessary to make
it possible to use Ninja generator.  Otherwise, the implicit
dependencies cause the following error:

    ninja: error: '../desktop/lang/fr/colobot.pod', needed by 'desktop/fr/colobot.6', missing and no known rule to make it


mgorny:
Ping.

Issue #1268, opened 2019-08-11
Crash on player name fifth character


osvetlik:
When I tried to enter my usual player name which is five characters long, the game crashed. It happens on any name just when you type the fifth character. With shorter name, I can normally play.

Erihel:
Is that last character non latin letter? Might be related to #922

osvetlik:
No, it's any five characters, 'aaaaa' for example. But there's another curious thing, this only happens the first time, when no configuration folder exists. When creating a second user, I can type as many characters I want. When I remove the configuration folder (~/.var/app/info.colobot.Colobot), it starts crashing again. As long as no user exists, it crashes on the fifthe character, once a user is created, it stops crashing for all the other users. I even tried to remove just the savegames for the existing user, but it is still possible to create a new user with a longer name.

tomangelo2:
Which OS it is? I assume Linux, but what distro? Game is installed from repository or ran from AppImage?

osvetlik:
Fedora 30, installed via gnome-software, but not from RPM, some FlatPak or something like this, I assume. Version 0.1.12-alpha.

osvetlik:
Yes, it's flatpak, I managed to run it from bash and here is the result:
/usr/include/c++/8.2.0/bits/basic_string.h:1039: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::const_reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator[](std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::size_type) const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::const_reference = const char&; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::size_type = long unsigned int]: Assertion '__pos <= size()' failed.

Unhandled exception occurred!
==============================
SIGABRT, abort
==============================

tomangelo2:
Couldn't reproduce that. Probably must be something more.

suve:
I am the maintainer of the `colobot` package in Fedora, so I looked into this. Since you mention Flatpak, and I haven't created a flatpak version of the Fedora package, I guess you installed it from FlatHub. Looking at FH, the source repo for the Flatpak is here: https://github.com/flathub/info.colobot.Colobot

[...]

PR #1276, opened 2019-10-07
Detect invalid values in StringUtils::Utf8CharSizeAt()

suve:
This changeset fixes the `StrUtils::Utf8CharSizeAt()` method not recognizing UTF-8 continuation bytes and erroneously returning `3` for those.

Related to issue #1268.

### DIFF
--- a/src/common/stringutils.cpp
+++ b/src/common/stringutils.cpp
@@ -173,14 +173,19 @@ int StrUtils::Utf8CharSizeAt(const std::string &str, unsigned int pos)
     if (pos >= str.size())
         return 0;
 
-    if ((str[pos] & 0x80) == 0)
-        return 1;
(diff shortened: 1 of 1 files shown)


suve:
One thing to consider here - should the code `return 0;` on an invalid byte? If you scroll the code just a bit further, you'll see that returning 0 will cause `Utf8StringLength()` to loop infinitely.

Two ideas:
1. Use `return -1;` could be more appropriate, as it allows to distinguish between the "index past string length" case from the "malformed UTF-8" case. That still wouldn't solve the issue of calling code looping infinitely.
2. Go the "no silent errors" route and just throw an exception.

suve:
Changed to throw `std::invalid_argument`.

PR #1288, opened 2020-01-22
Save screenshots using PHYSFS instead of FILE*

melex750:


CookiePLMonster:
Doesn't `COutputStream` destructor close the stream/file? That'd make explicit `close` calls before returns redundant, as RAII should do that instead.

melex750:
It looks like ```~COutputStream``` and superclass d-tors do not close the stream and
there are explicit ```close``` calls in many other places so maybe it was designed that way.

CookiePLMonster:
I see, that's a shame. You'd expect a stream-like interface to do RAII properly, but looks like it's not the case.

melex750:
Inherited data member ```COutputStream::m_buffer``` does close the stream in its d-tor,
so it probably does work as you suggested.

PR #1290, opened 2020-02-02
Support new-format text models through the same code path as old models; update Blender script

immibis:
I made several fixes to the Blender import/export addon (see the commit messages for details).

Also, `COldModelManager` will now load .txt format models. There is a `CModelManager` which is supposed to load these models, but it seems to work completely different to `COldModelManager`. This patch allows .txt models to be loaded exactly the same way as .mod models, from the models folder. The caller does need to know the extension, so you can't just replace a .mod with a .txt without changing the code.


immibis:
When I click on the "Details" for the test failures, it doesn't show any failures.

MrSimbax:
The failures are from the linter and were fixed in https://github.com/colobot/colobot/pull/1295

Issue #1289, opened 2020-01-30
CBotExprLitString.cpp: error: expected unqualified-id before '&' token


suve:
Hi, I'm the maintainer of the `colobot` package in Fedora. During the automated Mass Rebuild of packages for the upcoming Fedora 32, the `colobot` package failed with the following error:
```
/builddir/build/BUILD/colobot-colobot-gold-0.1.12-alpha/src/CBot/CBotInstr/CBotExprLitString.cpp: In static member function 'static CBot::CBotInstr* CBot::CBotExprLitString::Compile(CBot::CBotToken*&, CBot::CBotCStack*)':
/builddir/build/BUILD/colobot-colobot-gold-0.1.12-alpha/src/CBot/CBotInstr/CBotExprLitString.cpp:118:59: error: expected unqualified-id before '&' token
  118 |                             catch (const std::out_of_range& e)
      |                                                           ^
/builddir/build/BUILD/colobot-colobot-gold-0.1.12-alpha/src/CBot/CBotInstr/CBotExprLitString.cpp:118:59: error: expected ')' before '&' token
  118 |                             catch (const std::out_of_range& e)
      |                                   ~                       ^
      |                                                           )
/builddir/build/BUILD/colobot-colobot-gold-0.1.12-alpha/src/CBot/CBotInstr/CBotExprLitString.cpp:118:59: error: expected '{' before '&' token
/builddir/build/BUILD/colobot-colobot-gold-0.1.12-alpha/src/CBot/CBotInstr/CBotExprLitString.cpp:118:61: error: 'e' was not declared in this scope
  118 |                             catch (const std::out_of_range& e)
      |                                                             ^
```
The compiler used is GCC 10.0.1 and the commit used is tagged `colobot-gold-0.1.12-alpha`.

You can view the full build log here: https://kojipkgs.fedoraproject.org/work/tasks/8749/41138749/build.log

Any help will be much appreciated.

melex750:
Maybe ?
```c++
#include <stdexcept>
```

suve:
Yep, that worked. Thanks!

melex750:
This issue also exists for CBotExprLitChar.cpp on dev branch.
Maybe @suve could try compiling dev and find other errors.

suve:
I had to re-build the `colobot` package for Fedora 33 and came across another missing include on the `master` branch:
```
[...]

PR #1291, opened 2020-02-07
Add missing includes in CBotExprLitString.cpp

suve:
Without this `#include`, GCC 10.0.1 used in Fedora 32 (unreleased yet) complains loudly.

Fixes issue #1289.

### DIFF
--- a/src/CBot/CBotInstr/CBotExprLitString.cpp
+++ b/src/CBot/CBotInstr/CBotExprLitString.cpp
@@ -24,6 +24,8 @@
 
 #include "CBot/CBotVar/CBotVar.h"
 
+#include <stdexcept>
+
 namespace CBot
 {
 


src/CBot/CBotInstr/CBotExprLitString.cpp
```diff
@@ -16,6 +16,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program. If not, see http://gnu.org/licenses
  */
+#include <stdexcept>
```
melex750: I think this will cause the linter to complain.
#includes with ```<...>``` angle brackets should be placed after #includes with ```"..."``` quotes and in alphabetical order.

suve: Fixed by ammending the commit.

PR #1292, opened 2020-02-24
Add info about Fedora package to README

szpak:
To let Fedora people know there is an easy way to install Colobot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ On some Linux distributions there are also distribution packages available:
  * Debian Sid (unstable): http://packages.debian.org/sid/colobot
  * Arch Linux (AUR): https://aur.archlinux.org/packages/colobot-gold
  * openSUSE: http://software.opensuse.org/download.html?project=games&package=colobot
+ * Fedora: https://src.fedoraproject.org/rpms/colobot
+   - `sudo dnf install colobot`
 
 
 ## Compiling and running the game


elkoniu:
I found this PR by mistake walking around repo. I am on Fedora and it was truly useful for me.

szpak:
@MrSimbax Could you merge it?

README.md
```diff
@@ -30,6 +30,8 @@ On some Linux distributions there are also distribution packages available:
  * Debian Sid (unstable): http://packages.debian.org/sid/colobot
  * Arch Linux (AUR): https://aur.archlinux.org/packages/colobot-gold
  * openSUSE: http://software.opensuse.org/download.html?project=games&package=colobot
+ * Fedora: https://src.fedoraproject.org/rpms/colobot
+   - `sudo dnf install colobot`
```
elkoniu, line 34: You should stick to link only rather than full install command - other links do not have them.

szpak: I could, however, on that page there is no instructions how to install the package. It would be much more intuitive for the users to leave it as:
```
 * Fedora: sudo dnf install colobot
```

(without the link at all).

WDYT?

elkoniu: I agree with MrSimbax #2: https://github.com/colobot/colobot/pull/1292#pullrequestreview-406767494

szpak:
It was just a quick fix via GH interface. As rebasing is problematic from UI I created an another PR: #1297.

PR #1307, opened 2020-07-04
Partially modernize CMake

immibis:
CMake now allows targets to export include directories (`target_include_directories` with `PUBLIC` or `INTERFACE`) and other targets which link with that target will automatically import the include directories as well. Same for libraries. Thus we can bundle all those INCLUDE_DIRS and LIBRARY variables into the libraries themselves.

Out of the libraries that cmake finds for us, Boost and libpng already do this.

`INTERFACE IMPORTED` declares a dummy empty library which depends on the actual one we want. CMake *does* have a way to say that SDL2::SDL2 is the *actual* SDL2 library - but it's more complicated, and it's different depending on e.g. whether SDL2 is a static or shared library. This way is easier.

Note, this PR does not convert the entire build system to "modern CMake" - only one part.


immibis:
Closed since Simbax has already done this in his branch

Issue #1306, opened 2020-07-03
Modernize and clean up the build process


MrSimbax:
We're currently using a very old version of CMake (pre 3.0) and the 3 version seems almost like changing from Python 2 to Python 3. It is generally advised to do things a little bit differently now. See https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1 for example. I think we should update to avoid compatibility issues in the future and hopefully it will simplify/fix some things.

The plan:

- [ ] Bump CMake versions.
- [ ] Rewrite various CMakeLists.txt files so that they follow good practices and use new CMake features.
- [ ] Remove as much platform/compiler dependent code as possible, so the building process is more robust (this is especially important for Windows, our way of handling this platform looks like a bunch of hacks which barely work... sometimes. Maybe it can't be done better, IDK, I hope it can). Do we really need to check for specific compiler versions and other low-level details like that?
- [ ] Remove unnecessary Find*.cmake files. Some of them are bundled with new versions of CMake. If not, then they probably could use an update too.

It's not exactly related to CMake but in the meanwhile I'd also like to address some other technical issues I noticed, which might be more controversial.

- [ ] Bump C++ standard to C++17. Unless we want to support old compilers, I don't see why not. It won't hurt, it may help in the future.
- [ ] Remove/replace unnecessary dependencies. For example, do we still need boost if both filesystem and regex are available in the standard library?
- [ ] Replace platform-dependent code with header-only libraries where it makes sense. Then we could simplify and remove some ifs from CMakeLists.txt which add specific flags, additional libraries etc. depending on the platform (adding flags by hand should be generally avoided in modern CMake). Why header-only? Well, they are simple to use and don't add another component to the already long list of dependencies to install for users.
- [ ] Get rid of colobot-lint, it wasn't updated in a while and there are now seemingly good linters available for C++, like clang-tidy, so it no longer seems necessary. Correct me if I'm wrong.
- [ ] Use git submodules or CMake's FetchContent for dependencies in the lib folder? This way they will be easier to keep up to date.
- [ ] Remove the -Werror flag ( https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#treat-warnings-as-errors )
- [ ] Move our CI to GitHub Actions. Why? Well, it's basically free CI, so we might as well at least try it and if it works well, we may no longer need the Jenkins server.
[...]

PR #1313, opened 2020-07-06
Remove dependency on `sed` for simple substitution in CMake

MrSimbax:
This is part of issue https://github.com/colobot/colobot/issues/1306

The current implementation didn't work for me due to escaping issues, then I discovered that the `sed` binary itself from GnuWin32 is buggy (option `-i` makes it spawn temporary files and they're not removed automatically).

So, yeah, I've just wasted my time implementing a basic `sed` in CMake. But it works at least in the case of `colobot.pot`.

I also added some comments so nobody in the future has to waste their time like me on figuring out what those parts of the scripts do and why.

Also changed the comment in the `restext.cpp` file so it says to use a multiplatform CMake command instead of the unportable `make update-pot` command.

### DIFF
--- a/cmake/replace_in_file.cmake
+++ b/cmake/replace_in_file.cmake
@@ -1,0 +1,37 @@
+# The script replaces all occurences of REGEX_MATCH with the REPLACE_WITH string.
+# The regex rules are like in CMake except the matches are done line by line.
+# The purpose of the script is to remove the dependency on `sed` for simple substitutions.
+# It's not a function so that it can be used in add_custom_command().
+#
+# Parameters:
+# - REGEX_MATCH
+# - REPLACE_WITH
+# - FILE_PATH
+#
+# Example usage:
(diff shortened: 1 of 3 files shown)


MrSimbax:
Apparently, the `file(WRITE ...)` command doesn't respect its input and transforms all line endings into `\r\n` on Windows. I haven't noticed this thanks to `git` `autocrlf ` option. Unfortunately, according to the documentation of CMake 3.18 there's no way to specify line endings for the `file(WRITE ...)` command.

I'm leaving this pull open for now since the issue with `sed` still needs fixing for compatibility with MSVC. Maybe me or someone else will figure it out later.

MrSimbax:
I'm closing this PR because I found a way to fix the sed command. See https://github.com/colobot/colobot/pull/1329 It's a shame that CMake doesn't provide a multiplatform way to replace text in files.

PR #1314, opened 2020-07-06
Make the update-pot target run automatically

MrSimbax:
Because people forget about running it
despite the comment in `restext.cpp`.

If there's a reason it should not be in the ALL target then discuss below.

### DIFF
--- a/po/CMakeLists.txt
+++ b/po/CMakeLists.txt
@@ -19,7 +19,7 @@ add_custom_command(OUTPUT ${_potFile}
     VERBATIM
 )
 
-add_custom_target(update-pot DEPENDS ${_potFile})
+add_custom_target(update-pot ALL DEPENDS ${_potFile})
 
 # Generate translations
 


krzys-h:
I believe the reason is that it modifies the source files that get commited to the repo. You can't depend on gettext to generate exactly the same file on different platforms even if the source hasn't changed due to version differences, sometimes it can reorder things etc. (we've run into it before I think?). This takes it from "people forget to run update-pot" to "people accidentally commit insignificant changes to .pot files".

And you still need to remember to run make just before commit anyway. I can see a scenario where you do something, test your changes, realize that it works but you forgot a comma, add the missing comma and commit without running make again. Doing it explicitly is better imo.

MrSimbax:
>I believe the reason is that it modifies the source files that get commited to the repo.
The translations in data are automatically changing all the time, so what's the difference?

krzys-h:
It's probably "the person writing the scripts in the main repo thought about this, but the person writing the data translations didn't" ¯\\\_(ツ)\_/¯

But if you want an actual argument, iirc the data files use a python script to generate the .pot files rather than depending on the system gettext binaries, which makes it less prone to changes depending on what platform you are running it on. But I never really liked the fact that the data files do this automatically because if you change something in the scene.txt file, run make and then revert the change, the translations get lost.

MrSimbax:
You have a point. Maybe we should switch to Python script in this case too? Reproducible builds and all that stuff...

Either way, it seems we'll have to make an enormous banner, and put it somewhere where all contributors have to look before creating a pull request with a typo or a change in translatable text, saying that they should also update translations.
...
In fact, such a banner would be useful for us devs too.

PR #1329, opened 2020-07-10
Fix update-pot not working with MSBuild

MrSimbax:
Fixes the `'\1' is not recognized as an internal or external command` error when trying to run the target `update-pot` with MSBuild.

The `|` characters probably messed with cmd.exe or PowerShell syntax rules so they were replaced with slashes `/`.

Also, at least some implementations of sed for Windows would produce CRLF line endings instead of LF line endings. The issue is fixed by adding the flag `-b`.

Also changed the `make update-pot` comment to a portable cmake command. `make update-pot` only works if a generator producing Makefiles was used with `cmake`. The `cmake` command to build a specific target will work for any generator.

Note: I recommend using the following `sed` implementation: https://github.com/mbuilov/sed-windows instead of the one provided by the GnuWin32 project since the latter doesn't clean up temporary files which are created when the in-place option `-i` is used.

### DIFF
--- a/po/CMakeLists.txt
+++ b/po/CMakeLists.txt
@@ -12,7 +12,7 @@ add_custom_command(OUTPUT ${_potFile}
     COMMAND ${XGETTEXT_CMD} ${colobot_SOURCE_DIR}/src/app/app.cpp --output=${_potFile} --no-wrap
     COMMAND ${XGETTEXT_CMD} ${colobot_SOURCE_DIR}/src/common/restext.cpp --output=${_potFile} --no-wrap --join-existing --no-location --keyword=TR
     COMMAND ${XGETTEXT_CMD} ${colobot_SOURCE_DIR}/src/script/script.cpp --output=${_potFile} --no-wrap --join-existing --no-location
-    COMMAND sed -i -e "s|^\\(\"POT-Creation-Date:\\).*$|\\1 DATE\\\\n\"|" ${_potFile}
+    COMMAND sed -bi -e "s/^\\(\"POT-Creation-Date:\\).*$/\\1 DATE\\\\n\"/" ${_potFile}
 
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     COMMENT "Extract translatable messages to ${_potFile}"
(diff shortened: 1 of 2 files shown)


krzys-h:
Is the `-b` option even a thing in Linux sed?
```
╭─krzys_h@krzysh.pl ~
╰─➤  LC_ALL=C sed --help                                                                                                                                                                                                                 4 ↵
Usage: sed [OPTION]... {script-only-if-no-other-script} [input-file]...

  -n, --quiet, --silent
                 suppress automatic printing of pattern space
  -e script, --expression=script
                 add the script to the commands to be executed
  -f script-file, --file=script-file
                 add the contents of script-file to the commands to be executed
  --follow-symlinks
                 follow symlinks when processing in place
  -i[SUFFIX], --in-place[=SUFFIX]
                 edit files in place (makes backup if SUFFIX supplied)
  -l N, --line-length=N
                 specify the desired line-wrap length for the `l' command
  --posix
                 disable all GNU extensions.
  -r, --regexp-extended
                 use extended regular expressions in the script.
  -s, --separate
                 consider files as separate rather than as a single continuous
                 long stream.
  -u, --unbuffered
                 load minimal amounts of data from the input files and flush
                 the output buffers more often
  -z, --null-data
                 separate lines by NUL characters
      --help     display this help and exit
      --version  output version information and exit

If no -e, --expression, -f, or --file option is given, then the first
non-option argument is taken as the sed script to interpret.  All
remaining arguments are names of input files; if no input files are
specified, then the standard input is read.

GNU sed home page: <http://www.gnu.org/software/sed/>.
General help using GNU software: <http://www.gnu.org/gethelp/>.
E-mail bug reports to: <bug-sed@gnu.org>.
Be sure to include the word ``sed'' somewhere in the ``Subject:'' field.
```

MrSimbax:
I tested the changes also on Ubuntu 18.04 and update-pot worked fine so I guess it does have the flag or at least it is ignored with no error.

MrSimbax:
The flag is listed here for GNU sed https://www.gnu.org/software/sed/manual/sed.html

krzys-h:
Ok, so I guess it's just not displayed in the short listing. My bad.

MrSimbax:
I guess the flag is not listed on Linux platforms because it doesn't do anything there :P

PR #1330, opened 2020-07-10
Use CMake targets instead of variables

MrSimbax:
This is definitely work in progress.

The plan:
- [x] Use targets from 3rd party libs instead of variables
- [x] Remove redundant variables in src/CMakeLists.txt
- [x] Make tests compile
- [x] Make tools compile
- [x] Update GTest and make it a submodule
- [x] Fix linter in CI
- [ ] ~~Find all dependencies of dependencies and platform dependencies (yes, this is madness) needed for static linking (who knows, maybe we'll finally be able to compile statically on Linux after all this)~~
- [ ] ~~Clean up static linking (at least on Windows)~~
- [x] Make it build on Ubuntu
- [x] Make it build with MXE
- [x] Make it build with MSVC

This PR is a major step towards a better CMake. Hopefully next steps towards modernizing the build process will be smaller and easier.

New features:
- Everything is targets and is used as it should in modern CMake. No more ${OPTIONAL_LIBS} variables and so on.
- HIDE_CONSOLE flag. It should work on MSYS2 too. Note: it only makes sense on Windows, otherwise does nothing.
- ~~BUILD_STATIC flag. It sets default values for ${LIB}_STATIC, which can still be controlled by the user if only some libraries should be STATIC. I advise against such odd configurations in case of MSVC though, since there may be linking problems with the MSVC runtime library.~~
- Add USE_STATIC_RUNTIME flag, mainly useful for MSVC (it uses linker flag /MT instead of /MD)
- Google Test was updated to new version and is now a submodule.
- Building tests and tools is now possible on MSVC.
- GitHub Actions is using a native Windows machine with MSVC to build the binary, ~~although without sound for now since OpenAL cannot be build with vcpkg for some reason on the GH Actions machines (see comment https://github.com/colobot/colobot/pull/1330/files#diff-898e737ee4e56ba042bda31764bed49eR34).~~
- The C++ standard is now set consistently by CMake, we no longer have to worry about giving specific flags to different compilers, at least in this case.

Caveats:
~~- Removed MSVC_STATIC flag and replaced it with BUILD_STATIC + more {LIB}_STATIC flags. In theory the BUILD_STATIC flag should be platform independent but it was not tested outside of MSVC. I believe more must be done about linking, since we're finding library packages "by hand" (because most Linux packages are outdated and don't export targets in {Lib}Config.cmake, and some libs simply don't generate this file at all), we should figure out their dependencies and link them properly too.~~
~~- I wasn't able to make it build on MXE. I hope everybody agrees we can simply drop support for it now since we can have CI on a Windows machine. Cross-compilation is simply too painful.~~
- Release builds on MSVC no longer hide console by default. The behavior was replaced with more flexible HIDE_CONSOLE flag.
- Support for folders in IDEs like Visual Studio (without the Ninja generator) and Xcode is dropped. Since nobody seems to use them, I think it's ok. It may be redone properly in a modern way in another pull request.
- There was a lot of changes to finding libraries. It's possible they may break building in some environments (especially in MXE). Although I think the finding should be now more robust in general.
- The game probably no longer compiles on Visual Studio before 2017 or even before 2019? Not really an issue in my opinion.
~~- MAXSTACK in CBOT was reduced significantly because the "stack overflow" test on MSVC was failing.~~
- MSVC still not tested with dynamic linking. I consider this a subject for another PR though as we never really supported dynamic linking with MSVC anyway. I believe the changes in this PR, if dynamic linking doesn't work already, will make it easier to support it.
~~- Until the issue with OpenAL is resolved (see above), the Windows builds have no sound. Since we're not merging to `dev` yet I think we can give it a pass for now and wait with merging to `dev` until the issue is fixed or a workaround presents itself.~~
- Google Test is now a submodule :P
- Unfortunately, there is more Find*.cmake modules now. Many libraries still don't export targets in *Config.cmake files, and even then it'll take some time before they appear in Linux package managers. So we have to find a lot of stuff "manually" and modify old Find*.cmake modules so that they export imported targets.
- Probably more I haven't listed here.


.github/workflows/build.yml
```diff
@@ -5,39 +5,84 @@ on: [push, pull_request]
...
-      if: matrix.container == ''
     - uses: actions/checkout@v2
+    - name: Checkout the Google Test submodule
+      run: git submodule update --init -- lib/googletest
+    - name: Install Colobot dependencies (for Linux)
+      run: sudo apt-get update && sudo apt-get install -y --no-install-recommends build-essential cmake libsdl2-dev libsdl2-image-dev libsdl2-ttf-dev libsndfile1-dev libvorbis-dev libogg-dev libpng-dev libglew-dev libopenal-dev libboost-dev libboost-system-dev libboost-filesystem-dev libboost-regex-dev libphysfs-dev gettext git po4a vorbis-tools librsvg2-bin xmlstarlet
+      if: matrix.target_os == 'linux'
+    - name: Install Ninja
+      uses: seanmiddleditch/gha-setup-ninja@master
+      if: matrix.target_os == 'windows'
```
krzys-h, line 25: I no longer like how the `if` steps work. I was hoping we could eventually replace them with platform-independent actions like `run-cmake`, but it seems like this is going in exactly the opposite direction. Perhaps it would be cleaner to separate this into `build_linux` and `build_windows` instead, or somehow extract the platform-dependent dependency setup elsewhere?

.github/workflows/build.yml
```diff
@@ -5,39 +5,84 @@ on: [push, pull_request]
...
+      run: cmake -G "Ninja" -DCMAKE_C_COMPILER=cl.exe -DCMAKE_CXX_COMPILER=cl.exe -DVCPKG_TARGET_TRIPLET=x64-windows-static -DCMAKE_TOOLCHAIN_FILE="${{ env.VCPKG_ROOT }}\scripts\buildsystems\vcpkg.cmake" -DBUILD_STATIC=1 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}\build\install -DDEV_BUILD=1 -DPORTABLE=1 -DTOOLS=1 -DTESTS=1 -DDESKTOP=1 -DOPENAL_SOUND=0 ..
       if: matrix.target_os == 'windows'
     - name: Run CMake (for Linux)
       working-directory: build
-      run: cmake -DCMAKE_INSTALL_PREFIX=/install -DCMAKE_SKIP_INSTALL_RPATH=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DDEV_BUILD=1 -DPORTABLE=1 -DTOOLS=1 -DTESTS=1 -DDESKTOP=1 ..
+      run: cmake -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/build/install -DCMAKE_SKIP_INSTALL_RPATH=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DDEV_BUILD=1 -DPORTABLE=1 -DTOOLS=1 -DTESTS=1 -DDESKTOP=1 ..
```
krzys-h, line 78: There was a reason this was not a full path, but don't ask me what it was because I don't remember. Have you checked that the defaults for things like the datadir are correct with this on both Windows and Linux?

CMakeLists.txt
```diff
@@ -257,19 +256,27 @@ option(INSTALL_DOCS "Install Doxygen-generated documentation" OFF)
 # Build OpenAL sound support
 option(OPENAL_SOUND "Build OpenAL sound support" ON)
 
+# Build Colobot with static libraries
+option(BUILD_STATIC "The default linking option for external libraries" OFF)
+
+# CBot can also be a static library
+option(CBOT_STATIC "Build CBot as static libary" ${BUILD_STATIC})
```
krzys-h, line 263: Have you checked how this behaves when changing options for an existing build dir without clearing the CMake cache? I mean something like this:
```
mkdir build; cd build
cmake -DBUILD_STATIC=ON ..  # enable static builds
cmake -DBUILD_STATIC=OFF .. # disable static builds, without clearing the build dir. Are options like CBOT_STATIC stuck on "true" after this?
```

CMakeLists.txt
```diff
@@ -301,13 +308,13 @@ set(Boost_USE_STATIC_RUNTIME OFF)
...
 
+set(GLEW_USE_STATIC_LIBS ${GLEW_STATIC})
 find_package(GLEW REQUIRED)
 
 if (OPENAL_SOUND)
     find_package(OpenAL REQUIRED)
-    include_directories(${OPENAL_INCLUDE_DIR})
-    find_package(LibSndFile REQUIRED)
 endif()
+find_package(SndFile REQUIRED)
```
krzys-h, line 317: This is not required if sound support is disabled, as it's not used anywhere in that case

src/CBot/CBotDefines.h
```diff
@@ -20,7 +20,7 @@
 #pragma once
 
 #define    STACKMEM    1                /// \def preserve memory for the execution stack
-#define    MAXSTACK    990              /// \def stack size reserved
+#define    MAXSTACK    530              /// \def stack size reserved
```
krzys-h, line 23: try https://docs.microsoft.com/en-us/cpp/build/reference/f-set-stack-size?view=vs-2019 instead?

src/CMakeLists.txt
```diff
@@ -1,20 +1,10 @@
 # Compile flags as defined in global CMakeLists
+# Todo: replace with target properties
```
krzys-h, line 2: is this TODO supposed to still be there?

src/CMakeLists.txt
```diff
@@ -589,79 +469,126 @@ set(BASE_SOURCES
...
-)
+# Additional libraries per platform
+if(MXE) # MXE requires special treatment
+    target_link_libraries(colobotbase PUBLIC ${MXE_LIBS})
+elseif(PLATFORM_WINDOWS)
+    # because it isn't included in standard linking libraries
+    if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+        if(BUILD_STATIC)
+            find_package(Intl REQUIRED)
+            #todo figure out why those dependencies are needed and find them in a better way
```
krzys-h, line 522:
```suggestion
            # TODO: figure out why those dependencies are needed and find them in a better way
```

src/common/config.h.cmake
```diff
@@ -12,8 +12,6 @@
 #define HAVE_CFLOCALECOPYCURRENT 1
 #endif
 
-#cmakedefine GLEW_STATIC
```
krzys-h, line 15: I think the GLEW headers did actually require this to work. Does the new code set it somewhere else?

test/cbot/CMakeLists.txt
```diff
@@ -1,19 +1,13 @@
...
+add_executable(CBot_console console.cpp)
+target_link_directories(CBot_console PRIVATE
     ${CMAKE_CURRENT_SOURCE_DIR}
     ${CMAKE_CURRENT_BINARY_DIR}
 )
-
-# Libraries
-set(LIBS
-    CBot
-    colobotbase # Needed for error strings (TODO: why are they on Colobot side? :/)
```
krzys-h, line 11: you removed the comment :<

.github/workflows/build.yml
MrSimbax: I agree. Was thinking about splitting the builds up but was too busy trying to make it work in the first place.

.github/workflows/build.yml
```diff
@@ -5,39 +5,84 @@ on: [push, pull_request]
...
+        vcpkgArguments: 'boost-system boost-filesystem boost-regex boost-lexical-cast boost-bimap boost-algorithm boost-property-tree boost-optional boost-range sdl2 sdl2-ttf sdl2-image glew libpng libwebp tiff gettext libsndfile libvorbis libogg physfs'
+        # SHA-256 hash of the list of packages above, for caching purposes
+        appendedCacheKey: '6654161b6d89bd7ad950e2eb4c1bb85c191b98202da52fc6fd18fbea28a8ea50'
+      if: matrix.target_os == 'windows'
+    - name: Upload vcpkg logs
+      uses: actions/upload-artifact@v2
+      with:
+        name: vcpkg_logs
+        path: vcpkg/buildtrees/openal-soft/*.log
+      if: always()
```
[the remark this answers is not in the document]
MrSimbax: OpenAL is not currently being installed so nothing will be uploaded. The always() is necessary for the upload to work even after fail, I'm not aware of a better way to make this work only when the previous step fails. Also, I left it there for debugging purposes until the issue with OpenAL is fixed. I probably should remove this anyway, the code is in history and there doesn't seem much we can do about it anyway. The only fix on our side would require patching vcpkg or OpenAL by ourselves somehow. I subscribed to related issues in other repos to know when the issue is fixed, which is hopefully soon.

.github/workflows/build.yml
MrSimbax: I'm aware. I was hoping that whatever the issue was it no longer exists since the intruduction of the USE_RELATIVE_PATHS flag. I'm yet to test the AppImage from the builds though.

src/common/config.h.cmake
MrSimbax: Yes. I was getting a lot of warnings with the new FindGLEW that GLEW_STATIC is defined twice. Don't ask me how and where it is defined then though. I suspect there's some logic about it in the FindGLEW module. The static targets have probably set INTERFACE_COMPILE_DEFINITIONS correctly so we don't have to add definition by hand.

Also, I recall I gave the explanation for the change in the related commit message.

test/cbot/CMakeLists.txt
MrSimbax: Ah, sorry, will bring it back :D I'm not sure why I did this, probably in the hurry I accidentally removed it without thinking.

src/CMakeLists.txt
MrSimbax: I'm not sure any more. We set pretty much the same flags for every target so maybe in this particular case global variables are the way to go, otherwise we would need to repeat a lot of code for setting the flags. I still don't like the fact that we touch the CMAKE_CXX_FLAGS family as it's apparently against good practices but I'm not yet sure what exactly is the "good" practice here and any hints on I've found on the Internet are too vague.

src/CBot/CBotDefines.h
MrSimbax: Why not https://stackoverflow.com/a/156730 while we're at it?

Is there something magical about the number 990?

krzys-h: Yes, there is. It.... allocates 1000 elements for the stack data structure, because it catches any overrun up to 10 stack levels. (╯°□°)╯︵ ┻━┻
https://github.com/colobot/colobot/blob/17447e813b97a68861c2ad257e7cead16724c16b/src/CBot/CBotStack.cpp#L53-L80
The fact that it prevents an actual C++-level stack overflow somewhere in the deep `CBotInstr::Execute` call tree is just a side effect.

src/common/config.h.cmake
krzys-h: Oh yeah, you did, I just didn't really look at the commit messages as I assumed all important things are in the PR summary.

CMakeLists.txt
MrSimbax: No, I haven't, but I doubt it will change the values. It's what I would expect honestly. In my experience changing variables without clearing cache is usually a bad idea anyway and ends up badly. So I haven't even considered this use case and I'm not sure what would be the expected behavior even be. What if the flags are not set consistently? For example, some libraries are STATIC and some are not? What should switching the BUILD_STATIC variable then do? Force rewrite of the cache variables? This would overwrite any changes in the cache set by the user for example from the GUI.

Anyway, maybe we should just drop the (probably bad?) idea of mixing static and dynamic libraries (except the ones we build, like CBOT) altogether and just force everything to be build statically or dynamically with only this one flag, with maybe a changed name to USE_STATIC_EXTERNAL_LIBS or something. I'm definitely not an expert on the matter of linking though so I don't know what are good practices here.

CMakeLists.txt
MrSimbax: It is still required though or I wouldn't have moved this :) I don't know the sound part of the code but if this library shouldn't be required if there's no sound then the abstract interfaces should be rewritten. I had 4 linker errors IIRC so there are at least 4 places where we use libsndfile regardless if OPENAL_SOUND is true or not.

src/common/config.h.cmake
MrSimbax: I looked into this. I don't see FindGLEW defining GLEW_STATIC anywhere. I suspect vcpkg is actually setting it and it's not related to the new FindGLEW module at all. This remind me that this PR should be tested with MSYS2. Might be one of the reasons it didn't build with MXE actually...

What should've be done here is wrapping this definition around `#ifdef GLEW_STATIC` to avoid the "macro redefinition" warnings.

CMakeLists.txt
krzys-h: I'm not sure what to do here either, but I don't think I've ever seen an option to set static linking on per-library basis - it's always either all dynamic or all static.

I also prefer to clean the build dir when switching cmake options, but technically this is a use case CMake supports. One place where it could cause trouble is if somebody tries to (god forbid) use CMake GUI to configure the build options, as in that case BUILD_STATIC would do absolutely nothing.

krzys-h: Also see "What is the CMake cache?" in https://cmake.org/runningcmake/ - it seems that editing the options without clearing is the recommended workflow

CMakeLists.txt
krzys-h: That's weird and sounds like a bug. I remember we used to have some problems with OpenAL on MXE builds and back then it built fine with OPENAL_SOUND set to OFF.

src/common/config.h.cmake
krzys-h: Adding MSYS2 builds to CI too may be a good idea. I think it even comes preinstalled on GH Actions Windows runners.

If this is the only problem with MXE builds, we can try ti bring it back until the OpenAL issue is fixed.

CMakeLists.txt
MrSimbax: So should I try to fix this in this PR or leave it for another PR/someone else to fix?

CMakeLists.txt
krzys-h: After thinking about this, the best option I can think of is something like
```cmake
option(STATIC default OFF)
if (STATIC)
    option(STATIC_CBOT default ON)
    # ...
endif()
```
This still gives you the fine-grained control while keeping the simple "all static" option, and works well with the CMake cache.

Build all static: `-DSTATIC=ON`
Build all static except GLEW: `-DSTATIC=ON -DSTATIC_GLEW=OFF`
Build only GLEW static: this one is problematic, as you need to list all libraries with STATIC_*=OFF ... but it is possible, and I don't expect "single static library" to be used too often

CMakeLists.txt
krzys-h: I think you can just leave it for now

CMakeLists.txt
MrSimbax: I think this flag should just be removed or left as it is -- as a convenient way to force everything to be static on the first run, although I'd rather remove it since it doesn't play well with the cache and I don't think it's worth going out of our way to implement a logic handling this variable how we want. It was supposed to be mainly a convenience thing.  

Also, I introduced a USE_STATIC_RUNTIME flag mainly useful for Windows (although it should be possible with GCC too, but linking statically on Linux doesn't make much sense as far as I know). It better describes what MSVC_STATIC did.

The biggest problem right now is finding the transitive dependencies, especially on Windows since we have to either link those dependencies if the build is static or install the DLLs to the output folder if the build is dynamic. We use a workaround for the former (separately for MXE and MSVC) and used a script for the latter in the past. I hope there's a better way to do this with modern CMake, but we'll have to find those dependencies ourselves since we still can't use find_package(... CONFIG) reliably. But that's a complicated issue I don't want to solve in this already huge PR.

src/common/config.h.cmake
MrSimbax: I think I'll leave MSYS2 builds for another PR. I'd like to introduce an option to install DLLs to CMake first.

Issue #1163, opened 2018-05-09
Prevent creation of overcharged PowerCells


krzys-h:
in produce() and scene files

rbcat:
You're just sweeping it under the rug rather than fixing some actual bugs, aren't ya? OK, OK, I'm not commissioned to tell you what to do...

krzys-h:
If it's not supposed to ever happen, it's not a bug. Having an overcharged PowerCell is like having a string variable "abcd" with a length of 6 - that breaks the invariants that the code assumes to be met. You shouldn't be suprised if the function `get_half_of_string(s)` returns "abc", the bug is not in that function, the bug is that such a weird string is allowed to exist.

rbcat:
Your example with a string is an example of absolute heresy while our case is a mere code imperfection.
> that breaks the invariants that the code assumes to be met
Nowadays you should write functions with as less assumptions as it is possible, and, having such concern on the background, these three functions with their childish imperfections look especially rough and silly, don't they? Colobot, however, is an old enough game coming from the times when a lot of stuff was different, if not polar opposite, thus I will agree with your point nevertheless.

Somebody might want to eventually make a refactoring on them, though....

krzys-h:
If you want to make a power cell that has more charge, we'd have to change the capacity parameter (yes there is such a thing in the code, just not externally accessible), NOT overcharge it. Letting the battery have >100% charge in any form is a bug.

Issue #1071, opened 2018-02-02
Overcharged or discharged power cells take wrong texture


rbcat:
Ever tried to create PowerCell with energy level below 0 or above 1? They start taking part of their textures that they normally shouldn't.

krzys-h:
Overcharged power cells should really not ever exist, so closing this. See #1163.

PR #1332, opened 2020-07-16
Fix for issue #1163

Droog71:
Implemented clamp function as requested in previous pull request #1228 which was closed. There were some formatting issues with the two changed files, so I force pushed the dev branch over this one, then added the new changes, causing the closure of the previous pull request.

This pull request is a fix for issue #1163


src/level/robotmain.cpp
```diff
@@ -3406,6 +3406,14 @@ void CRobotMain::CreateScene(bool soluce, bool fixScene, bool resetObject)
 
                 try
                 {
+                    float min = 0;
+                    float max = 100;
+                    if (params.type == OBJECT_POWER || params.type == OBJECT_ATOMIC) //prevent creation of overcharged or negatively charged power cells
+                    {
+                        max = 1;
+                    }
+                    params.power = Math::Clamp(params.power,min,max);
```
MrSimbax, line 3415: Formatting issue: spaces after commas.

src/object/object_manager.cpp
```diff
@@ -162,7 +162,14 @@ CObject* CObjectManager::CreateObject(Math::Vector pos, float angle, ObjectType
...
     params.type = type;
-    params.power = power;
+    
+    float min = 0;
+    float max = 100;
+    if (params.type == OBJECT_POWER || params.type == OBJECT_ATOMIC) //prevent creation of overcharged or negatively charged power cells
+    {
+        max = 1;
+    }
+    params.power = Math::Clamp(power,min,max);
```
MrSimbax, line 172: This code is almost the same, I think it would be a good idea to extract into a function/method.

Droog71:
Issue #1071 seems to still exist in spite of this change. I remember testing this fix back in 2018 and it was working, but with this current pull request implemented, it seems that the command "produce(PowerCell,2);" still creates a power cell with the wrong texture and 200% power. Not sure if the fix no longer works or there is something wrong with my installation or if there is some other issue.

MrSimbax:
Does it work with the power level set from the scene.txt? Maybe there's one more place where the power is set or there's a bug in the code somehow.

Droog71:
> Does it work with the power level set from the scene.txt? Maybe there's one more place where the power is set or there's a bug in the code somehow.

The fix is actually working. The problem was with my installation of the game. I have two versions of the game installed and ran the wrong one when I was testing earlier which caused some confusion for me. Sorry about that.

MrSimbax:
Good question.

If we only have to care about limits at creation time, then a good place to verify the param may be in the CObjectManager::CreateObject() method (there's already some kind of "validation" of object ID in it). Another good place would be CObjectFactory::CreateResource(). I think the latter makes more sense as it's closer to the PowerCells logic than the more general CreateObject() method. And the power level is actually set there.

If the limits should be kept after the object creation, then some setter would be a good place to implement such logic I guess. But let's keep things simple and assume this only needs to be done at creation time.

Droog71:
> Good question.
> 
> If we only have to care about limits at creation time, then a good place to verify the param may be in the CObjectManager::CreateObject() method (there's already some kind of "validation" of object ID in it). Another good place would be CObjectFactory::CreateResource(). I think the latter makes more sense as it's closer to the PowerCells logic than the more general CreateObject() method. And the power level is actually set there.
> 
> If the limits should be kept after the object creation, then some setter would be a good place to implement such logic I guess. But let's keep things simple and assume this only needs to be done at creation time.

I see. The power level adjustment in object_manager.cpp already takes place in CObjectManager::CreateObject(), the one that passes the power level through as a parameter, on line 159. The change I made to robotmain.cpp might be redundant because directly after the power level is clamped the CreateObject() function is called like so: "CObject* obj = m_objMan->CreateObject(params);" So, it looks like I am clamping the value twice. Once before the function is called, and again inside the function itself. Either I overlooked this when I initially made this change or I am missing something and there was a valid reason for it. If not, the change to robotmain.cpp could probably be discarded and the clamp to power level will only appear in object_manager.cpp.

MrSimbax:
Ah, you're right! I was too tired to notice I guess.

So discard the change in robotmain then. Whether to move the remaining code to CreateResource() or not is up to you, I'll be fine with either.

Droog71:
> Ah, you're right! I was too tired to notice I guess.
> 
> So discard the change in robotmain then. Whether to move the remaining code to CreateResource() or not is up to you, I'll be fine with either.

I tested the game without the code in robotmain and it apparently is not redundant and is required to prevent creating overcharged powercells in scene.txt files. That must be why I included it originally. I will see if i can create a function to clamp the power level somwhere that is called from both robotmain and object_manager.

MrSimbax:
Ah, the robotmain is invoking the overloaded version which only has `params` as argument. That's why. Put the code there or in CreateResource() and it should work.

Droog71:
> Ah, the robotmain is invoking the overloaded version which only has `params` as argument. That's why. Put the code there or in CreateResource() and it should work.

Ok, I removed the code from robotmain.cpp and created a function called ClampPower in object_manager.h which is now used in the CreateObject methods in object_manager.cpp.

src/object/object_manager.cpp
```diff
@@ -162,16 +164,19 @@ CObject* CObjectManager::CreateObject(Math::Vector pos, float angle, ObjectType
     params.pos = pos;
     params.angle = angle;
     params.type = type;
-    
+    params.power = ClampPower(type,power);
```
MrSimbax, line 167: It's not needed here. You're now definitely clamping twice.

src/object/object_manager.cpp
```diff
@@ -162,16 +164,19 @@ CObject* CObjectManager::CreateObject(Math::Vector pos, float angle, ObjectType
...
+    return CreateObject(params);
```
MrSimbax, line 168: Here's where the call to the "real" `CreateObject()` function happens where the clamping should actually happen. This method here is only a convenience thing so you don't have to create the ObjectCreateParams yourself. There shouldn't be any logic happening here beyond wrapping the arguments in an ObjectCreateParams object.

Droog71: Ah, I see it clearly now "return CreateObject(params);". Should have paid more attention to that last line. Thanks!

src/object/object_manager.h
```diff
@@ -179,7 +179,10 @@ class CObjectManager : public CSingleton<CObjectManager>
 
     //! Checks if any of team's objects exist
     bool TeamExists(int team);
-
+    
+    //!Prevents creation of overcharged power cells
+    float ClampPower(ObjectType type, float power);
```
MrSimbax, line 184: This function should be private.

src/object/object_manager.cpp
```diff
@@ -128,7 +128,7 @@ CObject* CObjectManager::GetObjectByRank(unsigned int id)
 }
 
 CObject* CObjectManager::CreateObject(ObjectCreateParams params)
-{
+{
```
MrSimbax, line 131: Whitespace at end of line

src/level/robotmain.cpp
```diff
@@ -3405,7 +3405,7 @@ void CRobotMain::CreateScene(bool soluce, bool fixScene, bool resetObject)
                 m_ui->GetLoadingScreen()->SetProgress(0.25f+objectProgress*0.75f, RT_LOADING_OBJECTS, details);
 
                 try
-                {
+                {
```
MrSimbax, line 3408: Whitespace at end of line.

src/object/object_manager.cpp
```diff
@@ -141,7 +141,9 @@ CObject* CObjectManager::CreateObject(ObjectCreateParams params)
             m_nextId = params.id + 1;
         }
     }
-
+
```
MrSimbax, line 144: Linter complains about whitespace here too but I don't see anything huh. Maybe \r\n instead of \n?

src/object/object_manager.cpp
```diff
@@ -141,7 +141,9 @@ CObject* CObjectManager::CreateObject(ObjectCreateParams params)
             m_nextId = params.id + 1;
         }
     }
-
+    
+    params.power = ClampPower(params.type,params.power);
+
```
MrSimbax, line 146: And here another whitespace issue.

src/object/object_manager.h
```diff
@@ -179,7 +179,10 @@ class CObjectManager : public CSingleton<CObjectManager>
...
+
```
MrSimbax, line 185: Also another suspicious whitespace linter complaining here.

src/object/object_manager.cpp
Droog71: Actually, with this removed, bots using produce() always produce power cells with zero energy, so I had to add it back.

MrSimbax: Was that really the only way to fix this? Did you try debugging why produce() did that?

Droog71: When power cells are created from scene files in robotmain.cpp, it uses CreateObject(params) and the power is clamped by the ClampPower method. That was working. 

In scriptfunc.cpp, produce() uses CreateObject(pos, angle, type, power). You pointed out that CreateObject(params) within this method, so it seems like the value should be clamped just like it is with scene files, but the cells are always produced with zero power.

Hmmmm in the middle of typing this I think I figured it out. Missing params.power = power in the CreateObject(pos, angle, type, power) method. I never put this back after deleting the call to ClampPower.

Pushing fix now. Sorry about that!

src/object/object_manager.h
```diff
@@ -179,7 +179,7 @@ class CObjectManager : public CSingleton<CObjectManager>
 
     //! Checks if any of team's objects exist
     bool TeamExists(int team);
-
+
```
melex750, line 182: > src/object/object_manager.h:182: [whitespace] Whitespace at end of line

MrSimbax:
Linter still somehow complains about
```
src/object/object_manager.h:182: [whitespace] Whitespace at end of line
```

Droog71:
> Linter still somehow complains about
> 
> ```
> src/object/object_manager.h:182: [whitespace] Whitespace at end of line
> ```

I think I found a way to check for trailing white spaces so I can look through things before I submit pull requests, but i'm not seeing anything here: https://imgur.com/a/9Q1yetx

src/object/object_manager.h
```diff
@@ -303,6 +303,7 @@ class CObjectManager : public CSingleton<CObjectManager>
     //@}
 
 private:
+    float ClampPower(ObjectType type, float power); //!Prevents creation of overcharged power cells
```
krzys-h, line 306: This comment applies to CleanRemovedObjectsIfNeeded, not ClampPower. Either move it to the line above, or replace `//!` with `//<` (the first variant would be preferred in most cases). There should also be a space between `!` and the text.

See https://www.doxygen.nl/manual/docblocks.html

krzys-h:
The linter succeeds on the "push" trigger inside your repo:
https://github.com/Droog71/colobot/runs/885368676
Whaaaaat? The only reasonable explanation would be that it occurs only when merged into the target branch, but the target branch doesn't have the whitespace error either. Really weird.

krzys-h:
It looks like the linter ran on old version of the branch
```
/usr/bin/git log -1
commit 00ff9bfc2b9d4a3ce389c28c0f85163ec4b23a39
Author: Droog71 <71droog71@gmail.com>
Date:   Sat Jul 18 10:59:55 2020 -0400

    Merge bea97eb52518cf1371f8e4511f67a24f390d5481 into 0a3cdd342bf786d2798d0225e8397a7f3b45adcb
```
Seems like GitHub Actions bug to me? (it did the checkout before GitHun generated the merge commit? I remember our old Jenkins setup waited for that sometimes). Either way, your last commit fixed the whitespace error.

PR #1230, opened 2018-12-31
Fix for issue #1221

Droog71:
Fixes issue #1221: Crashes caused by produce() attempting to create objects with a type value that doesn't exist in object_type.h.

PR #1335, opened 2020-07-23
Allow producing objects with no keyword in CBOT

MrSimbax:
After the changes from https://github.com/colobot/colobot/pull/1230 it's no longer possible to create objects like Robbie or Visit.


immibis:
OBJECT_MAX isn't a valid ID

MrSimbax:
It is, the function checks if `static_cast<ObjectType>(someInt)` is a safe operation and it is for OBJECT_MAX too. Maybe I should choose a better name or write a documentation about this.

PR #1345, opened 2020-08-15
Fix abs() only working on floats

melex750:
[#993 comment](https://github.com/colobot/colobot/issues/993#issuecomment-658684070)


melex750:
Viewing the [merge commit](https://github.com/colobot/colobot/commit/dbc13bad2032850cc8a6c3bfb387af4d01a27460) shows this error: (It doesn't seem to show anywhere in the pull request)

abs_error

There's possibly a need to ```#include <stdlib.h>``` for the C++11 version,
or maybe using ```labs()``` or ```std::abs``` is the way to go here.

PR #1357, opened 2020-09-20
Fix crashes in SatCom on some \button characters

### LINKED ISSUE #1356: Game crash when loading SatCom
```[TRACE]: System event EVENT_KEY_UP:
[TRACE]:  virt    = false
[TRACE]:  key     = 1073741882
[TRACE]:  rTime = 0.000000
[TRACE]:  kmodState = 1000
[TRACE]:  mousePos = 0.000000, 0.000000
[TRACE]:  mouseButtonsState = 00
[TRACE]:  customParam = 0
[DEBUG]: Deactivated pause mode - engine|hide_shortcuts
[TRACE]: Reset key states
[DEBUG]: Activated pause mode - engine|hide_shortcuts|mute_sound|camera
[TRACE]: Opening SDL memory wrapper for file '/fonts/dvu_sans.ttf'
[DEBUG]: Loaded font file /fonts/dvu_sans.ttf (font size = 19.2)

Unhandled exception occurred!
==============================
Type: std::invalid_argument
Message: Unexpected UTF-8 continuation byte
==============================

This is usually caused by a bug. Please report this on http://github.com/colobot/colobot/issues
including information on what you were doing before this happened and all the information below.
==============================
You seem to be running a custom compilation of version git-dev~rb17a4391, but please verify that.

The game was in phase PHASE_SIMUL (ID=8)
Last started level was: category=missions chap=1 rank=2
==============================

Sorry for inconvenience!
[Thread 3776.0x2660 exited with code 1]
[Thread 3776.0x27d8 exited with code 1]
```
There is no backtrace available, It happens on any language except English.

Edit: Long log with steps between last log line and error here: https://pastebin.com/dNGif37R
Edit_2: It happened during mission 1.2, but it doesn't happen on 1.1 and 1.3.
Edit_3: On other planets it also happens "randomly". 2.1 and 3.3 are affected by it, but other Moon and Tropica missions are good.
Edit_4: Blaming @MrSimbax for breaking it, git bisect pointed this commit https://github.com/colobot/colobot/commit/7449111a00c7e62a898f54d3c59dcefe6b307ead#diff-39d9ebe680a2aa2cdcb487e420d1d7de as first bad one. I've checked commit before that one https://github.com/colobot/colobot/commit/6c715ffef9f2323f7134b1252f92bac36caaf6f7 and it works.

MrSimbax:
The issue is that the \button fragments in SatCom are replaced with
invalid UTF-8 bytes, hence those bytes need to be handled as a special
case when checking for UTF-8 character length.

When the text is processed e.g. in `Gfx::CText::Justify`, there are
generally two arrays: `text` and `format`. The second one is interpreted
as a font for a given character, e.g. `text[i]` has font `format[i]`.

Now, in the loops the index for `text` is increased by the length of
the character, e.g. 1 to 4 bytes, whereas index for `format` is
only incremented. It seems there's an assumption that `format` treats
multibyte characters as one byte, which doesn't seem to be true by
looking at the process of filling up the `format` array. This can result
in using wrong font e.g. FONT_SATCOM for the \button character which
should be FONT_BUTTON, hence the `StrUtils::Utf8CharSizeAt` method
complains about unexpected continuation byte via exception, causing
the crash.

Incrementing the index by the UTF-8 length seems to have fixed the
issue. However, I am not sure if I haven't broken anything else by this
and if I analyzed the intended behaviour correctly.
Gfx::CText needs a major refactor IMHO.

Fixes #1356 hopefully.

### DIFF
--- a/src/graphics/engine/text.cpp
+++ b/src/graphics/engine/text.cpp
@@ -459,7 +459,7 @@ float CText::GetStringWidth(const std::string &text,
         width += GetCharWidth(ch, font, size, width);
 
         index += len;
-        fmtIndex++;
+        fmtIndex += len;
     }
 
     return width;
@@ -607,7 +607,7 @@ int CText::Justify(const std::string &text, std::vector<FontMetaChar>::iterator 
         }
 
         index += len;
-        fmtIndex++;
+        fmtIndex += len;
     }
 
     return index;
@@ -685,7 +685,7 @@ int CText::Detect(const std::string &text, std::vector<FontMetaChar>::iterator f
 
         pos += width;
         index += len;
-        fmtIndex++;
+        fmtIndex += len;
     }
 
     return index;


tomangelo2 (approved):
Works now

PR #1371, opened 2021-01-03
Fix MSVC building issues by correcting CMakeLists.txt

jaladreips:
This change should probably be followed by a change to install instructions for MSVC


MrSimbax (approved):
Sorry for making it wait so long. After I assigned myself to this, I got some unrelated compilation problems, got discouraged, went to sleep and then forgot about the review.

PR #1391, opened 2021-02-22
Fix uploading linter results from forks to GitHub

krzys-h:
This should make linter results appear correctly on pull requests from forks. This used to be broken because GitHub doesn't trust code from forks so it doesn't give them write access to the check status API.

Now, the job running on the fork (with no write access) will create an artifact with linter results, and a job running on the main repo (with write access) will pick it up and upload it to the API. This approach is based on a blog post from GitHub themselves: https://securitylab.github.com/research/github-actions-preventing-pwn-requests

As a bonus, I updated the check pull request target job to automatically send a comment explaining the reason for failure:
image

### DIFF
--- a/.github/workflows/verify-pr-target.yml
+++ b/.github/workflows/verify-pr-target.yml
@@ -1,14 +1,21 @@
 name: Verify pull request target
 
-on: [pull_request]
+on: [pull_request_target]
 
 jobs:
   check_pr_target:
     runs-on: ubuntu-latest
     steps:
-    - name: Wrong pull request target
-      run: echo "This pull request targets the master branch. Please edit the pull request to target dev." && exit 1
+    - name: Send comment if wrong pull request target
       if: github.base_ref == 'master'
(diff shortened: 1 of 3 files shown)


.github/workflows/verify-pr-target.yml
```diff
@@ -1,14 +1,21 @@
...
 
-on: [pull_request]
+on: [pull_request_target]
 
 jobs:
   check_pr_target:
     runs-on: ubuntu-latest
     steps:
+    - name: Send comment if wrong pull request target
+      if: github.base_ref == 'master'
```
MrSimbax, line 10: We'll have to remember to change this after reorganizing and renaming the branches.

MrSimbax (approved):
Nice work, as always.

A side note, shouldn't there be similar changes in the data repo?

krzys-h:
> A side note, shouldn't there be similar changes in the data repo?

For the dev/master comment, good point, I totally forgot because I was focused on the linter which is not present in the data repo

PR #1406, opened 2021-03-21
Add Missions+ gamemode, add new CBOT categories and functions, update Sniffers

Emxx52:


src/level/robotmain.cpp
```diff
@@ -3784,6 +3785,12 @@ void CRobotMain::CreateScene(bool soluce, bool fixScene, bool resetObject)
             m_build |= BUILD_FLAG;
         }
 
+        if (m_levelCategory == LevelCategory::GamePlus && !m_ui->GetPlusResearch() && !resetObject)  // new game plus?
```
MrSimbax, line 3788: Ugly comment! Move the condition into a method or free function like "bool isNewGamePlus()".

src/common/event.h
```diff
@@ -481,7 +486,7 @@ enum EventType
     EVENT_OBJECT_BNUCLEAR   = 1060,
     EVENT_OBJECT_BPARA      = 1061,
     EVENT_OBJECT_BINFO      = 1062,
-    EVENT_OBJECT_BDESTROYER = 1063,
+    EVENT_OBJECT_BSAFE      = 1063,
```
MrSimbax, line 489: A little shame that we can't build Destroyer now but it's an issue to fix after UI refactoring.

src/object/interface/trace_drawing_object.h
```diff
@@ -32,24 +32,25 @@ enum class TraceColor
...
+    Pink = 9,
+    Purple = 10,
+    Orange = 11,
+    Beige = 12,
+    Brown = 13,
+    Skin = 14,
+    LightGreen = 15,
+    LightBlue = 16,
+    BlackArrow = 17,
+    RedArrow = 18,
```
MrSimbax, line 53: Why reorder?

MrSimbax (approved):
I trust it was tested thoroughly and works as intended.

PR #1407, opened 2021-03-29
Update polish translation

tomangelo2:


po/pl.po
```diff
@@ -275,10 +275,10 @@ msgid "Build a tracked sniffer"
 msgstr "Zbuduj szperacz na gąsienicach"
 
 msgid "Build a vault"
-msgstr ""
+msgstr "Zbuduj skrytkę"
```
elkoniu, line 278: I think "Zbuduj schron" may be more meaningful here.

Emxx52 (approved):

PR #1444, opened 2021-07-04
Slotted object interface

immibis:
* Delete CPoweredObject interface
* Delete CCarrierObject interface
* Create CSlottedObject interface which supports any number of "slots" in which objects can be held
* Add "pseudoslots" which are object-independent slot identifiers. The power cell slot could be slot 0 on one object type and slot 1 on a different object type but you can always use the power-cell pseudoslot to refer to it.
* Update the rest of the code to use CSlottedObject instead of CPoweredObject and CCarrierObject

This was used in the following demo where a research station has 2 slots: https://www.youtube.com/watch?v=y3ELoENnfeE


immibis:
There is also the possibility to implement this differently, e.g. by having an array of slots and getting the array through the interface

tomaszkax86 (approved):
Code reviewed and implementation tested. Seems fine. Ready for merge.

src/object/task/taskterraform.cpp
```diff
@@ -54,7 +54,7 @@ CTaskTerraform::CTaskTerraform(COldObject* object) : CForegroundTask(object)
     m_lastParticle = 0.0f;
     m_soundChannel = -1;
 
-    assert(m_object->Implements(ObjectInterfaceType::Powered));
+    assert(m_object->GetNumSlots() == 1);
```
lolbot-iichan, line 57: Should it be `assert(HasPowerCellSlot(m_object))` ?

lolbot-iichan:
This looks great.

Some usage ideas:
1. Add a HeavyGrabber, able of carrying lots of objects at multiple slots.
2. Convert the Vault to a 4-slotted object so that bots would know the positions of the 4 slots by IDs.

Adding water pump from the video provides even more possibilities like producing astronaut food from water+organic mater+powercells.

src/object/task/taskterraform.cpp
immibis: I think so

immibis:
>     1. Add a HeavyGrabber, able of carrying lots of objects at multiple slots.

If there was a HeavyGrabber, I can imagine it placing objects on a turntable on its back, and picking up the object in the currently selected turntable slot.

The reason I can imagine that is because we built one in real life :)

lolbot-iichan:
> > ```
> > 1. Add a HeavyGrabber, able of carrying lots of objects at multiple slots.
> > ```
> 
> If there was a HeavyGrabber, I can imagine it placing objects on a turntable on its back, and picking up the object in the currently selected turntable slot.
> 
> The reason I can imagine that is because we built one in real life :)

Wow, that would be a hybrid of a Grabber and a Scribbler.

X64E:
I liked the video with the barrel and the pump. I think this idea can be applied to new interesting missions. And the barrel can be used for different liquids.

src/object/auto/autopowercaptor.cpp
```diff
@@ -267,34 +266,22 @@ void CAutoPowerCaptor::ChargeObject(float rTime)
             }
         }
 
-        if (obj->Implements(ObjectInterfaceType::Powered))
+        if (obj->Implements(ObjectInterfaceType::Slotted))
         {
-            CObject* power = dynamic_cast<CPoweredObject&>(*obj).GetPower();
-            if ( power != nullptr && power->Implements(ObjectInterfaceType::PowerContainer) )
+            CSlottedObject* slotted = dynamic_cast<CSlottedObject*>(obj);
+            for (int slot = slotted->GetNumSlots(); slot >= 0; slot--)
```
melex750, line 272: Assert failed when trying to recharge.
It should be ```slotted->GetNumSlots() - 1```

src/object/auto/autopowerstation.cpp
```diff
@@ -134,32 +131,13 @@ bool CAutoPowerStation::EventProcess(const Event &event)
...
-                        powerContainer->SetEnergy(energy);
-                        if ( energy < freq )  freq = energy;
-                        big -= add/4.0f;  // discharge the large battery
-                    }
-                }
-            }
-
-            if (vehicle->Implements(ObjectInterfaceType::Carrier))
+            CSlottedObject* slotted = dynamic_cast<CSlottedObject*>(vehicle);
+            for (int slot = slotted->GetNumSlots(); slot >= 0; slot--)
```
melex750, line 137: Assert failed when trying to recharge.
It should be ```slotted->GetNumSlots() - 1```

melex750 (commented):
Found a small bug after merge.

PR #1458, opened 2021-08-23
Add optional team parameter to produce() function

lolbot-iichan:
Since team can't be changed after object creation, it would be useful to have team as an optional parameter for produce() function.

This can be used for dynamic creation of armies / soccer teams / racing cars / etc.

I have tried modding UEBS to have colored armies and it works.

изображение


tomangelo2:
Wait, what?

melex750:
Maybe it's confused when the feature branch is still based on ```master```, rebase??

lolbot-iichan:
>Maybe it's confused when the feature branch is still based on master, rebase??
Should I create a new PR? It's only one file, so it should be easy.

tomangelo2 (approved):

PR #1460, opened 2021-08-28
Enchance Portico unit

lolbot-iichan:
This PR makes Portico:
* Programmable and movable
* Damageable, destroyable, shielded
* More resistant to bullets using a hack similar to OBJECT_HUMAN's
* Controllable with UI
* Animated during free movement
* Have free space to walk between tracks (by moving central crashsphere upper)
* Absolutely resistant to incoming collisions, so that enemies won't be able to push it back
* Crashing everything it collides with by causing explosions

This PR also change CPhysics::ObjectAdapt to check ALL Crash Spheres of m_object instead of using GetFirstCrashSphere().
I have checked existing bots and insects and it seems that in version 0.2 all the movable objects has exactly one crashsphere, so this change should not affect performance of existing levels.

Here is this change in action:
[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=9Rf0WUrDPhY)

### DIFF
--- a/src/object/auto/automush.cpp
+++ b/src/object/auto/automush.cpp
@@ -279,6 +279,7 @@ bool CAutoMush::SearchTarget()
              type != OBJECT_LABO     &&
              type != OBJECT_NUCLEAR  &&
              type != OBJECT_PARA     &&
+             type != OBJECT_PORTICO  &&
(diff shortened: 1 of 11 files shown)


lolbot-iichan:
Abondoning this PR in favor to the modable colobot branch.

Issue #1463, opened 2021-09-29
SatCom text is offset to background


nipsufn:
Issue can be observed on Linux and Windows builds of 0.2.0 on high resolution screens (1440p) or when font size slider is set to small.
![image](https://user-images.githubusercontent.com/17983323/135277284-db89b1e8-71b3-45df-a89c-68ff64b16dad.png)

X64E:
![CBOT](https://user-images.githubusercontent.com/95232774/144445441-0c7ff89a-33f0-439d-a1a5-7d9f5ef3e7f5.gif)

PR #1464, opened 2021-09-29
fix SatCom content offset on screens with high resolution / small font size

nipsufn:
Built from master branch:
image
Fixed:
image
I've observed no negative impact on code editor, should fix https://github.com/colobot/colobot/issues/1463


MrSimbax (approved):
Seems fine. The width of the left margin is now constant regardless of the font size, which it probably should have always been. Thanks!

PR #1483, opened 2021-12-08
Refactor CBotUT.TestSaveStateIoFunctions

MrSimbax:
Split the test into several parametrised tests for each data type.

Add more tests with different values.

Fix the ReadBinary() function in order to fix the ReadDouble() function
for negative values (the sign bit was being lost).

Edit: ignore commits from #1481


X64E:
I can Approve:

cmake/FindSDL2.cmake and data folder chages.

tomangelo2 (approved):

Issue #1423, opened 2021-05-15
Program editor shows "?" instead of Cyrillic letters


pestoffne:
Colobot v0.1.12-alpha
Ubuntu 20.04

**Steps to reproduce**
1. Switch system keyboard layout to `Russian (Russian Federation)`
2. Open in-game program editor
3. Write few letters
4. Questions marks and squares appears instead of Cyrillic letters (see picture)

![Снимок экрана от 2021-05-16 01-58-30](https://user-images.githubusercontent.com/12718711/118375618-f60eb380-b5ec-11eb-9fb0-c13528bb9183.png)

Legoveed:
So, I think the charset does not support any symbols except english letters, numbers and special symbols

pestoffne:
Program editor can display local letters, see the first code line in the screenshot.
Also if I write text file in UTF-8 placed in colobot's programs directory and then load it in program editor it displays fully correctly.
The bug is that program editor adds exactly one byte for each key pressed. But some UTF-8 symbols (like Cyrillic letters) uses 2, 3 or 4 bytes.

X64E:
I think issue topic must be reanmed to "Russian absolutley no work in code editor"!
I test 2.0 version under win10. 

No characters can be entered in Russian. I think it's because of UTF 8 encoding. At the same time, the symbols are displayed normally for ready-made files.

Z128A:
@Legoveed
> So, I think the charset does not support any symbols except english letters, numbers and special symbols

Seriuosly? I copy Russian text to make varible and code works fine. Not worked only keyboard input.

![изображение](https://user-images.githubusercontent.com/95349617/145002790-6361ed05-7e45-4c76-bd90-742d758cfdf3.png)

PR #1487, opened 2021-12-14
Fixed input of long UTF-8 symbols in editor (#1423)

pestoffne:
Some UTF-8 symbols has 2, 3 or 4 bytes. So when processing text input event we should insert all bytes from event, not just first byte.
Tested using steps from #1423 on Ubuntu 20.04.


tomangelo2:
Definitely works better. I tried inserting some Polish diacritics, current dev only produces question marks, while with your fix letters are actually passed into game. However, when trying to type `Zażółć`, after inserting `ż` it removes first two letters, leaving only `ż`. Also, if you insert more of the letters and then press backspace, the letter is "converted" to question mark. I guess while inserting letters works now like it should, there is still problem with deleting them.
Moreover, when deleting all of the letters and trying to type it again, it crashes when trying to process `Z` letter, producing "Unexpected UTF-8 continuation byte" error. In debugger there is only a SIGHUP, making it harder to provide more details.
All of that happened on Create player screen.

X64E (approved):
Good work.

Yes, there may be problems with deletion, but it's still much better than watching the question mark when printing. I am in favor of integrating this pull-request.

X64E:
To tomangelo2:

I think this is due to the fact that a UTF-8 character can have a size: 1 byte, 2 bytes. 3 bytes, 4 bytes.

1 byte 0x00000000 — 0x0000007F: 0xxxxxxx
2 byte 0x00000080 — 0x000007FF: 110xxxxx 10xxxxxx
3 byte 0x00000800 — 0x0000FFFF: 1110xxxx 10xxxxxx 10xxxxxx
4 byte 0x00010000 — 0x001FFFFF: 11110xxx 10xxxxxx 10xxxxxx 10xxxxxx

To delete chars correctly, need to scan 4 bytes before the carriage and if they have signs of a long character, need to delete them as a sequence of several bytes.

MrSimbax:
I agree, we should merge the fix either way but first I'd like to ask @pestoffne -- should we merge the PR as-is or do you plan on making more changes? If the latter, we can wait until the issue with deleted characters is also resolved.

pestoffne:
@tomangelo2 wrote: "when trying to type `Zażółć`, after inserting `ż` it removes first two letters, leaving only `ż`".
I can not reproduce that. For example, I type `ż` by pressing `compose key` then `.` then `z`. How do you type Polish diacritics?

Other bugs, like:
* backslash converts two-bytes letter to question mark and
* game crushes with "Unexpected UTF-8 continuation byte" error

reproduces before and after my fix. Severity of second increases after fix because you now has more ways to accidentally crush the game.

@MrSimbax , this week I will try to fix backslash and left/right arrows in UTF-8 text to get rid of game crashing.

pestoffne:
@tomangelo2 , I tried Polish keyboard layout. `Zażółć` is being entered correctly in the "Player name" text box.

pestoffne:
Backspace/delete and left/right arrows are fixed.

Low severity bugs remains, like:
* `Ctrl` + `<--` does not count cyrillic letters as letters;
* it is possible to insert part of long UTF-8 symbol if not enough space in CEdit;
* "..." appears too early if program name has long UTF-8 symbols;
Снимок экрана от 2021-12-17 00-12-20

X64E (approved):

X64E:
@pestoffne
   "..." appears too early if program name has long UTF-8 symbols;

Great work. I think for easy calculation how charters in programm name used. We need to skip all chars 10xxxxxx.
Something like:

int lenght_of_name=0;
  for(int i=0; i<name_lenght; i++){
     if((name[i] & 0xC0) != 0x80){
        lenght_of_name++;
     }
  }

src/common/stringutils.cpp
```diff
@@ -182,10 +182,6 @@ int StrUtils::Utf8CharSizeAt(const std::string &str, unsigned int pos)
     if((c & 0xE0) == 0xC0)
         return 2;
 
-    // Invalid char - unexpected continuation byte
```
MrSimbax, line 185: Why is this exception removed? It caught more than one bug in our UTF-8 handling IIRC so I think it should stay, otherwise bugs will hide more easily.

src/ui/controls/edit.cpp
```diff
@@ -2303,12 +2306,15 @@ void CEdit::MoveChar(int move, bool bWord, bool bSelect)
         }
         else
         {
-            m_cursor1 --;
-            if ( m_cursor1 < 0 )  m_cursor1 = 0;
+            if ( m_cursor1 > 0 )
+            {
+                m_cursor1 --;
+                while ( m_cursor1 > 0 && (m_text[m_cursor1] & 0xC0) == 0x80 )  m_cursor1 --;
```
MrSimbax, line 2312: This `m_text[m_cursor1] & 0xC0) == 0x80` condition repeats a few times, maybe extract it to a method/function with a readable name. Not knowing UTF-8, I have no idea what's happening here just by reading this code.

src/ui/controls/edit.cpp
```diff
@@ -2353,8 +2359,11 @@ void CEdit::MoveChar(int move, bool bWord, bool bSelect)
         }
         else
         {
-            m_cursor1 ++;
-            if ( m_cursor1 > m_len )  m_cursor1 = m_len;
+            if ( m_cursor1 < m_len )
+            {
+                m_cursor1 ++;
+                while ( m_cursor1 < m_len && (m_text[m_cursor1] & 0xC0) == 0x80 )  m_cursor1 ++;
```
MrSimbax, line 2365: We use C++17 now, so I'd also replace the hex values with binary ones so it's easier to see what's going on.

MrSimbax (changes requested):

src/common/stringutils.cpp
X64E: I think more true variant is:
  
    if((c & 0x80) != 0) then exception
or

    if(( c & 0b10000000) != 0b00000000) then exception.

When char have size = 1 byte it must have mask 0b0xxxxxxx if byte does not match this mask, that means an error.

src/ui/controls/edit.cpp
X64E: This is not entirely fair, because the code already contains hex values.

src/ui/controls/edit.cpp
X64E: How to name this function?

X64E: This is not entirely fair, because the code already contains hex values.

src/ui/controls/edit.cpp
X64E: And how to name this function?

src/ui/controls/edit.cpp
MrSimbax: I am not asking for changing the whole file so how is it unfair?

src/common/stringutils.cpp
MrSimbax: I remember I needed to fix something in this function, and I consulted other team members regarding it, so I'm pretty sure it's correct. If the condition needs changing then so be it, but it's still an error and so there should be an exception.

X64E:
I think we need to fix only the part with exception. And do a merge. Otherwise, we may lose the pull request due to high requirements.

src/ui/controls/edit.cpp
MrSimbax: What? I don't understand what you want to say here.

I literally see exactly the same code in 4 places just in this PR, there's probably more places in the whole file. Only the `m_text[m_cursor1]` part changes which can be easily passed as parameter. It's breaking the (quite fundamental) DRY rule.

X64E: If strictly adhere to DRY. Ok.

MrSimbax:
> I think we need to (check or fix) only the part with exception. And do a merge. Otherwise, we may lose the pull request due to high requirements.

"High requirements"? The DRY rule is one of the most commonly known good coding style rules, and I haven't seen anyone being against it. Changing literals from hex to binary is not much to ask. You want to see high requirements, try creating a merge request for a popular open source project like the Linux kernel. And I am very much ready for discussion if anyone thinks my suggestions are not needed or harmful.

I get that you probably want the changes merged as fast as possible because of Russian characters support. But you should also understand that I, as one of the current maintainers of the project, and a somehow experienced software developer, very much care about the code quality, and code review is one of the ways to ensure that the code is of as high quality as possible.

Also, realistically, it'll be quite a while before the next release anyway so I see no reason to hurry that much.

The improvements I suggested are not critical (other than keeping the exception, I really have to see some good arguments why it should be removed because it helped find and fix a lot of issues I personally debugged) so even if the author ignores them I won't block the PR from merging, but then I'd probably implement the improvements myself sooner or later, after or during merge. It'd be a lot nicer and much less time consuming if the changes were already perfect upon merge though. Especially that I don't really feel like I'm asking for much. I don't require to change the whole file, unless OP wants to of course, I only looked at the new lines of code.

> we may lose the pull request due to high requirements.

Where does this lack of trust for the author comes from? :D The author would have to go out of their way to delete these changes from GitHub's server by, I imagine, nuking their fork, or force pushing. I sincerely doubt they would do that, I haven't seen anyone doing something like that yet, and there were far more heated discussions under pull requests in the history of this project. But if you're still worried so much about losing these changes then just pull their branch locally, you'll have your own copy they have no control over. The main strength of `git`, after all, is that it's a *distributed* version control system :)

src/common/stringutils.cpp
MrSimbax: Ah, I only now saw the commit message, my bad.

> Show questions instead of crushing in case of UTF-8 errors 

I am not entirely sure that removing this exception is the right way to achieve that. It was there in the first place for a reason. If you want that behaviour then I think a better solution would be to catch this exception in places where the error is ok to be ignored which I assume is not everywhere where this function is used.

I'd rather let the game crash with the exception than let some subtle bugs hide away by removing this exception. See relevant PR https://github.com/colobot/colobot/pull/1276

src/ui/controls/edit.cpp
MrSimbax: Regarding the name of the function, maybe `isUtf8ContinuationByte`?

```cpp
bool isUtf8ContinuationByte(char c)
{
    return (c & 0b1100'0000) == 0b1000'0000;
}
```

Now the line would look like this:

```cpp
while ( m_cursor1 > 0 && isUtf8ContinuationByte(m_text[m_cursor1]) ) m_cursor1 --;
```

This is much more readable IMHO, it is now obvious what the line does, assuming you know what is a continuation byte.

The function body also better describes itself thanks to the binary literals. Continuation byte is defined as a byte where the top two bits are `10`. So, firstly, since we only care about the top two bits, we do `c & 0b1100'0000` to zero the irrelevant bits, and then check if the two bits are indeed `10`.

src/common/stringutils.cpp
MrSimbax: I recalled what the function does and now I am definitely against removing these lines of code.

> I think more true variant is:

Your condition is incorrect. The condition checks whether `c` is the continuation byte which is defined as a byte which starts with `10`. If `c` starts with bits `11` then `c & 0x80` would result in `11000000` which is definitely not `0` so your condition is true but `c` is *not* a continuation byte in this case.

The function is correct. It assumes it is given the first byte of a UTF-8 character and returns the size in bytes of the UTF-8 character. The first few bits of the first byte define how long is the character. All the other bytes belonging to the UTF-8 character must be continuation bytes. See https://stackoverflow.com/a/9356203

The exception is thrown in case the first byte is a continuation byte because it is simply wrong according to UTF-8 specification. So it means there's something wrong with the whole input text (it contains invalid UTF-8 characters) or the function was given a wrong position. In both cases, the fault lies higher in the stack and so the decision of what to do with this error should be decided outside this function, somewhere higher in the stack. And this can be done by catching the exception.

pestoffne: > The exception is thrown in case the first byte is a continuation byte because it is simply wrong according to UTF-8 specification.

I agree.

> It caught more than one bug in our UTF-8 handling IIRC so I think it should stay, otherwise bugs will hide more easily.

I think the game should not crush in UTF-8 errors. Because
* players does not like when game is crushing;
* we can not prevent UTF-8 errors -- file can be broke outside the game, inconsistent sequence can be pasted from clipboard and now there is knows bugs.

> The game crashes because nobody catches the exception so if we want to prevent the crash then it should be caught in appropriate places where the function is used directly or indirectly.

Today I will try to implement it.

src/ui/controls/edit.cpp
pestoffne: isUtf8ContinuationByte function is good idea. I will add it.

src/ui/controls/edit.cpp
pestoffne: Master branch uses C++11. I did not notices that dev uses C++17. Binary constants definitely look more clear in that case.

src/common/stringutils.cpp
X64E: Your condition is incorrect. The original condition checks whether c is the continuation byte which is defined as a  byte
     which starts with 10. If c starts with bits 11 then c & 0x80 would result in 11000000 which is definitely not 0 so your
     condition is true but c is not a continuation byte in this case.

If c starts with bits 11 It would be cathed by previous if condition.

    if((c & 0xE0) == 0xC0)
        return 2;

There are many ways to solve the same problem. Which one is correct is a matter of taste.

MrSimbax: > If c starts with bits 11 It would be cathed by previous if condition.

Ok, so I went too far as to calling your suggestion incorrect because in our case I guess it would work. However, it relies on the earlier conditions. On the other hand, the current condition is more general and can (should) be easily replaced with a call to `isUtf8ContinuationByte(c)`. The whole if statement could even be moved to the beginning of the function and it would still work, which would not be possible with your suggestion.

> There are many ways to solve the same problem.

Yes.

> Which one is correct is a matter of taste.

No.

You mixed up correctness with form. A solution to a problem is either correct or not, and it only depends on the problem definition. It is not up to taste, correctness can be objectively verified by technical/scientific analysis, or even automatically with, for example, unit tests. It is the form of the solution that is a matter of taste, not correctness.

By the way, I've just realized this code probably has no unit tests. No wonder it causes so many problems.

MrSimbax:
I am not even going to bother answering this ridiculous Unicode suggestion.

> Can already merge what is there?

Only if the exception is restored. I am not letting that one slide. Unless another team member has another opinion on this matter, in such a case I shall shut up.

> And don't wait for the perfect S.O.L.I.D. code?

What in the lord's name is the problem here? I only gave a simple innocent code review with suggestions for simple improvements for an already nice pull request and gave the red light only because of the exception removal.

Demanding merge and immediate release isn't going to speed anything up and isn't helping anybody. The author can take as long as they like, as can we, because we're not on a payroll.

pestoffne:
Thanks for waiting.

@MrSimbax , all your suggestions are taken into account in the code.

@X64E , using Unicode instead of UTF-8 in editor is bad idea.

> ... Unicode, which always has a character equal to 2 bytes.

It is wrong. Latest Unicode version 14 has 144'697 characters (see https://www.unicode.org/versions/Unicode14.0.0/). 2-byte variable has 65'536 different values. So, Unicode 14 can't always use 2-byte characters. There is UTF-32 which always use 4-byte characters.

UTF-32 compared with UTF-8:
`+`  Fast jump to specified index
`-`  Each character takes up more memory
`-`  BOM is needed to distinguish little from big endian

Detailed and simple explanation here https://youtu.be/4MFcmreAUhs?t=350

src/ui/controls/edit.cpp
```diff
@@ -2303,12 +2307,18 @@ void CEdit::MoveChar(int move, bool bWord, bool bSelect)
...
         {
-            m_cursor1 --;
-            if ( m_cursor1 < 0 )  m_cursor1 = 0;
+            if ( m_cursor1 > 0 )
+            {
+                m_cursor1 --;
+                while ( m_cursor1 > 0 && StrUtils::isUtf8ContinuationByte(m_text[m_cursor1]) )
+                {
+                    m_cursor1 --;
+                }
```
MrSimbax, line 2316: Looks like a do...while loop would be a better choice here and below.

MrSimbax (approved):

CookiePLMonster:
The benefits of using UTF-32 are far less than the drawbacks. The only real drawback of UTF-8 is no random access to characters, the "complexity" required to deal with multibyte characters properly is not really a problem nor it is exclusive to UTF-8 - this should technically also be done even when operating on local codepages.

src/common/stringutils.cpp
```diff
@@ -172,21 +178,24 @@ std::wstring StrUtils::Utf8StringToUnicode(const std::string &str)
 int StrUtils::Utf8CharSizeAt(const std::string &str, unsigned int pos)
 {
     if (pos >= str.size())
-        return 0;
+        throw std::out_of_range("Index is greater than size");
```
CookiePLMonster, line 181: If that exception is needed, you might as well use `str.at(pos)` instead of `[]` and get it given to you without a prior check: https://en.cppreference.com/w/cpp/string/basic_string/at

MrSimbax:
I will merge it as-is, any additional fixes can be made in another pull request.

PR #1495, opened 2022-01-23
Add workaround for numpad home/end keys etc

rasmusgo:
SDL is not reporting numpad keys correctly when numlock is off. This is a workaround to make numpad home, end etc usable in the in-game editor. It seem to be a know issue in SDL: https://github.com/libsdl-org/SDL/issues/1766


tomaszkax86 (approved):
The code looks good and works as intended. I can finally use Home and End on numpad.

tomangelo2 (approved):
I've just checked it on my laptop, it works as well there.

Issue #812, opened 2016-08-06
Crash auto recovery seems to have removed the history of my missions.


VictorDenisov:
After several crashes I might have agreed to save after a crash and then recovered when I was prompted at the startup. I successfully completed the mission, but then I discovered that the history of the missions I passed is gone.

VictorDenisov:
commit: 407d8554669951be794788d37b268fb54bb67aec1

lisa50469:
I had this happen today.  9/11/2016.  Linux version.  Found out the 'break' key doesn't exist on my keyboard, so I could not use the cheat codes to do 'allmissions'

tomangelo2:
Console in Gold is available under ` key.

Ceeee:
This happened to me too, exactly as described. Game crash in 0.1.8 (Win) in the first level with spiders shooting at them. Then recovery, and with the solved level I'm back to Level 1 now.

attila123:
This also happened to me after the crash I described in #1027 .

attila123:
The correct cheat is 'allmission' (not 'allmissions').

PR #1498, opened 2022-02-02
Load game from command line

rasmusgo:
Loading a saved game from the command line makes it faster to iterate when developing.

Here is an example command line:

    ./build/colobot -loadsave Rasmus/Crystallium4withflags

### DIFF
--- a/src/level/robotmain.cpp
+++ b/src/level/robotmain.cpp
@@ -5839,6 +5839,17 @@ void CRobotMain::QuickLoad()
     m_playerProfile->LoadScene(dir);
 }
 
+void CRobotMain::LoadGameFromDirName(const std::string& gameDir)
+{
+    std::string dir = m_playerProfile->GetSaveFile(gameDir);
+    if(!CResourceManager::Exists(dir))
+    {
+        GetLogger()->Error("Save slot not found\n");
+        return;
+    }
+    m_playerProfile->LoadScene(dir);
+}
+
 void CRobotMain::SetExitAfterMission(bool exit)
 {
     m_exitAfterMission = exit;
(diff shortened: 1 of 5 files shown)


tomangelo2:
I think if this shouldn't be renamed to `loadsave`, or `loadsavegame`. We want to run other Epsitec games in future on this engine, so `loadgame` might be a little bit confusing in that case.

rasmusgo:
It's unfortunate that `save` is both a verb and a noun. `loadsave` and `loadsavegame` sounds a bit like the game is loaded and saved. The folder on disk is called `savegame` on disk though 🤷 

How big is the risk of confusion with the other games? Perhaps they will use `rungame` or simply `game`?

src/level/robotmain.cpp
```diff
@@ -5839,6 +5839,17 @@ void CRobotMain::QuickLoad()
...
 
+void CRobotMain::LoadSaveFromDirName(const std::string& gameDir)
+{
+    std::string dir = m_playerProfile->GetSaveFile(gameDir);
+    if(!CResourceManager::Exists(dir))
+    {
+        GetLogger()->Error("Save slot not found\n");
+        return;
+    }
+    m_playerProfile->LoadScene(dir);
```
melex750, line 5850: I think this will have the same issue as #812.

We could add a boolean parameter to ```CPlayerProfile::LoadScene()``` to handle these two special cases.

Maybe ```LoadScene(std::string dir, bool loadFinishedFromCat = false);```

rasmusgo: Ouch. Issue confirmed...

melex750: It looks that might work, maybe that's the proper fix for #812 too.

rasmusgo: That seems like it worked, at least for me.

tomangelo2 (approved):
Looks like everything works.

Issue #336, opened 2014-10-20
Rewrite goto() algorithm


krzys-h:
This continues issue #326

rasmusgo:
The current implementation of goto() is based on [beam search](https://en.wikipedia.org/wiki/Beam_search) which is heuristic search algorithm that trades away optimality and completeness for reduced memory usage. The search space is also a bit weirdly defined because it searches combinations of straight lines and turns instead of using nodes in a regular grid. When it finds a path around obstacles, the initial path is typically very chaotic but the robot will skip large parts by taking shortcuts. The shortcuts are found by looking for the node with the highest index that the robot has a straight line towards without any obstacles.

I think we should use a regular grid and a search algorithm with guaranteed results such as [A*](https://en.wikipedia.org/wiki/A*_search_algorithm), [BFS](https://en.wikipedia.org/wiki/Breadth-first_search) or similar. I think it is important to keep the shortcutting behavior, not to skip detours but to make the path straighter by allowing finer than 45 degree increments.

The memory usage will go up with these types of algorithms. Can we afford it? The worst-case memory complexity of A* and BFS  are both linear in number of nodes, but BFS stores less information per node. The current beam search implementation is using an occupancy grid with 640x640 resolution to cover the whole map (3200/BM_DIM_STEP = 640). If we keep that resolution and use BFS with one float per node to store a distance and one integer to identify nodes in the queue we would need at most 640 * 640 * (4+4) = 3276800 bytes (=3.125 MiB). I think that would be ok.

tomangelo2:
3MiB extra doesn't sound like a huge difference, Gold itself takes about 400-500MB of memory (408MB fresh after loading a level 2-4 in my case). I'd spare some megabytes of memory in exchange for reliable pathfinding in goto().

rasmusgo:
My current favorite candidate algorithm right now is bidirectional BFS.
* Same memory footprint as regular BFS.
* No sorting of the queue.
* Typically fewer nodes explored compared to regular BFS.
* Implementation can be staged and de-risked by first making regular BFS.

rasmusgo:
[...]

PR #1499, opened 2022-02-02
Better goto

rasmusgo:
This replaces beam search with ~~breadth-first~~ _A*_ search for pathfinding in goto, see https://github.com/colobot/colobot/issues/336. ~~This is still work in progress but it seems to work quite well so far.~~ _I have developed the algorithm to a state I am happy with._ Please test it!

## Extended description
Since this is a substantial change to a critical component in the game, I think it warrants an extended description. This should help during reviewing and also serve as a future reference when debugging or developing it further. **The most fundamental change in this PR is that it searches for paths in an explicit discrete grid instead of a state space search consisting of straight sections joined by turns.** This allows the search to be exhaustive because the number of nodes are finite.

### The grid (nodes)
The previous implementation used a grid to keep track of drivable areas. This grid has **not** been changed in this PR. It is now also used for the discretization of the possible positions. All intermediate positions between start and goal are aligned to this grid. The grid resolution is one grid cell per five in-game units. This resolution is now the main limitation of finding a path in crowded areas. Increasing the resolution increases memory usage and computational costs. The memory usage is mainly driven by storing distance estimates (upper bound from cell to start position) for all positions on the grid.

### The edges
I chose to use 8-way connectivity for modelling travelling from one node (grid cell) to another. This is not as limiting as it first seems because there is a shortcutting behavior that will drive straight to nodes further along the path if there is a straight line without obstacles. The costs are represented by integers instead of floats for increased performance and numerical stability. The main reason to use integer costs is that it allows for using a different queue when searching, more on that below. I am using 7 and 5 for diagonal and non-diagonal costs respectively. This is a reasonable approximation, 7/5 ~= sqrt(2), which underestimates the diagonal cost by 1%. The limited number of directions also affect the precision of the cost estimate. It may overestimate the cost up to 7.33%. I don't think this will lead to any issues, but if it does, it can be mitigated by adding eight more edges for a finer angular resolution and possibly increasing the scale of the integers, The extra edges would act as shortcuts for two regular steps, one diagonal and one non-diagonal.
 
### The search
There are plenty of algorithms for finding paths in discrete grids. Perhaps the most basic one is [breadth-first search](https://en.wikipedia.org/wiki/Breadth-first_search) which will find one of the paths with fewest edges. I implemented that first. In order to be able to find a path to any position in a circle close to the goal, the search is performed from goal position(s) to start position. This allows for several goal positions to be used. This is used for finding paths to objects, eg. `goto(radar(TitaniumCube).position);`.

The next step is  [Dijkstra's algorithm](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm) which will find one of the paths with the lowest cost. Dijkstra's algorithm relies on using a priority queue for ordering the nodes to explore. A [binary heap](https://en.wikipedia.org/wiki/Binary_heap) is often used for this as it relatively simple, efficient and implementations are abundant. However, if the priorities are integers, a [Bucker queue](https://en.wikipedia.org/wiki/Bucket_queue) can be used instead with improved performance characteristics. This is known as Dial's algorithm and I chose to implement that variation.

The search algorithms described above expands nodes outwards in a symmetric fashion until it happens to find a path between start and goal. A heuristic can be added in order to guide the search and find a solution faster by sorting the queue by the upper bound of the cost of the full path that goes through a particular node. This is known as [A*](https://en.wikipedia.org/wiki/A*_search_algorithm) and is probably the most commonly used pathfinding algorithm in games. I have not seen it implemented with a bucket queue before but I would be surprised if I'm the first one to do it. The heuristic most commonly used with A* is the Euclidean distance ("as the crow flies") but since we are using 8-way connectivity I made it the minimal cost in that metric.

I also implemented a fallback behavior for the case of unexpectedly wide range of priorities. This code is never used in the game because the game never requests big enough `goalRadius` for it to activate. I did test it by reducing the number of buckets to make sure it works though. It does add a bit of complexity that could be avoided so I would kind of recommend myself to "kill your darlings" and replace it with an assert.

### Suggestions for future improvements
* Increase the resolution of the grid close to obstacles by using some kind of non-uniform grid.
* Look for shortcuts more often to follow the path more smoothly.
* Remove the "beamleak" behavior that moves the bot away from nearby obstacles before starting the search.

### DIFF
--- a/src/object/task/taskgoto.cpp
+++ b/src/object/task/taskgoto.cpp
@@ -222,7 +222,7 @@ bool CTaskGoto::EventProcess(const Event &event)
             if ( m_bmCargoObject->GetType() == OBJECT_BASE )  dist = 12.0f;
         }
 
-        ret = BeamSearch(pos, goal, dist);
+        ret = PathFindingSearch(pos, goal, dist);
         if ( ret == ERR_OK )
         {
             if ( m_physics->GetLand() )  m_phase = TGP_BEAMWCOLD;
@@ -346,7 +346,7 @@ bool CTaskGoto::EventProcess(const Event &event)
         {
             m_physics->SetMotorSpeedX(0.0f);  // stops the advance
             m_physics->SetMotorSpeedZ(0.0f);  // stops the rotation
-            BeamStart();  // we start all
+            PathFindingStart();  // we start all
             return true;
         }
 
@@ -790,7 +790,7 @@ Error CTaskGoto::Start(Math::Vector goal, float altitude,
             }
         }
 
-        BeamStart();
+        PathFindingStart();
 
         if ( m_bmCargoObject == nullptr )
         {
@@ -826,7 +826,7 @@ Error CTaskGoto::IsEnded()
         {
             m_physics->SetMotorSpeedX(0.0f);  // stops the advance
             m_physics->SetMotorSpeedZ(0.0f);  // stops the rotation
-            BeamInit();
+            PathFindingInit();
             m_phase = TGP_BEAMSEARCH;  // will seek the path
         }
         return ERR_CONTINUE;
@@ -888,7 +888,7 @@ Error CTaskGoto::IsEnded()
             m_physics->SetMotorSpeedX(0.0f);  // stops the advance
             m_physics->SetMotorSpeedZ(0.0f);  // stops the rotation
 
-            m_bmIndex = BeamShortcut();
+            m_bmIndex = PathFindingShortcut();
 
             if ( m_bmIndex > m_bmTotal )
             {
@@ -1630,7 +1630,7 @@ void CTaskGoto::ComputeFlyingRepulse(float &dir)
 // Among all of the following, seek if there is one allowing to go directly to the crow flies.
 // If yes, skip all the unnecessary intermediate points.
 
-int CTaskGoto::BeamShortcut()
+int CTaskGoto::PathFindingShortcut()
 {
     int     i;
 
@@ -1647,7 +1647,7 @@ int CTaskGoto::BeamShortcut()
 
 // That's the big start.
 
-void CTaskGoto::BeamStart()
+void CTaskGoto::PathFindingStart()
 {
     Math::Vector    min, max;
 
@@ -1673,14 +1673,14 @@ void CTaskGoto::BeamStart()
     {
         m_physics->SetMotorSpeedX(0.0f);  // stops the advance
         m_physics->SetMotorSpeedZ(0.0f);  // stops the rotation
-        BeamInit();
+        PathFindingInit();
         m_phase = TGP_BEAMSEARCH;  // will seek the path
     }
 }
 
-// Initialization before the first BeamSearch.
+// Initialization before the first PathFindingSearch.
 
-void CTaskGoto::BeamInit()
+void CTaskGoto::PathFindingInit()
 {
     int     i;
 
@@ -1701,8 +1701,8 @@ void CTaskGoto::BeamInit()
 // ERR_CONTINUE if not done yet
 // goalRadius: distance at which we must approach the goal
 
-Error CTaskGoto::BeamSearch(const Math::Vector &start, const Math::Vector &goal,
-                            float goalRadius)
+Error CTaskGoto::PathFindingSearch(const Math::Vector &start, const Math::Vector &goal,
+                                   float goalRadius)
 {
     m_bmStep ++;
 
--- a/src/object/task/taskgoto.cpp
+++ b/src/object/task/taskgoto.cpp
@@ -1912,18 +1912,42 @@ Error CTaskGoto::PathFindingSearch(const Math::Vector &start, const Math::Vector
                 if (btX == goalX && btY == goalY)
                 {
                     m_bmPoints[m_bmTotal] = goal;
-                    break;
                 }
-                m_bmPoints[m_bmTotal].x = (btX + 0.5f) * BM_DIM_STEP - 1600.f;
-                m_bmPoints[m_bmTotal].z = (btY + 0.5f) * BM_DIM_STEP - 1600.f;
+                else
+                {
+                    m_bmPoints[m_bmTotal].x = (btX + 0.5f) * BM_DIM_STEP - 1600.f;
+                    m_bmPoints[m_bmTotal].z = (btY + 0.5f) * BM_DIM_STEP - 1600.f;
+                }
 
                 if (bestDistance == 0)
                 {
+                    if (goalRadius > 0.0f)
+                    {
+                        // Find a more exact position by repeatedly bisecting the interval.
+                        const float r2 = goalRadius * goalRadius;
+                        Math::Vector inside = m_bmPoints[m_bmTotal] - goal;
+                        Math::Vector outside = m_bmPoints[m_bmTotal-1] - goal;
+                        Math::Vector mid = (inside + outside) * 0.5f;
+                        for (int i = 0; i < 10; ++i)
+                        {
+                            if (mid.x*mid.x + mid.z*mid.z < r2)
+                            {
+                                inside = mid;
+                            }
+                            else
+                            {
+                                outside = mid;
+                            }
+                            mid = (inside + outside) * 0.5f;
+                        }
+                        m_bmPoints[m_bmTotal] = mid + goal;
+                    }
                     break;
                 }
             }
 
-            GetLogger()->Debug("Found path to goal with %d nodes and %d cost\n", m_bmTotal + 1, totalDistance);
+            const float distanceToGoal = DistanceProjected(m_bmPoints[m_bmTotal], goal);
+            GetLogger()->Debug("Found path to goal with %d nodes and %d cost. Final distance to goal: %f\n", m_bmTotal + 1, totalDistance, distanceToGoal);
             GetLogger()->Debug("m_bmStep: %d\n", m_bmStep);
             GetLogger()->Debug("m_bfsQueueMin: %d mod %d = %d\n", m_bfsQueueMin, NUMQUEUEBUCKETS, m_bfsQueueMin % NUMQUEUEBUCKETS);
             GetLogger()->Debug("m_bfsQueueCountPushed: %d\n", m_bfsQueueCountPushed);
--- a/src/object/task/taskgoto.cpp
+++ b/src/object/task/taskgoto.cpp
@@ -1689,8 +1689,15 @@ void CTaskGoto::PathFindingInit()
         m_bmIter[i] = -1;
     }
     m_bmStep = 0;
-    m_bfsQueueBegin = 0;
-    m_bfsQueueEnd = 0;
+    for (auto& bucket : m_bfsQueue)
+    {
+        bucket.clear();
+    }
+    m_bfsQueueMin = 0;
+    m_bfsQueueCountPushed = 0;
+    m_bfsQueueCountPopped = 0;
+    m_bfsQueueCountRepeated = 0;
+    m_bfsQueueCountSkipped = 0;
 }
 
 // Calculates points and passes to go from start to goal.
@@ -1709,7 +1716,7 @@ Error CTaskGoto::PathFindingSearch(const Math::Vector &start, const Math::Vector
     // Relative postion and distance to neighbors.
     static const int dXs[8] = {-1, 0, 1, -1, 1, -1, 0, 1};
     static const int dYs[8] = {-1, -1, -1, 0, 0, 1, 1, 1};
-    // static const float dDist[8] = {M_SQRT2, 1.0f, M_SQRT2, 1.0f, 1.0f, M_SQRT2, 1.0f, M_SQRT2};
+    // These are the costs of the edges. They must be less than the number of buckets in the queue.
     static const int32_t dDist[8] = {7, 5, 7, 5, 5, 7, 5, 7};
 
     const int startX = static_cast<int>((start.x+1600.0f)/BM_DIM_STEP);
@@ -1717,7 +1724,7 @@ Error CTaskGoto::PathFindingSearch(const Math::Vector &start, const Math::Vector
     const int goalX = static_cast<int>((goal.x+1600.0f)/BM_DIM_STEP);
     const int goalY = static_cast<int>((goal.z+1600.0f)/BM_DIM_STEP);
 
-    if (m_bfsQueueEnd == 0) // New search
+    if (m_bfsQueueCountPushed == 0) // New search
     {
         if (startX == goalX && startY == goalY)
         {
@@ -1731,8 +1738,9 @@ Error CTaskGoto::PathFindingSearch(const Math::Vector &start, const Math::Vector
             goalY >= 0 && goalY < m_bmSize )
         {
             const int indexInMap = goalY * m_bmSize + goalX;
-            m_bfsDistances[indexInMap] = 0.0;
-            m_bfsQueue[m_bfsQueueEnd++] = indexInMap;
+            m_bfsDistances[indexInMap] = 0;
+            m_bfsQueue[0].push_back(indexInMap);
+            m_bfsQueueCountPushed += 1;
             BitmapSetDot(1, goalX, goalY); // Mark as enqueued
         }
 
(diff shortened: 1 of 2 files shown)


melex750:
It does seem to work well.
I tested with this userlevel [MazesOnMarsV02.zip](https://github.com/colobot/colobot/files/7993711/MazesOnMarsV02.zip)

tomangelo2:
[check_levels.zip](https://github.com/colobot/colobot/files/8056809/check_levels.zip)
This mod contains modified mission 2.1, where both astronaut and bots are preprogrammed to do the task automatically. Goto() struggles with pathfinding here, sometimes it does have problems navigating astronaut from rocket to surface (and vice versa), sometimes it just gets completely stuck trying to get a piece of titaniumore, or just keeps walking into titanium cube. If you could try to look if it's possible to iron it out.

Also, your algorithm already does better job than original beam search, which tends to massively hangs whole game trying to calculate path to cells after building BotFactory, generating really confusing path. That issue might not be caused by bug in pathfinding, maybe it's another piece of code that's cranky, so if you cannot fix that, or it's out of scope - no problem, yours implementation is still an improvement over original one.

(I know that script isn't really finished, or it could be improved, its point is just to provide few situations where goto() still struggles).

rasmusgo:
> [check_levels.zip](https://github.com/colobot/colobot/files/8056809/check_levels.zip) This mod contains modified mission 2.1, where both astronaut and bots are preprogrammed to do the task automatically. Goto() struggles with pathfinding here, sometimes it does have problems navigating astronaut from rocket to surface (and vice versa), sometimes it just gets completely stuck trying to get a piece of titaniumore, or just keeps walking into titanium cube. If you could try to look if it's possible to iron it out.

Yikes! It looks like it finds reasonable paths but has a hard time following them. The robot starts by picking up a battery when I start the level. The robot drives to a spot where both the batteries and the titanium cubes are within reach but ends the drive facing the batteries instead of the titanium cubes. It should turn to face the titanium at the end of the drive. Is this problem new or old? 

> Also, your algorithm already does better job than original beam search, which tends to massively hangs whole game trying to calculate path to cells after building BotFactory, generating really confusing path. That issue might not be caused by bug in pathfinding, maybe it's another piece of code that's cranky, so if you cannot fix that, or it's out of scope - no problem, yours implementation is still an improvement over original one.

I agree with this analysis and I think it makes sense to limit the scope of this PR to pathfinding.

src/object/task/taskgoto.cpp
```diff
@@ -1698,199 +1727,326 @@ void CTaskGoto::BeamInit()
...
+                }
+                m_bmPoints[m_bmTotal].x = (btX + 0.5f) * BM_DIM_STEP - 1600.f;
+                m_bmPoints[m_bmTotal].z = (btY + 0.5f) * BM_DIM_STEP - 1600.f;
 
-// Is a right "start-goal". Calculates the point located at the distance "step"
-// from the point "start" and an angle "angle" with the right.
+                if (bestDistance == 0)
+                {
+                    break;
+                }
```
rasmusgo, line 1923: This will set an exact goal position if the `goalRadius` is zero but a rounded position when it isn't. I think this can cause problem when trying to pick up objects. It should be improved before merging.

src/object/task/taskgoto.cpp
```diff
@@ -49,9 +49,10 @@ const float FLY_DEF_HEIGHT  = 50.0f;    // default flying height
 // Settings that define goto() accuracy:
 const float BM_DIM_STEP     = 5.0f;     // Size of one pixel on the bitmap. Setting 5 means that 5x5 square (in game units) will be represented by 1 px on the bitmap. Decreasing this value will make a bigger bitmap, and may increase accuracy. TODO: Check how it actually impacts goto() accuracy
 const float BEAM_ACCURACY   = 5.0f;    // higher value = more accurate, but slower
-const float SAFETY_MARGIN   = 0.5f;     // Smallest distance between two objects. Smaller = less "no route to destination", but higher probability of collisions between objects.
+const float SAFETY_MARGIN   = 1.5f;     // Smallest distance between two objects. Smaller = less "no route to destination", but higher probability of collisions between objects.
 // Changing SAFETY_MARGIN (old value was 4.0f) seems to have fixed many issues with goto(). TODO: maybe we could make it even smaller? Did changing it introduce any new bugs?
 
+const int NB_ITER = 2000;  // Maximum number of iterations you have the right to make before temporarily interrupt in order not to lower the framerate.
```
melex750, line 55: I think for now NB_ITER should be around 200 as before, it will stop many instances of goto from using too much cpu time each frame. A lower number also lets you watch/pause the algorithm when using the debug feature.

src/object/task/taskgoto.cpp
```diff
@@ -126,7 +127,7 @@ bool CTaskGoto::EventProcess(const Event &event)
                         if (a || b)
```
melex750, line 127: The debug feature gets a boost in performance when we draw only the active area of the bitmap.
Like this a few lines above:
```diff
-                for (int x = 0; x < m_bmSize; x++)
+                for (int x = m_bmMinX; x <= m_bmMaxX; x++)
                 {
-                    for (int y = 0; y < m_bmSize; y++)
+                    for (int y = m_bmMinY; y <= m_bmMaxY; y++)
```

src/object/task/taskgoto.cpp
melex750: When goalRadius is not 0.0f the last point (m_bmPoints[m_bmTotal]) should be that distance from the position given by parameter 'goal'. This is currently expected during the final phases of the movement part of goto.

melex750 (commented):
This is a very good start, I think it's ok to merge this.

I've spent a few days testing and experimenting with these changes.
The algorithm works very good, it's compact and well organized.

I should confess that I have been experimenting with goto() for a long time.
I'm ready to add some things that will make this work a lot better than it already does.

In particular it will help to increase bitmap resolution by 100%.
The better bitmap makes it easy to add a simple path smoothing algorithm,
I have already made one that works really good with these changes.

The following code reduces rounding errors and distributes them
evenly across the bitmap, and as a side effect it visually aligns
the debug image with in game objects.
```c++
const float BM_DIM_STEP     = 2.5f;     // Size of one pixel on the bitmap...
const float BM_PIXEL_OFFSET = -2.5f;    // Alignment offset for x and y values of pixels in bit table.

inline int ToPixel(float n)
{
    return static_cast<int>((n+1600.0f+BM_PIXEL_OFFSET)/BM_DIM_STEP);
}

inline float FromPixel(int n)
{
    return static_cast<float>(n)*BM_DIM_STEP-(1600.0f+BM_PIXEL_OFFSET);
}

so... BitmapTestDot(0, ToPixel(pos.x), ToPixel(pos.z));

```

tomangelo2:
> Yikes! It looks like it finds reasonable paths but has a hard time following them. The robot starts by picking up a battery when I start the level. The robot drives to a spot where both the batteries and the titanium cubes are within reach but ends the drive facing the batteries instead of the titanium cubes. It should turn to face the titanium at the end of the drive. Is this problem new or old?

Well, this script was created just now, the only comparison I've did for this is current dev branch. Also, as I already mentioned on Discord, I was playing with movie sequences off, so I didn't take into account to wait with script execution before the cutscene ends, which is probably why robot skips completely delivering titanium to flat space and goes for battery cell for research center, then waits indefinitely.

Also for pathfinding, straight on the start the astronaut is coded to turn right and move by 20 units. Without it, Astronaut struggles to find a path to the titanium, when I tried previous pathfinding algorithm it just hanged out.

I've recorded some footage of how it looks in my case. There are few highlights I'd like to point out
* 01:10 - Astronaut cannot find a way to the blue flag. After few failed attempts, however, he finds a way.
* 01:35-01:48 - Astronaut cannot find a way to the batteries, unless I move him right to the edge of the ship. 
* 02:44-03:45 - Astronaut struggles with going off ship after grabbing cell. This, however, is a move() function. Right after move() function ends at 03:45, goto() takes control, walks again into that cube, and in 03:54 it completely hangs, not returning any value. Astronaut from that point doesn't do anything, apart from sliding from the ship. The move() function here is because goto() just hanged out immediately on previous pathfinding iteration, so I temporarily deleted one cube and made Astronaut go there with turn/move. Interestingly enough, now if I use goto() without move(), it finds path from there, only throws few errors.
* 03:01 - WingedGrabber cannot find path to flag closest to the buildings. Instead, goto(SpaceShip) is triggered. Once WingedGrabber finds itself on the ship, a path to flag is found. It happens for both ore pieces, probably for next ones too.
* 06:09 - WingedGrabber finds a path to TitaniumOre, but gets stuck trying to approach it. Eventually, it runs out of energy, trying to get to its position.

https://www.dropbox.com/s/857trk02q7wvfxe/2022-02-14%2019-40-39-1.mp4?dl=0

You may see that most issues there are with astronaut using goto(), then at less degree with WingedGrabber, while WheeledGrabber seems to work without any problem. As pathfinding is same for everyone, the issue might be related to how goto() handles different objects. So that in fact would appear to be another issue.

src/object/task/taskgoto.cpp
rasmusgo: Yes, but only approximately because it is aligned to the grid. I think the grid cells are big enough to cause issues. The `goalRadius` is 8 for picking up objects and the cell size is 5. Too close for comfort. It would make sense to adjust the distance to the goal in a similar way as it was done before with beam search.

melex750: Yes, that basically what I was trying to say.

src/object/task/taskgoto.cpp
rasmusgo: Sounds reasonable. Have you measured the time spent per frame with the new vs the old version? I haven't done anything thorough but I suspect that the new version costs less per iteration but reaches a shorter distance.

rasmusgo:
> This is a very good start, I think it's ok to merge this.
> 
> I've spent a few days testing and experimenting with these changes. The algorithm works very good, it's compact and well organized.
> 
> I should confess that I have been experimenting with goto() for a long time. I'm ready to add some things that will make this work a lot better than it already does.
> 
> In particular it will help to increase bitmap resolution by 100%. The better bitmap makes it easy to add a simple path smoothing algorithm, I have already made one that works really good with these changes.

I'm very happy to hear this! 😃

Increasing the bitmap resolution makes a lot of sense to me and I would love to hear more about the path smoothing algorithm you have made!

> The following code reduces rounding errors and distributes them evenly across the bitmap, and as a side effect it visually aligns the debug image with in game objects.
> 
> ```c++
> const float BM_DIM_STEP     = 2.5f;     // Size of one pixel on the bitmap...
> const float BM_PIXEL_OFFSET = -2.5f;    // Alignment offset for x and y values of pixels in bit table.
> 
> inline int ToPixel(float n)
> {
>     return static_cast<int>((n+1600.0f+BM_PIXEL_OFFSET)/BM_DIM_STEP);
> }
> 
> inline float FromPixel(int n)
> {
>     return static_cast<float>(n)*BM_DIM_STEP-(1600.0f+BM_PIXEL_OFFSET);
> }
> 
> so... BitmapTestDot(0, ToPixel(pos.x), ToPixel(pos.z));
> ```

I don't really get what this offset is meant do do. At first I thought you were rounding to the closest integer but that is not what is going on since the offset is the same size as the cell size and negative. Can you tell me more about it?

> Maybe I can open a PR to propose some other improvements.

Please do!

melex750:
The fact that the cell size is not a whole number and has .5 helps round to the nearest integer.
If you ```FromPixel( ToPixel(x) )``` the result is always within 2.5 from the original value, and the offset causes the pattern of the error to be centered on zero. So ```FromPixel( ToPixel(x) ) == abs( FromPixel( ToPixel(-x)) )```

Path smoothing using ```TestDot()``` and ```TestLine()```:
Made easier by changing ```m_bmPoints``` to ```std::list```.

First, adjust all points between start and goal, pulling them away from nearby obstacles.
Second, Remove all points between path corners with line of sight, divide the new longer segments
into smaller ones pulling the new points away from nearby obstacles.

Do that twice, first from goal to start, dividing into segments 5.5f long,
then from start to goal, dividing into segments 11.0f long.
The result is a more natural looking path around and between obstacles.

rasmusgo:
Fixing the goal radius helped! There is still one titanium cube that keeps getting bumped into but all four titanium ore are collected successfully!

tomangelo2:
I can confirm, it already does works a lot better, this mission is actually possible to finish now.

I'll try to check other levels for corner cases like this, but so far I really like how goto() turns out to be actually reliable.

rasmusgo:
> The fact that the cell size is not a whole number and has .5 helps round toward the nearest integer. If you `FromPixel( ToPixel(x) )` the result is always within 2.5 from the original value, and the offset causes the pattern of the error to be centered on zero. So `FromPixel( ToPixel(x) ) ~= abs( FromPixel( ToPixel(-x)) )` and `0.0f == FromPixel( ToPixel(0.0f) )` I chose the offset value by checking how wrong the result was converting To and FromPixel and by trial and error, I'm actually not exactly sure what I was doing there or why it seems to give more accurate results.

If we want to round towards the nearest integer instead of downwards (technically towards zero) we could simply use `std::round`. This should not change much though because I am already adding 0.5 before multiplying with `BM_DIM_STEP` when going from integer grid coords to world coords.

```c++
const int startX = static_cast<int>((start.x+1600.0f)/BM_DIM_STEP);
...
float floatX = (x + 0.5f) * BM_DIM_STEP - 1600.0f;
```

> Path smoothing using `TestDot()` and `TestLine()`: Made easier by changing `m_bmPoints` to `std::list`.

`std::list` is rarely the most performant choice even though it probably doesn't matter much here. When doing full passes over the data like this it can often be just as easy and more clear to create new vectors for each pass of modified data instead of modifying the same list in-place.

> First, adjust all points between start and goal, pulling them away from nearby obstacles. Second, Remove all points between path corners with line of sight, divide the new longer segments into smaller ones pulling the new points away from nearby obstacles.

Pulling away from nearby obstacles sounds like a good thing. How far do you move the points?

> Do that twice, first from goal to start, dividing into segments 5.5f long, then from start to goal, dividing into segments 11.0f long. The result is a more natural looking path around and between obstacles.

Why twice? Is it asymmetrical if you only do it once?

rasmusgo:
I would like to limit the scope of this PR to what it currently contains and let further enhancements come in other PRs.

melex750:
Maybe we can add a small heuristic cost when a point is near an obstacle with a formula like this one:
```c++
// Moves a point slightly away from nearby
// obstacles or to center between them.

Math::Vector CTaskGoto::FitPoint(Math::Vector p)
{
    // if the point is a little out of bounds, move it back in
    if (BitmapTestDot(0, ToPixel(p.x), ToPixel(p.z)))
    {
        float dist = 1000.0f; // seeking the smallest
        Math::Vector pnt = p;
        for (float ang = 0.0f; ang < 360.0f*Math::DEG_TO_RAD; ang += 30.0f*Math::DEG_TO_RAD)
        {
            float x = cosf(ang), z = sinf(ang);
            for (float d = 0.5f; d <= 10.5f; d += 0.5f)
            {
                Math::Vector test(p.x+d*x, p.y, p.z+d*z);
                if (!BitmapTestDot(0, ToPixel(test.x), ToPixel(test.z)))
                {
                    if (Math::DistanceProjected(test, p) < dist)
                    {
                        dist = Math::DistanceProjected(test, p);
                        pnt = test;
                        break;
                    }
                }
            }
        }
        p = pnt;
    }

    // find approximate center between nearby obstacles
    Math::Vector result(0.0f, 0.0f, 0.0f);

    for (float ang = 0.0f; ang < 360.0f*Math::DEG_TO_RAD; ang += 22.5f*Math::DEG_TO_RAD)
    {
        float d, x = cosf(ang), z = sinf(ang);
        for (d = 0.875f ; d < 14.0f; d += 0.875f)
        {
            if (BitmapTestDot(0, ToPixel(p.x+d*x), ToPixel(p.z+d*z))) break;
        }
        result.x += d * x;
        result.z += d * z;
    }

    return p + (result * 0.1f);
}
```

tomangelo2 (approved):
So far no issues found, new algorithm does find paths better than old one in difficult scenarios like navigating to objects on SpaceShip, so I think this PR can be merged.

PR #1501, opened 2022-02-10
Fix compilation under MacOS

tomangelo2:
TODO:
- [X] Fix homebrew installation command
- [X] Fix compilation
- [x] Fix OpenAL not being linked (compiler throws depreciation warnings, linker cannot find `OpenAL.framework` file)
- [x] Add MacOS builds to GitHub Actions
- [x] Fix compiled .dmg package (.dmg package is produced correctly, but it needs cloned submodules as well.)
Moved to another release:
- Look at alternatives for deprecated OS calls to make code more future-proof

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,7 +103,7 @@ jobs:
       run: make -j `nproc`
     - name: Install
       working-directory: build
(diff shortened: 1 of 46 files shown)


tomaszkax86 (approved):
Works fine.

PR #1507, opened 2022-03-06
Find space on spaceship

rasmusgo:
The cbot function `space` has been treating the spaceship and some other buildings as invalid places, This PR changes that to only take crash spheres and the terrain into account. This should make it easier to find free spots for placing stuff on the spaceship.

The releated cbot function `flatspace` is not affected and still treats the spaceship as an invalid place.


tomaszkax86 (approved):
The code is overall fine but needs 2 small fixes. To make things faster I'll approve and merge this PR and then apply my fixes on top of that.

PR #1512, opened 2022-03-18
Missing PL translation

krzysztofrewak:
Hi! I found out missing Polish translation in your PO files. I think it was French by mistake, but with this PR it should be good.

### DIFF
--- a/po/pl.po
+++ b/po/pl.po
@@ -86,7 +86,7 @@ msgid "Access to solution\\Shows the solution (detailed instructions for mission
 msgstr "Dostęp do rozwiązania\\Pokazuje rozwiązanie (szczegółowe instrukcje dotyczące misji)"
 
 msgid "Access to solutions\\Show program \"4: Solution\" in the exercises"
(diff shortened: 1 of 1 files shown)


po/pl.po
```diff
@@ -86,7 +86,7 @@ msgid "Access to solution\\Shows the solution (detailed instructions for mission
 msgstr "Dostęp do rozwiązania\\Pokazuje rozwiązanie (szczegółowe instrukcje dotyczące misji)"
 
 msgid "Access to solutions\\Show program \"4: Solution\" in the exercises"
-msgstr "Accčs aux solutions\\Programme \"4: Solution\" dans les exercices"
+msgstr "Dostęp do rozwiązań\\Pokaż program \"4: Rozwiązanie\" w ćwiczeniach"Brak cudzysłowu
```
smoorg, line 89: Why `brak cudzysłowu` is even there? Shouldn't it be removed?

krzysztofrewak: Of course. Fixed.

PR #1529, opened 2022-07-09
Fixes for space()

tomaszkax86:
Fixes for space() on spaceship and when it doesn't find a free location.


tomangelo2 (approved):
Nothing seems to break after this pull.

PR #1539, opened 2022-08-22
Add reference to the Void Linux package

tranzystorek-io:


tranzystorek-io:
I selected master as target branch on purpose, since this is a small non-code change to the README

Issue #1503, opened 2022-02-20
AlienWasp grab() and drop() functions are broken


tomangelo2:
I've noticed that in mission 9.1 my dev build started to crash some time after I started a game. I got such backtrace:

<details>
<summary>Click to expand</summary>

```gdb
(gdb) bt
#0  0x00007fff3b2890f3 in KERNELBASE!DebugBreak ()
   from C:\Windows\System32\KernelBase.dll
#1  0x00007fff3c04bb49 in msvcrt!_assert ()
   from C:\Windows\System32\msvcrt.dll
#2  0x00007ff645ae22df in CTaskManip::CTaskManip (this=0x2c8f2334f50,
    object=0x2c8f279cf80)
    at D:/msys64/home/user/colobot/src/object/task/taskmanip.cpp:61
#3  0x00007ff645cbf34b in MakeUnique<CTaskManip, COldObject*> ()
    at D:/msys64/home/user/colobot/src/common/make_unique.h:34
#4  0x00007ff645ccb14e in CTaskExecutorObjectImpl::StartForegroundTask<CTaskManip, TaskManipOrder&, TaskManipArm&> (this=0x2c8f279d020)
    at D:/msys64/home/user/colobot/src/object/implementation/task_executor_impl.cpp:145
#5  0x00007ff645bc4560 in CTaskExecutorObjectImpl::StartTaskManip (
    this=0x2c8f279d020, order=TMO_GRAB, arm=TMA_FFRONT)
    at D:/msys64/home/user/colobot/src/object/implementation/task_executor_impl.cpp:187
#6  0x00007ff645b083df in CScriptFunctions::rGrab (var=0x0,
    result=0x2c8f2175470, exception=@0xee6bffc478: 0, user=0x2c8f26e0a50)
    at D:/msys64/home/user/colobot/src/script/scriptfunc.cpp:2185
#7  0x00007ff645c6c538 in CBot::CBotExternalCallDefault::Run (
    this=0x2c8ee2e2c90, thisVar=0x0, pStack=0x2c8f27c0430)
    at D:/msys64/home/user/colobot/src/CBot/CBotExternalCall.cpp:160
#8  0x00007ff645c6c21a in CBot::CBotExternalCallList::DoCall (
    this=0x2c8ee2d5c40, token=0x2c8f27bb478, thisVar=0x0, ppVar=0xee6bffc650,
    pStack=0x2c8f27c0430, rettype=...)
[...]

Issue #930, opened 2017-03-09
Expanding level control abilities of CBOT [PROPOSAL]


Quartofel:
As @krzys-h mentioned on [forum](https://colobot.info/forum/showthread.php?tid=907&pid=7809#pid7809), didn't found it earlier anywhere here, so I've thought I'll write it down here.

_produce();_
- Add team variable for spawned bots;
- Add variable for initial shield level for spawned bots;
- Add variable for initial power level for spawned bots;
- Add possibility to precise initial program's directory;

_radar();_
- Add possibility to _radar_ from XYZ position;

_endmission;_
- Add _endmission;_ for CBOT (minor, as it's can be easily done with _produce();_ on met requirements);

_repair(id, rate, time);_
- Add command allowing repairing bots and buildings remotely (alternatively, negative values on _delete();_* could fix bots);

_charge(id, rate, time);_
- Add command allowing recharging bots and buildings remotely;

_maxflyheight_
- Add command allowing dynamically change max flying height;

_doneresearch(research name, team);_
- Add command allowing to instantly grant access to desired tech;

_expandhelp(file directory);_
- Add command allowing to expand information contained within SatCom during mission. Would greately improve storytelling abilities and made multi-stage objectives easily doable.

Misc non-CBOT related ideas:
- Add possibility to set teams for aliens;
- Add possibility to "find" neutral bots by teams and auto-setting their team to one of the finder's.

Obviously, all those functions should be regarded as "cheats", and should be used only for LevelControllers. Those improvments could greately increase possibilities for level creation. Obviously, I have no idea how hard would it be to implement any of above, and if the effort is worth it, but sharing those won't hurt anyone.

*Currently, _delete(id, negative value)_ drops target's _shieldLevel_ down to 0, regardless it's _shieldLevel_ before activating and, interestingely enough, without destroying the target.

krzys-h:
> radar(); - Add possibility to radar from XYZ position

This already exists, see search()

> endmission()

Also already exists. It's possible that it wasn't working at the time of writing this issue but it definitely does now.

```
endmission(result, delay)
result is one of ResultNotEnded, ResultLost, ResultLostQuick, ResultWin
delay is the EndMissionDelay value to use
```

> doneresearch(research name, team) - Add command allowing to instantly grant access to desired tech

Already possible. I think the command lacks team support though.
```
setresearchdone(getresearchdone() | ResearchXXXX);
```

> Add possibility to set teams for aliens

[...]

Issue #142, opened 2013-03-18
Interface improvements


krzys-h:
- AlienAnt
  - [ ] Add button for fire( _nearest object_ ), if it's in range (40 meters?)
- AlienSpider
  - [x] Add button for fire()
- AlienWasp
  - [x] Add button for grab()/drop()
- AlienQueen
  - [ ] Add buttons for produce(Alien____)
- Destroyer
  - [x] Add button for destroying robots (instead of time delay)
- Any (non-training) robot
  - [x] Add button for self-destruction (probably also as function in CBot) - changed, see comments below

PR #1543, opened 2022-09-04
Enchance alien units

lolbot-iichan:
This PR implements missing things requested at https://github.com/colobot/colobot/issues/142 and some more.
"Add possibility to set teams for aliens" was requested at https://github.com/colobot/colobot/issues/930
Maybe this also fixes https://github.com/colobot/colobot/issues/1503.

All aliens:
* can be selected if marked as "selectable=true" in scene file
* can be colored with `InsectColor team=<N> color=<R>;<G>;<B>`, similar to VehicleColor

AlienAnt:
* can fire with UI button and crosshair (ray tracing is used to detect which object to fire at)

AlienNest:
* now produce colored organic matter of one's team color

AlienQueen:
* can produce with UI buttons (see screenshot)
  * "Build" button (can be activated with Enter key) that activates the foreground task of producing selected object and moving forward for several seconds
  * 5 mode buttons to select what to produce: AlienAnt, AlienWasp, AlienSpider, AlienWorm, AlienNest
* if AlienQueen is selectable, it produces selectable objects, otherwise - produced objects are not selectable

AlienWasp:
* grab() fixed - now has a slot for grabbing like Subber

изображение


lolbot-iichan:
Maybe this also fixes #1503.

lolbot-iichan:
Abondoning this PR in favor to the modable colobot branch.

Issue #1551, opened 2022-09-13
Mathematical Instructions and Values (PI, E, Log, Abs, Root ...)


MarcosMD09:
Do you plan to add math instructions and constants to Colobot , such as ``PI``, ``E``, ``log``, ``abs``, ``root``, ...?

tomaszkax86:
Yes, we'll add various functions and mathematical constants to Colobot. This could take some time because all of them will require adding documentation and translation files.

MarcosMD09:
It's ok, it's just that I needed them for an advanced program.
<br>
<br>
By the way, here's an implementation of the `abs` statement:
```cpp
float Math::Abs(float value) {
    if(value < 0) {
        return value * -1;
    }

    return value;
}
```

tomaszkax86:
No, there's a C++ function for calculating this without conditional instructions and you're using very slow method with multiplication.

tomangelo2:
There already are some mathematical functions in game, like abs() or sqrt(), full list available in SatCom, in last section https://colobot.info/SatComOnline/master/cbot

It still lacks some functions, though.

MarcosMD09:
> No, there's a C++ function for calculating this without conditional instructions and you're using very slow method with multiplication.

Oh, okay

tomangelo2:
Was there a reason to close this issue? We could still add more mathematical functions/constants in future.

MarcosMD09:
I committed a error

MarcosMD09:
I committed a error.

MarcosMD09:
Logarithmic functions `log` and `log10`:

```cpp
bool​ ​rLog(CBotVar* var, CBotVar* result, ​int​& ​exception​, ​void​* user) 
 ​{ 
 ​    ​float​   value; 
  
 ​    value = var->​GetValFloat​(); 
 ​    result->​SetSetValFloat(logf(value)); 
 ​    ​return​ ​true​; 
 ​}

bool​ ​rLog10(CBotVar* var, CBotVar* result, ​int​& ​exception​, ​void​* user) 
 ​{ 
 ​    ​float​   value; 
  
 ​    value = var->​GetValFloat​(); 
 ​    result->​SetSetValFloat(log10f(value)); 
 ​    ​return​ ​true​; 
 ​}
```

MarcosMD09:
I suggested some functions, anyone can correct it?

PR #1553, opened 2022-09-18
Mathematical instructions and values (PI, E, Log, Log10, Root)

MarcosMD09:
As in #1551.

### DIFF
--- a/src/CBot/stdlib/MathFunctions.cpp
+++ b/src/CBot/stdlib/MathFunctions.cpp
@@ -292,7 +292,7 @@ void InitMathFunctions()
     CBotProgram::AddFunction("trunc", rTrunc, cOneFloat);
     CBotProgram::AddFunction("isnan", rIsNAN, cIsNAN);
     CBotProgram::AddFunction("log",   rLog,   cOneFloat);
-    CBotProgram::AddFunction("log10",   rLog10,   cOneFloat);
+    CBotProgram::AddFunction("log10", rLog10,   cOneFloat);
 }
 
 } // namespace CBot
--- a/src/CBot/stdlib/MathFunctions.cpp
+++ b/src/CBot/stdlib/MathFunctions.cpp
@@ -249,6 +249,28 @@ bool rIsNAN(CBotVar* var, CBotVar* result, int& exception, void* user)
     return true;
 }
 
+// Instruction "log(number)"
+
+bool rLog(CBotVar* var, CBotVar* result, int& exception, void* user)
+{
+    float   value;
+
+    value = var->GetValFloat();
+    result->SetValFloat(log(value));
+    return true;
+}
+
+// Instruction "log10(number)".
+
+bool rLog10(CBotVar* var, CBotVar* result, int& exception, void* user)
+{
+    float   value;
+
+    value = var->GetValFloat();
+    result->SetValFloat(log10(value));
+    return true;
+}
+
 } // namespace
 
 void InitMathFunctions()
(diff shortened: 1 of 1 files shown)


MarcosMD09:
At this moment there is no translations. I will add they later

melex750:
There are various overloads in C++ for ```log()``` and ```log10()``` to handle more data types than just ```float```, we can make a kind of polymorphic function to selectively call one of the overloads based on the given argument's type.
See functions ```cAbs()``` and ```rAbs()``` **[for example](https://github.com/colobot/colobot/blob/master/src/CBot/stdlib/MathFunctions.cpp#L143-L176)**.

MarcosMD09:
> There are various overloads in C++ for ```log()``` and ```log10()``` to handle more data types than just ```float```, we can make a kind of polymorphic function to selectively call one of the overloads based on the given argument's type.
> See functions ```cAbs()``` and ```rAbs()``` **[for example](https://github.com/colobot/colobot/blob/master/src/CBot/stdlib/MathFunctions.cpp#L143-L176)**.

Ok, I will do it.
Thanks you.

hexagonrecursion (changes requested):
The title of the pull request appears to be incorrect because it does not add PI and E and because `sqrt()` already exists in CBot. This pull request only adds log() and log10(). ~Other than that the code looks good.~

src/CBot/stdlib/MathFunctions.cpp
```diff
@@ -249,6 +249,52 @@ bool rIsNAN(CBotVar* var, CBotVar* result, int& exception, void* user)
...
+            break;
+    }
+    
+    return true;
+}
+
+CBotTypResult cLog(CBotVar* &var, void* user)
+{
+    if ( var == nullptr ) return CBotTypResult(CBotErrLowParam);
+    if ( var->GetType() != CBotTypFloat || var->GetType() != CBotTypDouble ) return CBotTypResult(CBotErrBadNum);
```
hexagonrecursion, line 289: This line is wrong. The parameter has to somehow simultaneously be float and double to pass this typecheck.

hexagonrecursion (changes requested):
1) Hold on. The following CBot code should compile, but I don't think your current implementation accepts this:
```c++
extern void Example()
{
    int x = 10;
    message(log(x));
}
```
2) I also get a bunch of compiler warnings:
```
src/CBot/stdlib/MathFunctions.cpp: In function ‘bool CBot::{anonymous}::rLog(CBot::CBotVar*, CBot::CBotVar*, int&, void*)’:
src/CBot/stdlib/MathFunctions.cpp:256:12: warning: enumeration value ‘CBotTypVoid’ not handled in switch [-Wswitch]
  256 |     switch (var->GetType())
      |            ^
src/CBot/stdlib/MathFunctions.cpp:256:12: warning: enumeration value ‘CBotTypByte’ not handled in switch [-Wswitch]
src/CBot/stdlib/MathFunctions.cpp:256:12: warning: enumeration value ‘CBotTypShort’ not handled in switch [-Wswitch]
src/CBot/stdlib/MathFunctions.cpp:256:12: warning: enumeration value ‘CBotTypChar’ not handled in switch [-Wswitch]
```
and more along the same lines
3) The added functions lack syntax highlighting and help "tooltips". I think this can be added in [IsFunction()](https://github.com/colobot/colobot/blob/d56c7dbfda6f87074aac8ffee1e22264caa01c32/colobot-base/src/script/cbottoken.cpp#L450-L450) and [GetHelpText()](https://github.com/colobot/colobot/blob/d56c7dbfda6f87074aac8ffee1e22264caa01c32/colobot-base/src/script/cbottoken.cpp#L545)

PR #1559, opened 2023-01-19
Add missing <cstdint> includes

suve:
Some parts of the code used the fixed-size type `uint32_t`, which is defined in `stdint.h` / `cstdint` - without including said header file. With GCC13, the inter-header dependencies seem to have changed and the file is no longer included transitively, requiring an explicit `#include`.


tomaszkax86 (approved):
I approve

PR #1576, opened 2023-04-19
Fix build with GCC 13

thesamesam:
GCC 13 (as usual for new compiler releases) shuffles around some internal includes so some are no longer transitively included.

See https://gnu.org/software/gcc/gcc-13/porting_to.html.

Bug: https://bugs.gentoo.org/899034

### DIFF
--- a/src/CBot/CBotFileUtils.h
+++ b/src/CBot/CBotFileUtils.h
@@ -19,6 +19,7 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


tomangelo2:
Already fixed with https://github.com/colobot/colobot/commit/d47e26586325ec11425cef5c95fc206dc103dbe2

PR #1583, opened 2023-06-04
Update data submodule url per new github policies

sparr:
https://github.blog/2021-09-01-improving-git-protocol-security-github/#git-protocol-troubleshooting

`git://` is no longer supported, and connection attempts will time out. This PR replaces with `git@github.com:` which may require a configured SSH key. Another less secure but more accessible replacement would be `https://`


sparr:
An equivalent change was already made in the dev branch.

PR #1585, opened 2023-06-09
Remove boost dependency

osiox:
### Brief
This change removes boost dependency, as suggested in #1306.
Boost has been replaced by c++ standard library, ICU and simpleini.
Not sure if it is still desired or acceptable, please check if you want/need such change.

I've tested this on Ubuntu with gcc and (once) on Windows 10 with MSVC, but I would suggest an additional verification before merging.

### Details

- Move common/stringutils to separate library (colobot_core).
  - Extracted to separate library to make it usable in CBot library.
  - **Proposal:**
  
      Different location of headers.
      While include style is the same ("core/strigutils.h"),
      the header is actually placed in core/include/core directory
      and core/include is added to include paths.
      This will lead to compilation errors if one uses core/* headers without adding colobot_core dependency.
      I suggest to use this style for whole project, after splitting src directory into multiple smaller libraries.
        
      This can of course be changed back to previous style (header in the same directory as cpp file) if desired.
        
- Replace Boost.Filesystem with std::filesystem.
- Replace Boost.Regex with std::regex.
- Replace Boost.Optional with std::optional.
- Replace Boost.Tokenizer with std based solution (std::getline).
- Replace Boost.PropertyTree with simpleini library.
- Replace Boost.Range (reverse) with std's reverse iterators.
- Replace Boost.Bimap with std::map, as it was only used to map TX_UNDEF to "undefined" string.
- Replace Boost's algoritm/string/case_conv (to_lower, to_upper) with ICU.
- Replace Boost's algoritm/string/replace (replace_all) with StrUtils::Replace (from colobot_core).
- Replace Boost's algoritm/string/join with std based solutions.
- Replace Boost's algoritm/string/trim with std based solution: StrUtils::Trim (added to colobot_core).
- Replace Boost's algoritm/string/split with std based solution: StrUtils::Split (added to colobot_core).
- Replace Boost's lexical_cast with StrUtils::ToString and StrUtils::FromString (from colobot_core).


osiox:
I was kinda expecting macosx build to fail, as I do not have mac to test it, but I totally do not know what is going on regarding windows build.

My vcpkg (& VS 2022) downloads grep from different url than colobot's CI and does it successfully. Is there something more I should change to make CI & vcpkg work ? (the "could not find ninja" error is also funny, considering ninja is installed and verified few steps before)

Btw. tell me if you consider the PR useless, I will not look into fixing CI builds then ;)

tomaszkax86:
Windows build was already broken for some reason, I think some download URL provided by vcpkg is broken. We'll have to fix that anyway so don't worry too much about it.

As for the PR, we won't be able to merge it to dev in this state. It most likely conflicts with a major feature branch that will be merged to dev and is waiting for release. But we can try to rebase your changes and fix merge conflicts after the merge happens.

We'll be doing some significant changes to repo structure and how source code is organized so I don't recommend introducing any such changes without prior consultation.

osiox:
I guess dev-graphics-overhaul ? I can rebase on top of it if removing boost is considered useful.

Anyway I've seen it mentioned in #1306 and decided to play with it. I've had fun refreshing my c++ a little and will not cry if this is closed :)

Sidenote: It seems I fixed macos build, it worked on my fork, so only windows fails to build on CI now.

osiox:
Rebase (& squash) on dev-graphics-overhaul done on [osiox:bye-boost-on-graphics-overhaul](https://github.com/osiox/colobot/tree/bye-boost-on-graphics-overhaul), just in case.

PatrickvL:
Was this merged into #1604 (and thus part of dev branch now)? Should in that case this pull request be closed?

osiox:
@PatrickvL It seems boost is gone on dev branch after graphics overhaul, so I guess you are right. Closing this PR.

PR #1593, opened 2023-06-28
Fix squashed planets when window isn't 4:3

### LINKED ISSUE #1584: Backgrounds of moon and earth are not circles
When watching the cut scene before the first moon mission, the earth and moon appear as ellipses. I suspect this is related to my screen resolution. The scenes should be cropped or letterboxed, not stretched.

immibis:
Fix #1584


tomaszkax86 (approved):
Verified, planets are finally proper circles.

PR #1606, opened 2023-08-23
Improve level loading time

tomaszkax86:
Added StrUtils::RemoveComments() and used it instead of std::regex in parsing level commands


tomangelo2 (approved):
Confirmed, lag on level select screen disappeared

immibis:
I believe the loop with text.erase is still O(n^2). It is possible to implement these kinds of removal algorithms in one pass, copying each character once (unless part of a comment) and truncating the string at the end

PR #1611, opened 2023-08-27
Build official releases through Github Actions

tomangelo2:
Made as draft just to check if it works with pulls as well.

Do not merge until:
- [ ] fixed Windows compilation
- [ ] artifact tested on Windows
- [ ] artifact tested on Linux
- [ ] artifact tested on Mac
- [ ] replace branch with `master`


hexagonrecursion:
Github says that you are trying to merge 579 commits. Did you mess something up in git? Github tells me that merging this branch into master will add 824 commits.

tomangelo2:
Probably. I've tried to update the branch so it would pull all changes from dev and apply new files on it. I thought rebasing dev into that branch will do that, but it seems to make a mess.

hexagonrecursion:
Here: found a version of your branch from before the rebase: https://github.com/colobot/colobot/compare/dev...hexagonrecursion:colobot:dev-release-ci-builds?expand=1 in case it helps.

hexagonrecursion:
> replace branch with master

Do you want to merge this into master? In that case why are you trying to get commits from dev?

tomangelo2:
I've targeted dev, so it would be easier to compare steps in case something wouldn't work (as it does right now). Also when pushed to `dev` it will eventually end up in `master` anyway.

> replace branch with master
That note is so I won't forget to change workflow to look into main branch before merge. Right now it only looks at this feature branch.
While merging I'll probably squash it all into single commit, since there are already some temporary commits. That should keep commit history clean.

tomangelo2:
Closing due to not having a skill to fight with runners.
MacOS is unhappy because `error: fixup_bundle: not a valid bundle`. Cthulhu knows what is this *invalid bundle*.
Windows is unhappy about linker throwing errors for some reason.
Linux is happy, but refuses to build AppImage for some reason

I cannot wrap my head around it, and there is no reason to keep that draft open.

Issue #1522, opened 2022-05-14
Moon Mission 1, and possibly others


Casteele72:
This is a minor issue, but still an issue. When playing Moon Mission 1, I have my bot do a `factory(WingedGrabber,"bot2main()")`, but the bot does not get the bot2main() program at all. It seems the default scenario example code to replace the EnergyCell is the only code that gets loaded in to the new bot.

hexagonrecursion:
It should be "bot2main", not "bot2main()". I will add an example to the documentation (eventually)

hexagonrecursion:
* The documentation improvement has been merged https://github.com/colobot/colobot-data/pull/89

I think this can be closed

PR #1620, opened 2023-11-16
Make factory() recognize the .cbot extension

hexagonrecursion:
The in-game editor saves programs with .cbot extension by default, but the `factory()` function only opens files with .txt extension. This pull request makes `factory()` open files with either .txt or .cbot extensions.

Note to self: after this is merged I will improve docs for factory():
- mention file extension requirement
- mention that you should write program/foo.txt instead of foo.txt
- remove incorrect link
- add public function example (see #1522)


hexagonrecursion:
Rebased because https://github.com/colobot/colobot/pull/1632 fixed one issue that made the build fail and one issue that caused a segfault.

tomaszkax86 (approved):
For now this is enough. The overall code could use a small rewrite and maybe API improvements but we'll do that another time.

PR #1626, opened 2023-11-21
Make flatspace() only return points on dry land

### LINKED ISSUE #918: The build() function cannot build at -0.5 altitude?
While it is already nearly impossible to write a fully auto script in mission 3-3 with the horribly buggy goto() function, the game hates me again?
Using the flatspace(), it returns some spot at -0.5 altitude, I know the astronaut can build on that altitude by hand, but the build() function says "Impossible under water".

I think I can still write a fully auto script by hardcoding all the exact coordinates for every step to complete the mission, but it is not what I intended...

### LINKED ISSUE #1625: flatspace() wants a pathological amount of flat space in some cases
When flatspace() evaluates a potential spot it finds the distance to the nearest object, and then checks that the radius of flat ground around that spot is at least half this distance (for some bizarre reason). Only if this is the case does flatspace() check the radius of flat ground the user actually requested. If the nearest object is far away, flatspace() may fail to find anything even though there is plenty of perfectly good flat ground.

https://github.com/colobot/colobot/blob/f07b7a1b60cd5685d7cd5f5677669423281fefa0/colobot-base/level/robotmain.cpp#L4229-L4236

I think this check should be removed.

To reproduce:

1. Install this mod [too-far.zip](https://github.com/colobot/colobot/files/13425455/too-far.zip)
2. Start the custom level "Too Far"
3. Find the program "TitaniumMarksTheSpot" in  the list of programs for the robot and run it
4. A message "space not found" will be printed. This indicates that flatspace() failed to find any valid spaces. This is because the only object in the level other than your robot is too far away and the only flat ground is near the center of the level
5. Debug spawn TitaniumOre nearby
6. Run TitaniumMarksTheSpot again. Now it successfully finds a space nearby.
7. Repeat step 6 as multiple times. It now finds plenty of usable spaces.

hexagonrecursion:
Fixes #918.

1. flatspace() is documentad as a function that should be "Useful for finding a place for a building"
2. You can not build under water

Also fixes #1625 and does some refactoring while I'm at it.

What I left for future consideration (will not be included in this pull request): flatspace() should probably return `new point(nan, nan, nan)` if it does not find a suitable location instead of returning `center` as it does now. This would be consistent with what `space()` does in this situation.

### DIFF
--- a/colobot-base/src/level/robotmain.cpp
+++ b/colobot-base/src/level/robotmain.cpp
@@ -4227,6 +4227,7 @@ bool CRobotMain::FlatFreeSpace(glm::vec3 &center, float minFlat, float minRadius
 	    
             if (SearchNearestObject(m_objMan.get(), pos, exclu) < space) continue;
             if (m_terrain->GetFlatZoneRadius(pos, minFlat) < minFlat) continue;
+            if (m_terrain->GetFloorLevel(pos) < m_water->GetLevel()) continue;
             center = pos;
             return true;
         }
--- a/colobot-base/src/level/robotmain.cpp
+++ b/colobot-base/src/level/robotmain.cpp
@@ -4225,11 +4225,8 @@ bool CRobotMain::FlatFreeSpace(glm::vec3 &center, float minFlat, float minRadius
             pos.y = 0.0f;
             m_terrain->AdjustToFloor(pos, true);
 	    
-            float dist = SearchNearestObject(m_objMan.get(), pos, exclu);
-	    if (dist < space) continue;
-            float flat = m_terrain->GetFlatZoneRadius(pos, minFlat);
-            if (flat < minFlat) continue;
-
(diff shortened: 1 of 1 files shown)


hexagonrecursion:
Rebased because https://github.com/colobot/colobot/pull/1632 fixed one issue that made the build fail and one issue that caused a segfault

hexagonrecursion:
@tomangelo2 @tomaszkax86 Ping? This pull request is still waiting for review.

tomaszkax86 (approved):

hexagonrecursion:
Hurray! Thanks @tomaszkax86! Thanks @Emxx52! :+1: :rocket: 🎉 Keep up the good work!

hexagonrecursion:
@Emxx52 I think this is a good candidate to be backported to the hotfix branch too (assuming the plan it to turn it into a hotfix release).

Issue #1636, opened 2023-12-09
Target "Colobot-Base" links to target "OpenAL::OpenAL" but the target was   not found


hexagonrecursion:
When building latest dev (6afdee472127c918ae1651a9c4f016fd71ed1b0d) I get this error.

```
git clone https://github.com/colobot/colobot.git
cd colobot
git submodule update --init
mkdir build
cd build
cmake -D TRANSLATIONS=OFF ../

CMake Error at colobot-base/CMakeLists.txt:7 (add_library):
  Target "Colobot-Base" links to target "OpenAL::OpenAL" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?


CMake Error at colobot-app/CMakeLists.txt:4 (add_executable):
  Target "colobot" links to target "OpenAL::OpenAL" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```

- Package libopenal-dev version 1:1.19.1-2build3 is installed
- cmake version 3.22.1
- Ubuntu 22.04.3 LTS

Workaround:
```
cmake -D TRANSLATIONS=OFF -D OPENAL_SOUND=OFF ../
```

hexagonrecursion:
Looks like colobot now requires cmake 3.25

https://cmake.org/cmake/help/latest/module/FindOpenAL.html#imported-targets

hexagonrecursion:
[cmake 3.25.0 was released Nov 16, 2022](https://github.com/Kitware/CMake/releases/tag/v3.25.0)
Based on [this post](https://askubuntu.com/a/355574/1189534) it appears Ubuntu 22.04.3 **LTS** will probably stay on cmake 3.22 forever (and [will eventually reach the end of "Standard Support" in June 2027](https://wiki.ubuntu.com/Releases)).

I am not sure what to do.
1. Should we require cmake 3.25?
2. Should we detect if cmake is too old and automatically disable OPENAL_SOUND?
3. Should we restore the cmake/FindOpenAL.cmake file and only use it if cmake if too old?

hexagonrecursion:
Note: [cmake version 3.27.9 can be installed via pip](https://stackoverflow.com/a/52265521/14516046)
```
apt remove cmake -y
pip install cmake --upgrade
```
[The pypi package appears to be an official (or at least officially endorsed) cmake build](https://cmake.org/download/#binary)

[...]

PR #1638, opened 2023-12-17
Define OpenAL::OpenAL target if it's missing

tomaszkax86:
Fixes issue #1636


tomangelo2 (approved):
Confirmed to fix on Ubuntu 22.04

PR #1642, opened 2024-01-02
Add glm and wget as build dependncies

sparr:
On Arch I installed `glm`. I wasn't able to update the Ubuntu instructions, which might require `libglm-dev` or similar.

I'll be making a PR against the data repo to use curl as an alternative to wget, but for now it's required.


AndreiCherniaev:
You are right! In Debian libglm-dev is[ what we need](https://github.com/colobot/colobot/pull/1643)!

tomangelo2 (approved):

PR #1645, opened 2024-02-11
Add Simplified Chinese interface translation

CloneWith:


CloneWith:
Seemingly the game cannot display CJK texts correctly (shown as blocks)

image

Need an appropriate font (e.g.`Noto Sans CJK`)

CloneWith:
Hi there. I would like to know more about the progress of this PR. Will this part of translation be accepted, or what do we still need to do to make it work?

tomaszkax86:
Hi. We'll need to change the font logic to support multiple fonts simultaneously, regardless of language selection. Currently we can have either DejaVu or CJK font but not both.

We'll also need to take a look at font rendering implementation because CJK fonts use much larger number of characters so they might cause performance issues due to the way they are processed internally.

Everything else seems fine. I can't judge translations because I don't know the language.

PR #1650, opened 2024-04-05
Remove DEV_BUILD

hexagonrecursion:
We have DEV_BUILD and we have COLOBOT_DEVELOPMENT_MODE. It was unclear to me which does what and which one I should use.

It turned out that DEV_BUILD has exactly the same effect as PORTABLE_SAVES.

I removed the DEV_BUILD option to reduce confusion.


hexagonrecursion:
- Merged by https://github.com/colobot/colobot/pull/1663

PR #1651, opened 2024-04-06
Fixed accidental removal of tooltip when displaying text with message()

### LINKED ISSUE #1649: The message() instruction and character escaping.
Using escaped characters inside the text works correctly. The problem occurs when displaying such text on the screen using the message() instruction.
All characters are output correctly except for "\\\\". Not only the slash is not output, but it also cuts off all the text following it.

For example, message("abc\\\\def"); will display only "abc".

tomaszkax86:
Fixes #1649

### DIFF
--- a/colobot-base/src/ui/controls/window.cpp
+++ b/colobot-base/src/ui/controls/window.cpp
@@ -154,22 +154,24 @@ CImage* CWindow::CreateImage(const glm::vec2& pos, const glm::vec2& dim, int ico
 
 // Creates a new label.
 
-CLabel* CWindow::CreateLabel(const glm::vec2& pos, const glm::vec2& dim, int icon, EventType eventMsg, std::string name, bool removeTooltip)
+CLabel* CWindow::CreateLabel(const glm::vec2& pos, const glm::vec2& dim, int icon, EventType eventMsg, std::string name)
 {
     CLabel* label = CreateControl<CLabel>(pos, dim, icon, eventMsg);
 
-    if (removeTooltip)
-    {
-        auto p = name.find("\\");
-        if (p == std::string::npos)
-            label->SetName(name, false);
-        else
-            label->SetName(name.substr(0, p), false);
-    }
-    else
-    {
+    auto p = name.find("\\");
+    if (p == std::string::npos)
         label->SetName(name, false);
-    }
+    else
+        label->SetName(name.substr(0, p), false);
+
+    return label;
+}
+
+CLabel* CWindow::CreateLabelRaw(const glm::vec2& pos, const glm::vec2& dim, int icon, EventType eventMsg, std::string name)
+{
+    CLabel* label = CreateControl<CLabel>(pos, dim, icon, eventMsg);
+
+    label->SetName(name, false);
 
     return label;
 }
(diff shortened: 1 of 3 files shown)


colobot-base/src/ui/controls/window.cpp
```diff
@@ -154,15 +154,22 @@ CImage* CWindow::CreateImage(const glm::vec2& pos, const glm::vec2& dim, int ico
...
+CLabel* CWindow::CreateLabel(const glm::vec2& pos, const glm::vec2& dim, int icon, EventType eventMsg, std::string name, bool removeTooltip)
 {
     CLabel* label = CreateControl<CLabel>(pos, dim, icon, eventMsg);
 
-    auto p = name.find("\\");
-    if (p == std::string::npos)
-        label->SetName(name);
+    if (removeTooltip)
+    {
+        auto p = name.find("\\");
```
hexagonrecursion, line 163: Do we actually rely on the tooltip-removing ability of this method? Where? How? Why? If this only displays the "main" part of the text, what displays the tooltip?

If we don't need this, we can just remove the whole find&substr part.

tomaszkax86: From my understanding, labels remove the tooltip part of the text by default. This might be to allow sharing of texts between labels and other controls that do display tooltips. So if you had a button with text `"New player\\Create new player profile"`, a button would have `"New player"` text and `"Create new player profile"` tooltip. But if you set this to a label, it would only display `"New player"` and had no tooltip.

I don't know why it's done like this, I find this very clunky and confusing; tooltip should always be separate from the main text. But I wrote this code to preserve existing behavior, even if it's weird, in order to hopefully not introduce any new bugs.

hexagonrecursion: I found an example, where we do depend on tooltip removal

This appears to contain both the tooltip for the slider and the label:

https://github.com/colobot/colobot/blob/03b2da88e5b83016b3a15453d2cd10214eb7d317/colobot-base/src/common/restext.cpp#L243

https://github.com/colobot/colobot/blob/03b2da88e5b83016b3a15453d2cd10214eb7d317/colobot-base/src/ui/screen/screen_setup_game.cpp#L104-L110

image

tomaszkax86: Turns out that refactoring SetName() will take a lot more work than I anticipated. I'd leave it the way it is, it can be addressed in a separate issue. I did the other change though.

hexagonrecursion (changes requested):
I have tested the changes and found a bug.

When I run the following code at first everything works as expected:
```
extern void object::New()
{
	message("foo\\bar");
	wait(1);
	message("foo\\bar");
	wait(1);
	message("foo\\bar");
}
```

image

But after one of the messages times-out the other two loose the part after the backslash:

image

I think this line is the culprit:

https://github.com/colobot/colobot/blob/b5b6e45f7811d9ae78ddca49b5e660e71bf44fa5/colobot-base/src/ui/displaytext.cpp#L393

hexagonrecursion:
Is this ready for review?

tomaszkax86:
I think current fix is enough for now, unless you found new bugs or something it's ready for review.

Changing SetName() and its weirdness will require proper refactoring. I'll make a separate issue for that.

tomangelo2 (approved):

PR #1654, opened 2024-04-21
l10n: Fix manpage generation when there's more than just fr(ench)

OdyX:
I've gone ahead and added an initial `desktop/po/de.po` german translation, and noticed that the manpage generation doesn't work anymore when there's more than one translation.

This should fix it


tomaszkax86 (approved):
Seems fine to me

PR #1657, opened 2024-04-23
Ubuntu 22.04 build: use the provided deps, not the submodules

OdyX:
This matches what is done in the Debian packaging (and avoid depending on the submodules)


tomaszkax86 (approved):
Approved

PR #1659, opened 2024-04-23
Update GitHub actions: build against Debian, and simplify / cleanup Github Actions

OdyX:


tomaszkax86 (approved):
Overall good changes I think. We might not need wget but we can change that later.

tomaszkax86:
@OdyX There's a merge conflict that needs to be resolved. Could you fix it? I could then merge the changes.

OdyX:
Force-pushed a cleaned-up history; thanks for the review!

PR #1661, opened 2024-05-05
Github actions: Make sure the local submodules are not used in Debian/Ubuntu builds

OdyX:


hexagonrecursion:
Hello. Could you please elaborate what is the purpose of the change? Why should continuous integration use distribution packages over compiling the dependencies from source code? What is the purpose of skipping this step if container is "" and os is "ubuntu-20"?

OdyX:
Good questions @hexagonrecursion :-)

I come to contribute here as Debian maintainer for the Colobot packages (which then get automatically integrated to Ubuntu. As it is standard practice in Debian to rely solely on distribution packages (to avoid code duplicates, enhance the security handling, etc). So the purpose of the proposed change is to make sure the builds work correctly when **not** using the embedded dependencies.

```
matrix.container != '' || matrix.host_os == 'ubuntu-22.04'
```
… is another way of saying "Debian (run in container) or recent-enough Ubuntu" (because we know nlohmann-json3 is not available in the older Ubuntus: https://launchpad.net/ubuntu/+source/nlohmann-json3)

hexagonrecursion (approved):

tomaszkax86 (approved):
Seems to work just fine.

Issue #1518, opened 2022-04-09
Game crashes on autosave with "Fatal error: got negative system counter difference!" in logs


Baguettedood:
OS: EndeavourOS (Arch) Linux with kernel 5.17.1-arch
My drive is an NVME SSD with a btrfs filesystem if it matters.
Game version: 0.2.0-alpha

After playing for a while, the game abruptly closes with the following messages at the end of its log:
```
[INFO]: Autosave!
[ERROR]: Fatal error: got negative system counter difference!
[ERROR]: This should never happen. Please report this error.
[INFO]: Exiting with code 0
[INFO]: Unloading files and closing device...
```

When reopening the game I can load the autosave that was just created immediately before the crash, though the screenshot seems to have failed to create properly.

It seems to be more likely to crash (and maybe only crashes at all) when the game speed is increased. I usually play at 2x speed as that has felt comfortable for me for a long time.

Attached log files:
[log_crash_2022-04-09_15_52.txt](https://github.com/colobot/colobot/files/8457403/log_crash_2022-04-09_15_52.txt)
[log_Crash_2022-04-09_22_24.txt](https://github.com/colobot/colobot/files/8457404/log_Crash_2022-04-09_22_24.txt)
[log_Trace_Debug_All_Crash_2022-04-10_02_07.txt](https://github.com/colobot/colobot/files/8457405/log_Trace_Debug_All_Crash_2022-04-10_02_07.txt)
[log_Trace_Crash_2022-04-10_02_14.txt](https://github.com/colobot/colobot/files/8457407/log_Trace_Crash_2022-04-10_02_14.txt)
[log_Trace_Crash_2022-04-10_02_21.txt](https://github.com/colobot/colobot/files/8457409/log_Trace_Crash_2022-04-10_02_21.txt)
[log_Trace_Crash_2022-04-10_02_48.txt](https://github.com/colobot/colobot/files/8457410/log_Trace_Crash_2022-04-10_02_48.txt) - In this last once I left the game idle to autosave a few times at various game speeds. The saves were successful until I increased to 2x speed.

As far as I can tell, manual saves and quicksaves are not affected.

Zemogiter:
I have the same issue. Win10, same game version
[log.txt](https://github.com/colobot/colobot/files/8985495/log.txt)
The only log file I was able to find. I was also playing at a multiplayer of a speed.

Bielecki:
[...]

PR #1663, opened 2024-05-08
Update utilities

tomaszkax86:
This pull requests updates a variety of CMake and system utilities.

General list of changes:
* CSystemUtils is now a proper singleton; you're not meant to create multiple not fully initialized instances of it.
* Time utilities now use `std::chrono::steady_clock` instead of `std::chrono::high_resolution_clock`. This fixes a problem with negative time counter differences.
* Extracted compiler- and platform-specific configuration to separate CMake files. With this change it should be easier to port Colobot to other platforms and use different compilers.
* The game can be run in a different directory than the executable's directory. Helps run the game from build directory on multi-config build systems.
* Removed some of explicit compiler options. These options should be specified by the user or in CMake configuration profiles.
* Changed default build type to RelWithDebInfo. There's no point in using Debug by default.
* Moved error handling (signal handlers) to CSystemUtils
* Implemented stack trace in critical error messages for Windows


hexagonrecursion:
> * Time utilities now use `std::chrono::steady_clock` instead of `std::chrono::high_resolution_clock`. This fixes a problem with negative time counter differences.

* Does this fix https://github.com/colobot/colobot/issues/1518 and https://github.com/colobot/colobot/issues/1679 ?

tomaszkax86:
> * Does this fix [Game crashes on autosave with "Fatal error: got negative system counter difference!" in logs #1518](https://github.com/colobot/colobot/issues/1518) and [Crash to desktop due to negative system counter difference #1679](https://github.com/colobot/colobot/issues/1679) ?

It should. High-res clock isn't guaranteed to be monotonic and can be aliased to system clock. So it can go back in time when the system randomly chooses to update the system clock. Steady clock is guaranteed to be monotonic and never go back in time.

tomangelo2 (approved):

colobot-base/src/common/system/system.h
```diff
@@ -75,14 +75,19 @@ enum class SystemDialogResult
 class CSystemUtils
 {
 public:
+    CSystemUtils();
```
hexagonrecursion, line 78: This constructor should be protectedc instead of public

colobot-base/src/common/system/system.cpp
```diff
@@ -149,13 +146,7 @@ SystemDialogResult CSystemUtils::ConsoleSystemDialog(SystemDialogType type, cons
...
 std::filesystem::path CSystemUtils::GetBasePath()
 {
-    if (m_basePath.empty())
-    {
-        auto* path = SDL_GetBasePath();
-        m_basePath = StrUtils::ToPath(path);
-        SDL_free(path);
-    }
-    return m_basePath;
+    return std::filesystem::current_path();
```
hexagonrecursion: Why? Before this `PORTABLE` builds searched for files relative to the location of the executable. Now we will only find the files if the current working directory happens to be the directory of the executable. Isn't this strictly less reliable?

tomaszkax86: Less reliable how? Unless you specifically change it, the current directory is the same as directory of the executable so there's no difference in behavior. This change removes requirement for strict location of the executable in relation to game files. You will now be able to place it in `bin` subdirectory, for example. In some ways it's more portable than before.

This change also allows running the project from the build directory on multi-config systems, like MSVC. There, the executable is placed in config-specific subdirectory even if you point `CMAKE_RUNTIME_OUTPUT_DIRECTORY` to the base build directory, meaning that running the project after building is impossible without first copying the executable. With this change you can run the executable from any build location because you can manually point the current directory to the base build directory and the game will find its assets normally.

Unless you mean that the current directory might be changed during execution and corrupt something as a result because the returned value is different. In that case we can cache the current directory on creation so even if it changes this method will return the same value.

tomaszkax86: Actually, the current directory is already cached as `m_basePath` in the constructor if you check the updated code:
```
CSystemUtils::CSystemUtils()
{
    m_basePath = std::filesystem::current_path();
}
```
The returned base directory value cannot change on its own. The value can still be changed by a derived class and this is done in the Linux implementation when building for AppImage is enabled. I think this is fine.

hexagonrecursion: > Unless you specifically change it, the current directory is the same as directory of the executable

Not if stared from the command line.

tomaszkax86: > Not if stared from the command line.

Running an executable from a directory other than its own directory is a deliberate choice and not something you do randomly. Under normal circumstances you don't want to run the game from any other directory, even from terminal.

And as a developer you want to be able to change executable's behavior by using whatever current directory is convenient to you. In this case, for the purpose of running the game from the build directory. I do this all the time as a developer in most of my projects.

Also, AppImage has override to use old behavior. Do you think there should be a CMake option for all platforms to change whether base path is set to executable's directory or current directory?

hexagonrecursion: > Do you think there should be a CMake option for all platforms to change whether base path is set to executable's directory or current directory?

No. No. No. There are too many build options as is. It's fine

colobot-base/src/common/system/system.h
tomaszkax86: Making the constructor protected is unnecessary because the class is already abstract, it contains pure virtual methods and cannot be instantiated. However, I made the destructor pure virtual to make this more clear.

hexagonrecursion: More clear? Nope. The opposite.
1. Someone who missed that the class has a pure virtual method (e.g. me) will just as easily miss a virtual destructor.
2. Someone who does notice the destructor will now have more questions. The first question will be "does this work?" (b.t.w [it does not unless you implement the destructor](https://stackoverflow.com/a/630960/14516046)) and the second question will be "why does this class need a pure virtual destructor?"

Issue #1623, opened 2023-11-19
Crash when loading a level after changing AI code


hexagonrecursion:
When loading a save file the game loads the stack frames of the currently running programs from the save, but it loads the AI code from the game or mod files. If the AI code has changed, the code and the stack do not match. In some cases this can cause a crash.

0. Here is an archive with a mod and a save where the below steps have already been done:  [crash.zip](https://github.com/colobot/colobot/files/13402785/crash.zip). Just install the mod and open the save <---- crash
1. Create a mod. Use this script as AI for a programmable CreateObject
```
extern void object::Bug()
{
	while(true) {
		message("start");
		wait(1);
		wait(1);
		wait(1);
		wait(1);
		wait(1);
		wait(1);
		wait(1);
		wait(1);
		wait(1);
		wait(1);
		wait(1);
	}
}
```
2. Start the level
3. save
4. Change the script to
```
extern void object::Bug()
{
	while(true) {
		message("start");
		wait(1);
	}
}
```
5. load <-- the game will crash

It is possible to fix this specific example by making CBotListInstr trust the stack a little less, but the solution scales poorly: every CBot instruction would have to fully distrust the stack contents because they may have been created by a different instruction entirely.

I think a better solution would be to compute some kind or hash or checksum ([CRC](https://en.wikipedia.org/wiki/Cyclic_redundancy_check)?) of the AI source code and only load the stack if it matches the value from the save file. Otherwise the safest option is to restart the AI code from the beginning - should work for most levels.

If someone wants to fix this, it may make sense to fix #614 at the same time

Issue #534, opened 2015-07-26
Colobot sometimes ignores scene file when starting code battle


CHmSID:
I'm talking about scripts specifically. Sometimes when I make a change to the scene file or the script that's used in that scene file, when I start the level no changes have been made. I'm sure I save the files.

For example, I wanted to disable the script of team2 in the code battle example, so in the scene file I changed team2.txt to team4.txt(where no such file exists). The console said it could not load the script from the scene file, but when I loaded the level, the robot had its old code still loaded.

Another one, I made a small change to my own script and after saving the file, reloaded the level. But the changes have not been made there. Furthermore, when I deleted all the code from the in-game editor, every time I started this level, the code was always empty (or simply not on the list if I've deleted it previously).

It all looks like the game is caching the scripts somewhere.

krzys-h:
Yes, it saves the scripts in player's directory. This is the same behaviour that saves the user-created scripts after mission restart. If you remove the file, the game assumes this script was created by the user and loads it. This is the intended behaviour.
Anyway, I think it would be better to never save scripts from scene files.

PR #1668, opened 2024-05-11
Save level-supplied programs when saving the game

hexagonrecursion:
CBot scripts that are loaded from the game level or from a mod (such as the scripts implementing enemy AI) may change from one version of the game or mod to the next.

Instead of preserving those scripts as part of the save file we were only storing file names and loading new modified versions of those scripts when when loading a save.

When saving the game we save the current execution state of all running programs. Attempting to continue running the program from where we left off after the source code changed may cause the script to behave strangely or even cause **the whole game** to crash.

I fixed this by saving all scripts when saving the game. This way the saved execution state always matches the source code.

* Fixes https://github.com/colobot/colobot/issues/1623
* Undoes "Never save copy of programs from level files when saving scene" 4182be321695306385543c0584cb780959ceb25e

Legacy save files do not contain some of the source code that was available when the game was saved. The version of the mod/game may have changed since - we don't know. Those files may (in general) contain classes and public functions. To make sure that loading a legacy save file does not crash the game we discard the current execution state of all scripts and start them over from the beginning.

Unlike player-written programs, level-supplied programs are not saved to be loaded when the level is restarted. This should prevent re-introduction of https://github.com/colobot/colobot/issues/534

### DIFF
--- a/colobot-base/src/level/robotmain.cpp
+++ b/colobot-base/src/level/robotmain.cpp
@@ -4782,7 +4782,7 @@ bool CRobotMain::IOWriteScene(const std::filesystem::path& filename,
     if (!ostr.is_open()) return false;
 
     bool bError = false;
-    long version = 1;
+    long version = 2;
     CBot::WriteLong(ostr, version);                 // version of COLOBOT
     version = CBot::CBotProgram::GetVersion();
     CBot::WriteLong(ostr, version);                 // version of CBOT
@@ -5018,7 +5018,7 @@ CObject* CRobotMain::IOReadScene(const std::filesystem::path& filename,
         bool bError = false;
         long version = 0;
         CBot::ReadLong(istr, version);             // version of COLOBOT
-        if (version == 1)
+        if (version == 2)
         {
             CBot::ReadLong(istr, version);         // version of CBOT
             if (version == CBot::CBotProgram::GetVersion())
--- a/colobot-base/src/object/auto/autoegg.cpp
+++ b/colobot-base/src/object/auto/autoegg.cpp
@@ -198,7 +198,6 @@ bool CAutoEgg::EventProcess(const Event &event)
             Program* program = programStorage->AddProgram();
             programStorage->ReadProgram(program, StrUtils::ToString(InjectLevelPathsForCurrentLevel(m_alienProgramName, "ai")));
             program->readOnly = true;
-            program->filename = m_alienProgramName;
             programmable->RunProgram(program);
         }
         Init();
--- a/colobot-base/src/object/auto/autofactory.cpp
+++ b/colobot-base/src/object/auto/autofactory.cpp
@@ -680,7 +680,6 @@ bool CAutoFactory::CreateVehicle()
             Program* prog = programStorage->AddProgram();
             programStorage->ReadProgram(prog, StrUtils::ToString(InjectLevelPathsForCurrentLevel(name, "ai")));
             prog->readOnly = true;
-            prog->filename = name;
         }
     }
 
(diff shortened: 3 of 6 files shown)


colobot-base/src/level/robotmain.cpp
```diff
@@ -4782,7 +4782,7 @@ bool CRobotMain::IOWriteScene(const std::filesystem::path& filename,
     if (!ostr.is_open()) return false;
 
     bool bError = false;
-    long version = 1;
+    long version = 2;
```
Emxx52, line 4785: This part of the code and similar version checks should be completely rewritten, but since it is outside of this PR's scope I'd suggest just leaving the 'version' as is for now.

Emxx52 (changes requested):

colobot-base/src/level/robotmain.cpp
hexagonrecursion: I do not understand. What do you mean by "completely rewritten"? Why?

Emxx52: I'd rather not have magic numbers to keep track of and update before every single major release anywhere in the code; or version defines like this one for that matter: https://github.com/colobot/colobot/blob/c56d240a0f8a65d5645c2454be95f93dde067658/CBot/src/CBot/CBotDefines.h#L28
robotmain.cpp has notoriously long stretches of barely readable code that should be rewritten, this is just one such fragment.

hexagonrecursion: This version filed should **not** be updated "before every single major release". This field should **only** be changed when we break backwards compatibility of the `cbot.run` file or, like in this case, when we can't trust old `cbot.run` files not to crash the game. I'll turn the magic number into a constant.

hexagonrecursion: > since it is outside of this PR's scope I'd suggest just leaving the 'version' as is for now.

Nope. This is an essential part of the pull request. The whole point is that we want to be able to change alien AI scripts without causing old save files to crash the game - we have to discard cbot.run.

Emxx52: I get the idea, I'm just not fond of the old code.

And I cannot be sure we're going to remember to update it (especially when it's hidden away like this and not at all documented) every time we change something in CBOT and/or the game that would break backwards compatibility. 

Besides, loading old save files is generally unsupported. We more often than not encourage the players to remove their save files after each update. Loading an old save file might work, or it might not - preventing a crash is always great though so I understand your reasoning. However, discarding `cbot.run` would already break some levels anyway so I don't really see the point of using such a save file.

On a related note, we might want to add a warning on loading old save files and actually make use of the part of the code that writes the game version to `data.sav` in the future.

hexagonrecursion: > We more often than not encourage the players to remove their save files after each update.

1. No we don't. If it's not in the user interface, it does not count because [uses don't read the documentation](https://www.joelonsoftware.com/2000/04/26/designing-for-people-who-have-better-things-to-do-with-their-lives/).
2. The two most recent release announcements [0.2.1](https://colobot.info/2023/08/08/gotofixes/) and [0.2.0](https://colobot.info/2021/08/21/planb/) did not contain such a warning either.
3. [INSTALL.md has no such warnings eitehr](https://github.com/colobot/colobot/blob/c56d240a0f8a65d5645c2454be95f93dde067658/INSTALL.md)
4. [And neither does README-dev.md](https://github.com/colobot/colobot/blob/c56d240a0f8a65d5645c2454be95f93dde067658/docs/README-dev.md)

> we might want to add a warning on loading old save files

My plan now is to add a new pull request that _prevents_ the game from listing a save file as available to be loaded unless the version of the game that created the file exactly matches the current release. Only supporting save files created by the same game release should make our lives much simpler (e.g. don't need to worry about breaking save files when introducing new built-in functions). We may still need the current pull request too because of mods as well as people play-testing the dev branch.

hexagonrecursion:
* This was made obsolete by #1683

Issue #1624, opened 2023-11-19
Underwater alien worms


hexagonrecursion:
In the level 2 "Subterranean strategy" of chaper 4 "Crystalium" the AI of AlienWorms sometimes makes them move underwater, where they can not be attacked. This is very annoying for full-automation playthroughs. I think the AI should be changed to make the worms stay on land.

![image](https://github.com/colobot/colobot/assets/52621858/f9e99e6f-b6ce-41a6-967e-3cde1a791e1a)

 #1623 should probably be fixed before fixing this.

Issue #1565, opened 2023-03-11
Update Function ant01.TXT


gamersalpha:
there are several issues with the given code, here's a corrected version:
```

`extern void object::Attack()
{
    object p;
    float dist, prox;
    point nav1, nav2, dest;
    boolean advance = true;

    errmode(0); // do not stop if there is an error
    while (ismovie() != 0)
        wait(1);

    while (true)
    {
        p = radar(WingedShooter, 0, 360, 0, 50);
        if (p == null)
        {
            nav1.x = position.x - 5;
            nav1.y = position.y;
            nav1.z = position.z;
            nav2.x = position.x + 5;
            nav2.y = position.y;
            nav2.z = position.z;

            while (true)
            {
                goto(nav1);
                p = radar(WingedShooter, 0, 360, 0, 50);
                if (p != null)
                    break;

                goto(nav2);
                p = radar(WingedShooter, 0, 360, 0, 50);
                if (p != null)
                    break;
            }
        }
        else
        {
            dist = distance(p.position, position);
            if (dist <= 40 && !advance)
            {
                fire(p);
                advance = true;
            }
            else
            {
                prox = dist - 5;
                if (prox > 40)
                    prox = 40;
                if (prox < 5)
                    prox = 5;
                dest.x = (position.x - p.position.x) * prox / dist + p.position.x;
                dest.y = (position.y - p.position.y) * prox / dist + p.position.y;
                dest.z = (position.z - p.position.z) * prox / dist + p.position.z;
                goto(dest);
                advance = false;
            }
        }
        wait(1);
    }
}
```
`

Here are the changes made to the original code:

Added wait(1) at the end of the while loop to avoid overloading the CPU.
Added nav1.z and nav2.z to set the correct z-coordinate for navigation points.
Changed fire(p.position) to fire(p) to fire at the target object instead of its position.
Changed p.position.z to p.position in the calculation of dest to use the object's position instead of just the z-coordinate.

hexagonrecursion:
* I will look into this after https://github.com/colobot/colobot/pull/1683 is merged

hexagonrecursion:
It appears that ant01.TXT is not used in any missions in vanilla - all references to it were removed in colobot-gold-0.1.3-alpha https://github.com/colobot/colobot-data/commit/17d36f64baa6dfa5dec2a47a669bc23f9a2f39b1. Why do you want to change dead code? It should probably be deleted instead.
```sh
$ git -C data/ log -S ant01.txt
[...]

Issue #935, opened 2017-03-13
Terranova 9.2 - mission impossible?


Anton-V-K:
The mission with shielder, subber and astranaut (no resources or weapons).
When the key is found and everyone are back on the space ship, the ants (one or two of them) are blocking the doors, thus preventing the launch. And it seems there is no way to kick them out... or do I miss something?

tomangelo2:
I think only way out of this situation is to start again and hope ants won't block doors this time. At least I didn't found out better solution for this. It was like that since original game, still it should be fixed at some point. Maybe ants should have some instruction in their script to step back if they're too close to SpaceShip?

RaptorParkowsky:
Well, this indeed happening sometimes. AI needs some rewriting.

X64E:
@krzys-h @tomangelo2 @MrSimbax 

Takeoff with anit is possible. I think issue may be closed. Tested on 2.0 version.
 
![изображение](https://user-images.githubusercontent.com/95232774/145621573-7e84a587-9bb5-4c45-a9f6-3ff5949e9b1c.png)

X64E:
Confirm if ants on the ship doors takeoff is not possible. Tested on 2.0 version.

hexagonrecursion:
* I will fix this after https://github.com/colobot/colobot/pull/1683 makes it possible to change the AI without risking game crash when old saves are loaded

PR #1683, opened 2024-07-04
Add a warning about unsupported save files

### LINKED ISSUE #1623: Crash when loading a level after changing AI code

hexagonrecursion:
## Problem

* Fixes #1623
* Also addresses some concerns I have about a number of edge cases where public classes or public functions may cause a crash when loading a save if there were changes to the CBot API (even if the changes do not break backward compatibility).

In https://github.com/colobot/colobot/pull/1668#discussion_r1599331147 @Emxx52  told me "loading old save files is generally unsupported".

## Solution

1. Loading an unsuported crashsave or quicksave now behaves as if there is no save
2. The -loadsave option also does not recognise a save as valid unless its version is supported
3. Loading an unsupported save via the save list asks the player to confirm

## Note to self: after this is merged I should implement a fix for:
* https://github.com/colobot/colobot/issues/1624
* https://github.com/colobot/colobot/issues/935
* https://github.com/colobot/colobot/issues/1565


hexagonrecursion:
* Blocked: waiting for #1781 to land due to merge conflicts

hexagonrecursion:
Who should I ask to review this?

eschan145 (approved):

tomaszkax86:
I'll review this today or tomorrow.

colobot-base/src/common/version.h.cmake
```diff
@@ -13,5 +13,9 @@ namespace Version
 
     static inline constexpr std::string_view FULL_NAME = "Colobot: Gold Edition @COLOBOT_VERSION_FULL@";
     static inline constexpr std::string_view VERSION = "@CMAKE_PROJECT_VERSION@";
+    static inline constexpr int MAJOR = @CMAKE_PROJECT_VERSION_MAJOR@;
+    static inline constexpr int MINOR = @CMAKE_PROJECT_VERSION_MINOR@;
+    static inline constexpr int PATCH = @CMAKE_PROJECT_VERSION_PATCH@;
+    static inline constexpr int TWEAK = @CMAKE_PROJECT_VERSION_TWEAK@;
```
Emxx52, line 19: TWEAK should be removed, we will not be using it in future releases.

colobot-base/src/ui/screen/screen_io_read.cpp
```diff
@@ -168,10 +171,22 @@ bool CScreenIORead::EventProcess(const Event &event)
...
+        if (!name.has_value()) return false;
+        if (IsVersionSaveSupported(name.value()))
         {
-            m_main->StopSuspend();
-            m_main->ChangePhase(PHASE_SIMUL);
+            doRead();
+        }
+        else
+        {
+            m_dialog->StartQuestion("Version unsupported. Open anyway?", false, false, false, doRead);
```
Emxx52, line 189: This text should be translatable (see [restext.cpp](https://github.com/colobot/colobot/blob/dev/colobot-base/src/common/restext.cpp), `update-pot`).

Emxx52 (changes requested):

colobot-base/src/common/version.h.cmake
tomaszkax86: Correct, we had ideas for it but that's unnecessary work for little benefits. Top level project's tweak number will also be removed.

CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ cmake_minimum_required(VERSION 3.21)
 
 cmake_policy(SET CMP0025 NEW)
 
+# Loading save files created by a previous version of the game is
+# unsupported with one exception: the first three version components have to match
+# and the fourth version component has to be greater or equal.
```
tomaszkax86, line 12: This description has no place in top level CMakeLists.txt. It should be moved near relevant implementation details, like in a comment before the function that checks save file version.

colobot-base/src/ui/screen/screen_io_read.cpp
```diff
@@ -168,10 +171,22 @@ bool CScreenIORead::EventProcess(const Event &event)
 
     if ( event.type == EVENT_INTERFACE_IOREAD )
     {
-        if(IOReadScene() && m_inSimulation)
+        auto doRead = [&]() {
```
tomaszkax86, line 174: Minor: `{` should be in next line

colobot-base/src/ui/screen/screen_io.h
```diff
@@ -41,6 +42,7 @@ class CScreenIO : public CScreen
     void IODeleteScene();
     void IOWriteScene();
     bool IOReadScene();
+    std::optional<std::filesystem::path> GetSceneName();
```
tomaszkax86, line 45: Minor: if possible, make the method `const` qualified

colobot-base/src/ui/screen/screen_io_read.cpp
tomaszkax86, line 189: I'd change the text to be more descriptive. Something like `"This save is from a different version of the game and might not work correctly. Do you want to open it anyway?"`

tomaszkax86 (changes requested):

CMakeLists.txt
hexagonrecursion: Removing TWEAK made this comment redundant

hexagonrecursion:
@tomaszkax86 @Emxx52 Thanks. Done

hexagonrecursion:
In case this somehow fell through the cracks: I have applied review suggestions. Can someone review this again please?

tomaszkax86 (approved):

Emxx52 (changes requested):
Works well, now it's only missing updated po files. Please run `make update-pot` and then build the project to generate them (you might need to rebase first).

hexagonrecursion:
> Works well, now it's only missing updated po files. Please run `make update-pot` and then build the project to generate them (you might need to rebase first).

Thanks. Done

TheGiraffe3 (approved):
waring in the title to warning?
Code changes look good.

hexagonrecursion:
> waring in the title to warning?

I do not understand. Are you talking about the commit message? Are you talking about the title of the pull request? Are you talking about the text of the new pop-up I added?

Are you suggesting we should change it?... Oh. I see the typo. Thanks.

hexagonrecursion:
@TheGiraffe3 in order to understand your comment I had to:
1. Notice that one of the words is spelled wrong
2. Somehow guess that your spelling is intentional

A more reliable way to communicate this would be: "There is a typo in the title: waring instead of warning"

TheGiraffe3:
My apologies.

Emxx52 (approved):

PR #1703, opened 2024-07-30
Allow camerafocus() to force first/third person view

hexagonrecursion:
We have an instruction called `camerafocus(object o)`. Unfortunately this instruction does not let us choose between first person view and third person view. This pull request adds an optional argument `camerafocus(object o, [int view])` and three constants:
* `camerafocus(obj, DefaultView)` is equivalent to `camerafocus(obj)`
* `camerafocus(obj, FirstPersonView)` selects the object and switches to the first person view
* `camerafocus(obj, ThirdPersonView)` selects the object and switches to the third person view

### DIFF
--- a/colobot-base/src/script/scriptfunc.cpp
+++ b/colobot-base/src/script/scriptfunc.cpp
@@ -72,6 +72,13 @@
 
 using namespace CBot;
 
+enum CameraView
+{
+    DefaultView,
+    FirstPersonView,
+    ThirdPersonView,
+};
+
 CBotTypResult CScriptFunctions::cClassNull(CBotVar* thisclass, CBotVar* &var)
 {
     return cNull(var, nullptr);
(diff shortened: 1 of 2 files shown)


colobot-base/src/script/scriptfunc.cpp
```diff
@@ -72,6 +72,13 @@
 
 using namespace CBot;
 
+enum CameraView
+{
+    DefaultView,
+    FirstPersonView,
+    ThirdPersonView,
+};
+
```
tomaszkax86, line 81: New enum types should be preferably `enum class`. Enumerators should use `SCREAMING_SNAKE_CASE`.

colobot-base/src/script/scriptfunc.cpp
```diff
@@ -3527,6 +3570,10 @@ void CScriptFunctions::Init()
     CBotProgram::DefineNum("ResearchBuilder",       RESEARCH_BUILDER);
     CBotProgram::DefineNum("ResearchTarget",        RESEARCH_TARGET);
 
+    CBotProgram::DefineNum("DefaultView",           DefaultView);
+    CBotProgram::DefineNum("FirstPersonView",       FirstPersonView);
+    CBotProgram::DefineNum("ThirdPersonView",       ThirdPersonView);
```
tomaszkax86, line 3575: CBot constants that act as enumerators should preferably have a common prefix rather than a suffix. And I'd use `Camera` rather than `View` for the prefix since this changes the camera type and the function that uses it has `camera` in the name. So something like:

`DefaultView` -> `CameraDefault`
`FirstPersonView` -> `CameraFirstPerson`
`ThirdPersonView` -> `CameraThirdPerson`

tomaszkax86 (changes requested):
Seems like a decent change overall but I have a couple of requests.

colobot-base/src/script/scriptfunc.cpp
Emxx52: Agreed. Tho I'd suggest naming the first and third person constants `CameraOnboard` and `CameraBack` respectively as these are the terms used both in Scene and behind the scenes (heh).

colobot-base/src/script/scriptfunc.cpp
hexagonrecursion: Done

colobot-base/src/script/scriptfunc.cpp
hexagonrecursion: Done

tomaszkax86 (approved):

hexagonrecursion:
Thanks, @tomaszkax86 and @Emxx52 👍 🚀 🎉

PR #1718, opened 2024-08-10
Remove commended code in CBotProgram

hexagonrecursion:
This code was commented out since the initial commit. Let's remove it


melex750:
I want to bring that one back into service...

hexagonrecursion:
Why? What is it supposed to do?

melex750:
> I want to bring that one back into service...

In fact, I already have in one of my many WIPs.

> Why? What is it supposed to do?

Hint: Overload resolution. I don't want to spoil the surprise, but we can if necessary.

PR #1720, opened 2024-08-13
Make public functions and classes in CBot not shared between teams

### LINKED ISSUE #526: Make public functions in CBot not shared between teams

hexagonrecursion:
* Fixes #526

All public functions and classes used to live in one shared namespace. This caused many issues including:
* Two teams in a codebattle could not define a public function with the same name.
* A public function defined in an alien could be called by a robot

Solution:
* In a codebattle each team should  have its own namespace
* Every bot or alien that the player can't control should be in a namespace separate from the player-controlled teams

- [x] Make public functions not shared
- [ ] Make public classes not shared

### DIFF
--- a/CBot/CMakeLists.txt
+++ b/CBot/CMakeLists.txt
@@ -133,6 +133,8 @@ target_sources(CBot PRIVATE
     src/CBot/CBotInstr/CBotWhile.h
     src/CBot/CBotProgram.cpp
     src/CBot/CBotProgram.h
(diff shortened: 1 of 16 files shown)


Zemogiter:
What if player uses a cheat code to controll a NPC bot or alien? Do they suddenly join the same namespace?

hexagonrecursion:
Cheat codes will work in a way that is useful for development and troubleshooting: using a cheat code **does not** change which namespace each bot or alien belongs to.

PR #1725, opened 2024-08-17
Fix proxyActivate on objects with a power cell

hexagonrecursion:
* Fixes #1175

Mousing over and clicking an object taht has a power cell used to bypass proxyActivate=1

PR #1726, opened 2024-08-17
Fix power=0

### LINKED ISSUE #1070: Attribute power=0 applied to object in scene.txt leaves it without power cell
When in scene.txt you create a bot with attribute power=0, in the game it will not get it's power cell having 0% power, there will be no power cell at all. For such there is power=-1 already, so power=0 should have been likely meaning that bot has power cell with 0 energy, but it was likely implemented incorrectly by Epsitec devs. In original level 4-4 "Lost Valley" we can find several bots scattered around the valley and they all have no power cells. Giving their positions and how many power cells are around, it looks really odd.

hexagonrecursion:
* Fixes #1070
* See also #1725

[The documentation says that power=0 is supposed to result in an empty power cell](https://colobot.info/files/old/original/colobotusere.pdf), but the game was spawning no power cell instead 

image

Now power=0 causes a bot to spawn with an empty power cell

### DIFF
--- a/colobot-base/src/object/motion/motionvehicle.cpp
+++ b/colobot-base/src/object/motion/motionvehicle.cpp
@@ -1057,7 +1057,7 @@ void CMotionVehicle::Create(glm::vec3 pos, float angle, ObjectType type,
     CreatePhysics(type);
     m_object->SetFloorHeight(0.0f);
 
-    if (power > 0.0f            &&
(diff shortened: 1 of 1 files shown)


hexagonrecursion:
Restarting CI

tomaszkax86 (approved):

Issue #1635, opened 2023-12-07
MSYS2/gcc won't compile the game


tomangelo2:
While trying to compile the `dev-modernize-build-system` branch on MSYS2/gcc this error shows up:
```
In file included from D:/msys64/mingw64/include/c++/13.2.0/bits/stl_algobase.h:63,
                 from D:/msys64/mingw64/include/c++/13.2.0/bits/specfun.h:43,
                 from D:/msys64/mingw64/include/c++/13.2.0/cmath:3699,
                 from D:/msys64/mingw64/include/c++/13.2.0/math.h:36,
                 from D:/msys64/mingw64/include/SDL2/SDL_stdinc.h:86:
D:/msys64/mingw64/include/c++/13.2.0/ext/numeric_traits.h: In instantiation of 'struct __gnu_cxx::__
numeric_traits_integer<__int128 unsigned>':
D:/msys64/mingw64/include/c++/13.2.0/bits/max_size_type.h:776:39:   required from here
D:/msys64/mingw64/include/c++/13.2.0/ext/numeric_traits.h:64:53: error: static assertion failed: inv
alid specialization
   64 |       static_assert(__is_integer_nonstrict<_Value>::__value,
      |                                                     ^~~~~~~
D:/msys64/mingw64/include/c++/13.2.0/ext/numeric_traits.h:64:53: note: 'std::__is_integer<__int128 u
nsigned>::__value' evaluates to false
D:/msys64/home/user/colobot-modern/colobot-base/src/app/app.cpp:69:1: error: ISO C++ does not suppor
t '__int128' for 'test' [-Wpedantic]
   69 | __int128 test;
      | ^~~~~~~~
In file included from D:/msys64/mingw64/include/c++/13.2.0/bits/chrono_io.h:39,
                 from D:/msys64/mingw64/include/c++/13.2.0/chrono:3330,
                 from D:/msys64/home/user/colobot-modern/colobot-base/src/common/timeutils.h:25,
                 from D:/msys64/home/user/colobot-modern/colobot-base/src/common/system/system.h:28,
                 from D:/msys64/home/user/colobot-modern/colobot-base/src/app/app.h:30:
D:/msys64/mingw64/include/c++/13.2.0/format: In instantiation of 'std::__format::_Formatting_scanner
<std::__format::_Sink_iter<wchar_t>, wchar_t>::_M_format_arg(std::size_t)::<lambda(auto:45&)> [with
auto:45 = __int128]':
[...]

PR #1730, opened 2024-08-17
Fix building on MSYS2/MinGW

Emxx52:
Thanks to @hexagonrecursion for pinning down the cause of the issue. (https://github.com/colobot/colobot/issues/1635#issuecomment-2266685425)


hexagonrecursion:
BTW what was the original problem with `putenv()`? Does anyone know? Was it fixed?

tomaszkax86 (approved):

PR #1763, opened 2024-08-28
Add Subber's claw to BuilderBots

melex750:
Discussion thread: #1514
Currently only works with ```grab()``` and ```drop()```

```c++
extern void object::Grab_Drop()
{
    (this.load == null) ? grab() : drop();
}
```


Emxx52:
While an interesting idea on paper, the feature would break the 'one unit, one tool' rule. The resulting amalgam bot looks rather weird - it screams non-vanilla. I don't think we can merge this PR, and there are no changes I could request. It's a neat idea for a mod though.

hexagonrecursion:
> It's a neat idea for a mod though.

Option 1: make it per-level setting (off by default)
Option 2: instead of adding a claw to a builder add a new bot type (4 types because of different movement options) Something like BuilderGrabber

This way a mod can use this if it wants to

Emxx52:
>This way a mod can use this if it wants to

We plan on creating a more robust modding system with which creating such a bot would be a matter of adding a line or two in a text file. Implementing such a feature would be a gargantuan task as you can imagine, so I can't promise we can deliver on that anytime soon though.

melex750:
That's ok, it was a fun experiment. It brought back an old program I wrote when Grabbers could build things.

Issue #855, opened 2016-11-07
Binary file access in CBOT


Adikso:
It would be nice if we could read and write other file types beside text files. For example reading data from bitmap file (which is very simple format) is not possible. Having binary file access would allow us to read every file type and parse it later in our code.

tomaszkax86:
While we're at it, it might be useful to add more variable types to CBot. `char`, `byte`, `short`, `long` and `double` can be essential when we deal with binary data.

immibis:
What is the purpose - when would an in-game robot need to access binary files?

melex750:
Maybe add these new methods to file class:
```c++
void    put   (byte c);    // insert a byte at the current output position
int     get   ();          // get the byte at the current input position or -1 if EOF
int     peek  ();          // peek the next byte without a change in position

void    seekp (long p);    // set the output position 
void    seekg (long p);    // set the input position

long    tellp ();          // get the current output position
long    tellg ();          // get the current input position

void    write (string s);  // write non-formatted string to file
string  read  (long n);    // read n bytes from file
```
Current file methods:
```c++
bool    open    (name, mode); // returns true if file was opened
void    close   ();           //

void    writeln (string s);   // writes string + '\n' to file
string  readln  ();           // reads a line, discards (CR)LF, returns string

bool    eof     ();           // returns true when end-of-file was reached
```
BTW, in-game docs do not mention the following:
 * open() returns boolean.
 * second param for open() has default value "r"
 * file constructor can be used with open() args.
 * file class has public field 'filename'

Issue #694, opened 2015-12-20
CBOT strings should be arrays


MrSimbax:
Right now strings can't be treated like an array, which is really odd and uncomfortable. While we're at it, it would be cool if we could use the array access operator like in Python: [p:q] returning the part of the array from index p inclusive to q exclusive.

krzys-h:
:+1: definetly going to happen

krzys-h:
CBot doesn't have a char type, so either we implement it or use one-char strings.
If we use strings, what do you think should be the expected outcome of this code:
```C++
string s = "Colobot";
s[3] = "test";
message(s);
```

MrSimbax:
CBOT should have char type and the code above should give a compilation error.

tomaszkax86:
```string``` is a special compound type, not a simple array of characters. If we want to be language correct, ```[]``` should be defined as some kind of indexing operator which could be generalized to other types. ```char``` type needs to be implemented and should be Unicode character type. Slicing is a Python specific feature and does not make any sense in CBot syntax. If you want it in CBot, it has to be specialized.
There's also a question of whether ```string``` should be mutable or immutable, and whether it is passed by value or reference. Colobot is kind of multi-threaded environment so it has to be taken into account.

krzys-h:
> If we want to be language correct, ```[]``` should be defined as some kind of indexing operator which could be generalized to other types.

That's exactly what I would want to do. At least internally, I'm not sure if it should be also provided to user-defined classes, and if yes, what syntax would be used for that.

> Slicing is a Python specific feature and does not make any sense in CBot syntax.

:+1:

> There's also a question of whether string should be mutable or immutable, and whether it is passed by value or reference.

Strings are always passed by value. Everything except classes (with exception of intrinsic classes) is passed by value.

MrSimbax:
Any argument why the slicing feature doesn't make any sense in CBOT? Because for me it totally does. And it is convenient so I don't see a reason why wouldn't we include it instead of some C-like functions like strleft, strmid, strright...

From CBOT programmer's point of view, string IS a simple array of characters, at least I've always treated them as C-strings. I don't care how does this work under the hood. If you want to make strings better then make them a proper object type and get rid of those C-like functions, which are currently the only way to work with strings in CBOT. The question is if you want to make CBOT more Java-like or C++-like.

[...]

PR #1767, opened 2024-08-31
Print byte as a number

### LINKED ISSUE #1446: message doesn't display the numeric value of type byte, but the character representation (game crashes sometimes)
Test case:

    byte a = -1;
    message (a);
    byte b = 256;
    message (b);
    byte c = a - 1;
    message (c);
    byte d = b + 1;
    message (d);

There is no compiler or runtime error.
`message` just shows strange characters.

This is probably also true for the other integer types, but it's not so obvious because numbers outside the range don't occur so often. 
Assigning a number outside the valid range should be detected already at compile time.
Leaving the range in a calculation should produce a runtime error.

hexagonrecursion:
Fix #1446

In CBot the type `byte` behaves like `int`, but with a smaller range of values. The `message()` function used to incorrectly interpret it as utf-8 encoded text. Now printing it should correctly display it as a number.

```c++
extern void New()
{
	byte b = 88;
	message (b);
}
```

image

### DIFF
--- a/CBot/src/CBot/CBotVar/CBotVarValue.h
+++ b/CBot/src/CBot/CBotVar/CBotVarValue.h
@@ -290,6 +290,13 @@ public:
     {
         this->m_val = ~(this->m_val);
     }
+
+    std::string GetValString() const override
+    {
+        std::ostringstream s;
+        s << +this->m_val;
+        return s.str();
+    }
 };
 
 } // namespace CBot
--- a/test/src/CBot/CBot_test.cpp
+++ b/test/src/CBot/CBot_test.cpp
@@ -3542,3 +3542,12 @@ TEST_F(CBotUT, TestFinally) {
         CBotErrBadType2
     );
 }
+
+TEST_F(CBotUT, TestByteToString) {
+    ExecuteTest(
+        "extern void TestByteToString() {\n"
+        "    byte c = 90;\n"
+        "    ASSERT(c + \"\" == \"90\");\n"
+        "}\n"
+    );
+}


melex750:
I think this should be done in CScriptFunctions::rMessage() with var->GetValInt() so we can still append raw bytes to a string.

hexagonrecursion:
@melex750 Must the interface of the CBot `string` type interpret it as a sequence of bytes? Must we repeat the mistakes of python2?

melex750:
I don't know. #694 #855

hexagonrecursion:
@melex750 

#855 does not mention strings at all. Sure, we would need a way to take a range of bytes from a binary file or socket and interpret it as a string, but this does not require `string + byte` to treat the string as a sequence of bytes rather than as text

In #694 [the consensus seems to be that `string` should act as an array of CBot type `char` - unicode code points](https://github.com/colobot/colobot/issues/694#issuecomment-167260371)

`string` is for storing text. When reading of binary files is implemented we will need a new type to represent a sequence of bytes (or maybe we can use `byte[]` - an array of type CBot type `byte`)

Trying to make `string` serve both purposes would result in a hodgepodge interface where some instructions treat `string` as a sequence of characters and some - as a sequence of bytes. It's much better to have two types for two different purposes plus some way to convert one to another.

Allowing raw bytes to be appended to a string locks us into using utf-8 as the internal representation. This makes the "get n-th Unicode code point" an inefficient O(n) operation.

CC @tomaszkax86

melex750:
> ```string``` should act as an array of CBot type ```char``` - unicode code points

This sounds good to me. We could also make ```string``` an object and even add methods for various purposes.

tomaszkax86 (approved):

PR #1770, opened 2024-09-01
Let wasps and winged bots fly regardless of research

### LINKED ISSUE #1093: Make flying objects able to fly regardless of status of research on jet engines
1. If you control a wasp manually and jet engines are not researched in current level, you are not able to make it fly. If it's in the air when you take control over it it drops down instantly.
2. If you find a winged bot you're unable to make it fly as well, even though it was created before you and by logic should already be supplied with necessary technologies.

I propose to make all the objects that are supposed to fly able to fly regardless of status of research on jet engines. Only Me (astronaut) should be in exceptions, because he always has visible jet engine as a part of it's suit. With that, only astronaut's jet engine and your ability to construct new bots supplied with jet engines will depend on this research.

hexagonrecursion:
* Fix #1093

### DIFF
--- a/colobot-base/src/object/old_object.cpp
+++ b/colobot-base/src/object/old_object.cpp
@@ -2256,7 +2256,7 @@ bool COldObject::EventProcess(const Event &event)
                     axeY *= 0.5f;
                 }
 
-                if ( !m_main->IsResearchDone(RESEARCH_FLY, GetTeam()) )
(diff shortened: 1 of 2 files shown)


tomaszkax86 (approved):

hexagonrecursion:
@tomaszkax86 Thank you :+1: :1st_place_medal: :+1: :+1:

PR #1747, opened 2024-08-21
Convert more paths to std::filesystem::path

hexagonrecursion:

PR #1781, opened 2024-10-02
Convert even more paths to std::filesystem::path

hexagonrecursion:
* Continuation from #1747

* **Note: this pull request includes** #1778

## Context

In our codebase std::string usually (always?) uses utf-8. Conversions of std::filesystem::path to/from std::string provided by the C++ standard library assume std::string uses "native narrow encoding". This was causing encoding problems on windows.

### DIFF
--- a/colobot-base/src/ui/controls/edit.cpp
+++ b/colobot-base/src/ui/controls/edit.cpp
@@ -806,23 +806,22 @@ void CEdit::HyperFlush()
 void CEdit::HyperHome(std::string filename)
 {
     HyperFlush();
-    HyperAdd(filename, 0);
+    HyperAdd(TempToPath(filename), 0);
 }
 
 // Performs a hyper jump through a link.
 
-void CEdit::HyperJump(std::string name, std::string marker)
+void CEdit::HyperJump(const std::filesystem::path& name, std::string marker)
 {
     if ( m_historyCurrent >= 0 )
     {
         m_history[m_historyCurrent].firstLine = m_lineFirst;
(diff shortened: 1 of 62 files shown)


hexagonrecursion:
I have eliminated all usage of TempToString() and TempToPath(). I kept the functions themselves in case we want to refactor more paths

colobot-base/src/ui/controls/edit.cpp
```diff
@@ -1617,7 +1616,7 @@ bool CEdit::ReadText(const std::filesystem::path& filename)
             if ( m_bSoluce || !bInSoluce )
             {
                 HyperLink link;
-                link.name = GetNameParam(buffer.data()+i+3, 0);
+                link.name = StrUtils::ToPath(GetNameParam(buffer.data()+i+3, 0));
```
melex750, line 1616: Many SatCom links are broken here most likely due to help files using backslash in file paths.

OS: Xubuntu 24.04.1 LTS

melex750: This line from log seems to show backslash as part of the filename:

> [ERROR]: Failed to load text file help/E/cbot\radarall.txt

hexagonrecursion: Found it. I removed this line:

https://github.com/colobot/colobot/blob/84ac0d3ba8bc7c590da0b999f132f765f6d56ed1/src/ui/controls/edit.cpp#L819

Will fix soon

PR #1789, opened 2024-11-27
Reenable -Werror in CI

hexagonrecursion:
For now - only for gcc because there are still some clang warnings we did not tackle

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -54,7 +54,8 @@
             "inherits": ["Linux-CI-base"],
             "cacheVariables": {
                 "CMAKE_C_COMPILER": "gcc",
-                "CMAKE_CXX_COMPILER": "g++"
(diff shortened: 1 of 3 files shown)


tomaszkax86 (approved):

TheGiraffe3 (approved):

hexagonrecursion:
I see some new warnings in dev. Will fix soon.

tomaszkax86:
We probably shouldn't rely on assert() to handle error detection. When it is disabled these two functions invoke undefined behavior. They should instead throw an exception or cause program termination with stack trace information.

hexagonrecursion:
> We probably shouldn't rely on assert() to handle error detection. When it is disabled these two functions invoke undefined behavior. They should instead throw an exception or cause program termination with stack trace information.

* Thanks. Done: https://github.com/colobot/colobot/pull/1795

PR #1792, opened 2024-12-06
Tests: print stack trace from exceptions

hexagonrecursion:
Sometimes our tests throw exceptions. This is easy to debug when the error can be reproduced locally, but I recently had to debug an issue that I could only reproduce in CI on macos-14. It turns out that gtest can't print the stack trace when it catches an exception. Figuring out what code throws and why without a stack trace turned out to be impossible.

I am adding a workaround to make similar issues easier to debug in the future:
1. Tell gtest to not catch exceptions
2. Wrap cpptrace around RUN_ALL_TESTS() to capture and print the stack trace as well as the error message

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -299,6 +299,14 @@ if(TESTS)
     # Hippomocks library
     add_subdirectory(${colobot_SOURCE_DIR}/lib/hippomocks)
 
+    include(FetchContent)
+    FetchContent_Declare(
+      cpptrace
+      GIT_REPOSITORY https://github.com/jeremy-rifkin/cpptrace.git
+      GIT_TAG        v0.7.3 # <HASH or TAG>
+    )
+    FetchContent_MakeAvailable(cpptrace)
+
     # Tests targets
     enable_testing()
     include(GoogleTest)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -299,6 +299,14 @@ if(TESTS)
     # Hippomocks library
     add_subdirectory(${colobot_SOURCE_DIR}/lib/hippomocks)
 
+    include(FetchContent)
+    FetchContent_Declare(
+      cpptrace
+      GIT_REPOSITORY https://github.com/jeremy-rifkin/cpptrace.git
+      GIT_TAG        v0.7.3 # <HASH or TAG>
```
hexagonrecursion, line 306: TODO:

https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare

> Where contents are being fetched from a remote location and you do not control that server, it is advisable to use a hash for GIT_TAG rather than a branch or tag name. A commit hash is more secure and helps to confirm that the downloaded contents are what you expected.

hexagonrecursion: I think I should use a submodule to be consistent with how other dependencies are handled

PR #1801, opened 2024-12-13
Fix race condition in worker_thread.h

hexagonrecursion:
The fields of CWorkerThread that are used in `Run()` have to be constructed before the constructor of std::thread is called to prevent a race condition.


hexagonrecursion:
It would be great if this was merged soon because I trigger this bug a lot when using gdb making debugging other issues harder

tomangelo2 (approved):

Issue #1012, opened 2017-07-27
invui cheat code makes cheat console and pause menu invisible


tomangelo2:
This makes sense as it's part of UI, but it makes it harder to disable this cheat when it's no longer needed. I suggest that console could be actually visible after pressing cheat button when invui is active.

PR #1808, opened 2024-12-15
Unhide UI when typing cheat

hexagonrecursion:
* fix https://github.com/colobot/colobot/issues/1012

It's hard to type cheats when the UI is hidden. The simplest solution I could think of: activating cheat console ends the effect of `invui`

### DIFF
--- a/colobot-base/src/level/robotmain.cpp
+++ b/colobot-base/src/level/robotmain.cpp
@@ -809,6 +809,7 @@ bool CRobotMain::ProcessEvent(Event &event)
                 if (m_phase == PHASE_SIMUL) m_cmdEditPause = m_pause->ActivatePause(PAUSE_ENGINE);
                 m_cmdEdit = true;
                 m_commandHistoryIndex = -1; // no element selected in command history
+                m_engine->SetRenderInterface(true);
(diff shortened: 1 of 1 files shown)


hexagonrecursion:
Alternative solution:
```diff
void CEngine::DrawInterface()
{
    m_device->SetDepthTest(false);
    m_device->SetTransparency(TransparencyMode::NONE);

    SetInterfaceCoordinates();

    // Force new state to disable lighting
    m_interfaceMode = true;

    // Draw the entire interface
    Ui::CInterface* interface = CRobotMain::GetInstancePointer()->GetInterface();
-    if (interface != nullptr && m_renderInterface)
+   if (interface != nullptr && (m_renderInterface || CRobotMain::GetInstancePointer()->GetCmdEdit()))
    {
        interface->Draw();
    }
```

Note: `GetCmdEdit()` would return `m_cmdEdit`.

This way you can type cheats, but the state of m_renderInterface is preserved. What do you think?

tomangelo2:
I like the second approach more. This way if someone hides the UI and then inserts another cheat code, they don't need to hide UI again.

PR #1819, opened 2024-12-18
Save command console history

melex750:
Hopefully saving Devs, Level Authors, and Beta-Testers a lot of keystrokes.

### DIFF
--- a/colobot-base/src/common/settings.cpp
+++ b/colobot-base/src/common/settings.cpp
@@ -124,6 +124,12 @@ void CSettings::SaveAudioSettings()
     GetConfigFile().SetBoolProperty("Setup", "FocusLostMute", m_focusLostMute);
 }
 
+void CSettings::SaveCommandHistory(const std::deque<std::string>& commands)
+{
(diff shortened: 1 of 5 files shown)


colobot-base/src/common/settings.cpp
```diff
@@ -269,7 +275,9 @@ void CSettings::LoadSettings()
 
     CInput::GetInstancePointer()->LoadKeyBindings();
 
-
+    std::deque<std::string> commands;
+    GetConfigFile().GetDequeProperty("Setup", "CommandHistory", commands);
+    main->SetCommandHistory(commands);
```
hexagonrecursion, line 280: SetCommandHistory() takes ownership of the memory instead of creating a copy. Someone could later introduce a bug by expecting it to create a copy. Use `std::deque<std::string>&&` in SetCommandHistory and use std::move here to make the ownership transfer more obvious.

hexagonrecursion (approved):

tomangelo2 (approved):

PR #1820, opened 2024-12-19
Improved warning dialog

melex750:
Added ability to scroll, select, and copy text in the warning dialog.  Allowing for more verbose/helpful error messages.

P.S. Level parser exceptions are displayed in this dialog.

### DIFF
--- a/colobot-base/src/ui/maindialog.cpp
+++ b/colobot-base/src/ui/maindialog.cpp
@@ -38,6 +38,7 @@
 #include "sound/sound.h"
 
 #include "ui/controls/button.h"
+#include "ui/controls/edit.h"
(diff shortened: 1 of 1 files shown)


hexagonrecursion (changes requested):
Why only for warnings? Why not make CMainDialog::StartInformation() always use CreateEdit?

melex750:
> Why only for warnings? Why not make CMainDialog::StartInformation() always use CreateEdit?

The main reason is CLabel centers the text by default and I didn't mess up the other things that might rely on that.

hexagonrecursion (approved):

tomangelo2 (approved):

PR #1821, opened 2024-12-19
Fix initial text in CBOT editor save dialog

melex750:
Specifically, after load then save, or save twice a file in a sub-folder.


hexagonrecursion (approved):
I tested this fix. LGTM

tomangelo2 (approved):

PR #1823, opened 2024-12-20
Fix loadig of the first `object`

### LINKED ISSUE #1696: [bug] Object position is lost on save/load
Under some circumstances the position of an `object` is not saved/loaded properly
## To reproduce
1. Start the below script
2. Quicksave while `wait(5)` is executing
3. Quickload

```c++
extern void object::New()
{
	object foo = radar(Me);
	wait(5);
	message(foo.position+"");
}
```
## Expected
The position of the astronaut is printed
## Actual result
```
point(x=undefined, y=undefined, z=undefined)
```
## Additional info
Adding code to access `foo.position` before save/load prevents the position from being undefined, but the coordinates printed are stale - if the astronaut moves, the old position will be printed.
```c++
extern void object::New()
{
	object foo = radar(Me);
	message(foo.position+"");
	wait(5); // save&load here
	message(foo.position+"");
}
```
## Affected versions
* gold-0.2.1 - affected
* dev (5bbe0f2b3e7846c80f25d4d464ea375a4970f91b) affected
* Other versions - not tested

hexagonrecursion:
* fix #1696

CreateObject ids used to start with 0. Because of this CBot values of type `object` referring to the first CreateObject in the scene used to not load correctly.

Solution: start numbering the objects from 1 instead.

### DIFF
--- a/colobot-base/src/object/object_manager.cpp
+++ b/colobot-base/src/object/object_manager.cpp
@@ -45,7 +45,7 @@ CObjectManager::CObjectManager(Gfx::CEngine* engine,
                                                oldModelManager,
                                                modelManager,
                                                particle)),
-    m_nextId(0),
+    m_nextId(1),
     m_activeObjectIterators(0),
     m_shouldCleanRemovedObjects(false)
 {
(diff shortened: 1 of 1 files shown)


tomaszkax86 (approved):

hexagonrecursion:
This is possibly made obsolete by https://github.com/melex750/colobot/pull/2

PR #1837, opened 2024-12-29
Fix `definetly` to `definitely`

eschan145:


hexagonrecursion (approved):

tomaszkax86 (approved):
Maybe in the future could include more than just a single word fix?

PR #1838, opened 2024-12-30
Fix controller command after loading a save

melex750:


hexagonrecursion:
For posterity: the bug was that the cheat command `controller` would do nothing after save&load even though the level controller was still present and working

hexagonrecursion (approved):

tomaszkax86 (approved):

PR #1849, opened 2025-01-16
Do not define main() as extern "C"

suve:
gcc15 disallows this when building with `-Wpedantic`.
```
error: cannot declare ‘::main’ with a linkage specification [-Wpedantic]
```

### DIFF
--- a/colobot-app/src/main.cpp
+++ b/colobot-app/src/main.cpp
@@ -94,10 +94,14 @@ The current layout is the following:
  - src/script - link with the CBot library
 */
 
-//! Entry point to the program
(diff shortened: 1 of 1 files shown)


eschan145:
Although I might be wrong, I believe this is intended to prevent name mangling and because SDL declares a `main()` function.

suve:
You're right. SDL declares its own `main()` function, and then uses a macro to rename your `main` into `SDL_main`. However, this isn't done for all platforms.

eschan145 (approved):

tomaszkax86 (approved):

PR #1850, opened 2025-01-19
Improve logging of errors when files fail to open

hexagonrecursion:
We now log:
1. The **exact** filename we passed into PHYSFS_open...()
2. The error code from PHYSFS_open...()
3. The result of trying to stat the file (is it a directory? is it readonly?)

### DIFF
--- a/colobot-base/src/common/CMakeLists.txt
+++ b/colobot-base/src/common/CMakeLists.txt
@@ -35,6 +35,8 @@ target_sources(Colobot-Base PRIVATE
     resources/outputstream.h
     resources/outputstreambuffer.cpp
     resources/outputstreambuffer.h
+    resources/physfs_utils.cpp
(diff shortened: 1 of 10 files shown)


colobot-base/src/common/CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ target_sources(Colobot-Base PRIVATE
     resources/sdl_memory_wrapper.h
     resources/sndfile_wrapper.cpp
     resources/sndfile_wrapper.h
+    resources/physfs_utils.cpp
+    resources/physfs_utils.h
```
melex750: I think it's supposed to be in alphabetical order.

hexagonrecursion: Thanks. Done.

tomaszkax86 (approved):

PR #1897, opened 2026-04-27
agent-server: expand API and add full exercise test suite (ch1-ch7)

RicardoRdzG:
## Summary

- Adds `/objects`, `/get_speed`, `/set_speed`, `/get_program`, key `down`/`up` actions, `walk_me_to`, and program-slot helpers to the agent server and `AgentClient`
- New `test_08` covers the program API; `test_09` covers all 7 Chapter 1 exercises; `test_10` covers 31 exercises across Chapters 2–7 (38 levels total, all passing)
- Two CBot script bugs required test-side workarounds:
  - **ExchangePost race condition** (Remote Control #2, ch6/lvl3): at 16× simulation speed, the slave can read `param=0` between two consecutive `send()` calls in the controller; fixed by sending `param` before `order` in the override script
  - **CBot `int`/`nan` sentinel deadlock** (Remote Control #4, ch7/lvl2): `int m_type = nan` is stored as 0 (not NaN) because `static_cast<int>(NaN)` is implementation-defined; comparisons against `nan` always yield false, deadlocking the exchange class; fixed by using `-1` as a sentinel in a replacement class-based protocol deployed via a prepare hook

## Test plan

- [ ] Run `COLOBOT_AGENT_URL=http://localhost:7777 python3 -m pytest tests/agent/ -v` and verify all 38 tests in test_09/test_10 pass
- [ ] Run the full suite (`tests/agent/`) to confirm no regressions in test_01–08
- [ ] Build with `cmake --preset MacOS-CI && cmake --build --preset MacOS-CI` to confirm no C++ compile errors from the new agent server endpoints

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RicardoRdzG:
Opened by mistake — PR should target the fork, not the upstream repo.

